PR #17, opened 2014-12-16
Added ability to pass a carrier param to address.Verify.

sawyer:


jmalatia:
What are the options for the carrier param in address.Verify?  Only usps seems to work?  is UPS an option?

att14:
The only other available carrier is Fedex currently.

jmalatia:
Is this documented anywhere?  As I also see other things in the code that aren't in the API docs at easypost.com.

att14:
Anything that is not documented currently is pre-release and added on a customer request basis. If you have any feature requests, not relating to this client library, feel free to email support@easypost.com.

PR #29, opened 2015-04-17
Added property for specifying carrier accounts

swrhim:
Addded unit test to test new property

### DIFF
--- a/EasyPostTest/ShipmentTest.cs
+++ b/EasyPostTest/ShipmentTest.cs
@@ -5,13 +5,16 @@ using System.Linq;
 using System.Collections.Generic;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace EasyPostTest {
+namespace EasyPostTest
+{
     [TestClass]
-    public class ShipmentTest {
+    public class ShipmentTest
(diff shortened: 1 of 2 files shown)


EasyPostTest/ShipmentTest.cs
```diff
@@ -166,5 +182,39 @@ public class ShipmentTest {
...
+            var valid = true;
+            if (shipment.rates.Count > 0)
+            {
+                foreach (var rate in shipment.rates)
+                {
+                    if (rate.carrier_account_id != "ca_qn6QC6fd") valid = false;
+                }
+            }
+
+            Assert.IsTrue(valid);
```
att14: Should be able to use [`List<T>.TrueForAll`](https://msdn.microsoft.com/en-us/library/kdxe4x4w%28v=vs.110%29.aspx) here. Everything else looks fine, but can you change your project settings to put the curly brace on the same line as the method definitions and reformat these files?

swrhim:
Hi @att14,

I made the changes that you suggested. 

Thanks

att14:
Thanks! I'll release a new version tomorrow.

PR #33, opened 2015-07-21
Added async wrappers for all API calls

dmmatson:
I ended up wrapping several of your API calls up so they can be called with the C# 5 async/await paradigm in my project. So with a little extra work I added an asynchronous version of every API call. If you feel like this adds too much cruft to the project no worries.

In full disclosure this was mostly a copy+paste job so I haven't tested every call individually.

### DIFF
--- a/EasyPost/Address.Async.cs
+++ b/EasyPost/Address.Async.cs
@@ -1,0 +1,75 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace EasyPost {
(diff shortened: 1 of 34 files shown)


att14:
This is cool. I would like to see some tests though. I'll take a look at this when I get a chance.

att14:
I copied the code from the README you added into a test and it doesn't compile. If you want this to get merged, I'll need to see more comprehensive tests and updated README. Closing for now.

PR #35, opened 2016-01-07
Refactor client connection instantiation to support multiple API keys

kcamp:
Here is the complete set of changes to help the EasyPost lib support multiple API keys per application.

As discussed, the current limitation of the static fields ```EasyPost.Client.apiKey``` creates contention across multiple calls to ```EasyPost.Client``` (as in web requests/threads/series of method calls), which can be seen by running the ClientTest fixture in its entirety prior to commit 5198aef8be53bd102b7e5093d571fe7acb4eea60

By refactoring the ApiKey + ApiBaseUrl to a separate ```ClientConfiguration``` class and by providing the appropriate constructor overloads we can create separation; further, adding a well-known location for the ```EasyPost.Request``` class to delegate the construction of the ```EasyPost.Client``` class, it opens up paths for managing multiple API Keys for cases where the operating environment may be multi-tenanted.

### DIFF
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -29,7 +29,7 @@ namespace EasyPost {
         public Client(ClientConfiguration clientConfiguration) {
             System.Net.ServicePointManager.SecurityProtocol = Security.GetProtocol();
 
-            if (clientConfiguration == null) throw new ArgumentNullException(nameof(clientConfiguration));
+            if (clientConfiguration == null) throw new ArgumentNullException("clientConfiguration");
             configuration = clientConfiguration;
 
             client = new RestClient(clientConfiguration.ApiBase);
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -12,18 +12,27 @@ using System.Reflection;
 
 namespace EasyPost {
     public class Client {
+
+        internal const string ApiBaseUrl = "https://api.easypost.com/v2";
+
         public static string apiKey { get; set; }
         public static string apiBase { get; set; }
 
         public string version;
 
         internal RestClient client;
+        internal ClientConfiguration configuration;
 
-        public Client(string apiBaseUrl = "https://api.easypost.com/v2") {
+        public Client(string apiBaseUrl = ApiBaseUrl)
+            : this(new ClientConfiguration(apiKey, apiBase ?? apiBaseUrl)) { }
+
+        public Client(ClientConfiguration clientConfiguration) {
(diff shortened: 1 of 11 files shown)


EasyPost/Client.cs
```diff
@@ -12,18 +12,27 @@
...
-        public Client(string apiBaseUrl = "https://api.easypost.com/v2") {
+        public Client(string apiBaseUrl = ApiBaseUrl)
+            : this(new ClientConfiguration(apiKey, apiBase ?? apiBaseUrl)) { }
+
+        public Client(ClientConfiguration clientConfiguration) {
             System.Net.ServicePointManager.SecurityProtocol = Security.GetProtocol();
 
-            apiBaseUrl = apiBase ?? apiBaseUrl;
-            client = new RestClient(apiBaseUrl);
+            if (clientConfiguration == null) throw new ArgumentNullException(nameof(clientConfiguration));
```
att14: This is introduced in .NET 4.6, this library was written with 4.5. Is there an equivalent function in 4.5?

att14: This being `nameof`

EasyPostTest/ClientFactoryTest.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using EasyPost;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace EasyPostTest
+{
+    [TestClass]
+    public class ClientFactoryTest {
+
+        private class FakeClientFactoryForKeyPlusUrl : IClientFactory {
```
att14: Getting a build error here in 4.5. "The type or namespace name 'IClientFactory' could not be found (are you missing a using directive or an assembly reference?)"

EasyPost/Client.cs
att14: The library should be compatible with 4.0 as well.

EasyPostTest/ClientFactoryTest.cs
kcamp: I think this was probably because of the 'nameof' mistake?  the csproj files appear correct, I cannot repro.

EasyPost/Client.cs
kcamp: yeah, i missed that stray 'nameof' - cleaned that up, pushed the fix.

kcamp:
the last fix was wrong; fixing that up.  will reopen when it's ready.

kcamp:
can't reopen, will create new.  (sorry!)

att14:
btw, just pushed a fix for the tests.

PR #38, opened 2016-01-12
Refactor ClientManager implementation to support instantiation of Client per-caller

kcamp:
Here is a revision of the work that was done to revise the implementation of ClientManager; what I ultimately need is to be able to control the API key that is used per-request [to the EasyPost API] on a per-caller basis.

For example, if I have a multi tenant app; consider TenantA and TenantB.  They will both have their own separate API keys.  The IIS process will handle requests from multiple users; any given request could be from a user on TenantA or TenantB.  I need to correctly associate the downstream request [made to EasyPost] to the correct API; for this to work, the ```EasyPost.Client``` must be constructed properly.

The one thing that I would consider as an improvement would be to refactor ```ClientManager``` to ```ClientConfigurationManager``` and simply manage the issuance of the configuration object, rather than force the developer to be aware of the implementation details of ```EasyPost.Client```

Let me know what revisions [if any] may be desirable.


kcamp:
Also, I cleaned up the ordering/implementation of expected vs. actual in a couple of unit tests to be more consistent.

att14:
Awesome! Looks good to me.

Issue #64, opened 2016-12-08
Add carrier_notification_email to Options


mwaldt:

mwaldt:
https://github.com/EasyPost/easypost-csharp/pull/65

PR #65, opened 2016-12-08
Added carrier_notification_email

mwaldt:
https://github.com/EasyPost/easypost-csharp/issues/64


att14:
Do we also want to add `carrier_notification_sms`?

mwaldt:
I went ahead and also added carrier_notification_sms

PR #68, opened 2017-01-25
Fixed 'DateTimeInvalidLocalFormat'

Giorgio-Montagnoli:
Additional information: This can happen when calling DateTime.ToString using the 'z' format specifier, which will include a local time zone offset in the output. In that case, either use the 'Z' format specifier, which designates a UTC time, or use the 'o' format string, which is the recommended way to persist a DateTime in text. This can also occur when passing a DateTime to be serialized by XmlConvert or DataSet. If using XmlConvert.ToString, pass in XmlDateTimeSerializationMode.RoundtripKind to serialize correctly. If using DataSet, set the DateTimeMode on the DataColumn object to DataSetDateTime.Utc.


Giorgio-Montagnoli:
Fix the "Validation failed: Weight is not a number" error
----
When you are not in US (we are in Italy) you got this error if you use a number with decimal values.
I could fix setting the Thread Culture in my program, but should be better to force it in your code, cause the data you sent is always in "en-US" format.

EasyPost/Client.cs
```diff
@@ -20,6 +21,7 @@ public class Client {
             configuration = clientConfiguration;
 
             client = new RestClient(clientConfiguration.ApiBase);
+            Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
```
att14: This seems like it would clobber an existing `CurrentCulture` set by a user of this client. My thought would be that this should be left up to the user, not the library.

EasyPost/Request.cs
```diff
@@ -103,7 +103,7 @@ public class Request {
                         result.AddRange(FlattenParameters(list[i], string.Concat(parent, "[", pair.Key, "][", i, "]")));
                     }
                 } else if (pair.Value is DateTime) {
-                    result.Add(new Tuple<string, string>(string.Concat(parent, "[", pair.Key, "]"), Convert.ToString(((DateTime)pair.Value).ToString("yyyy-MM-ddTHH:mm:sszzz"))));
+                    result.Add(new Tuple<string, string>(string.Concat(parent, "[", pair.Key, "]"), ((DateTime)pair.Value).ToLocalTime().ToString("yyyy-MM-ddTHH:mm:sszzz")));
```
att14: This could be correct, but I can't say for certain unless I see a failing test and the fact that this fixes it with the expected output.

Giorgio-Montagnoli:
I propose to you a new way to fix the decimal separator issue.

EasyPost/Client.cs
Giorgio-Montagnoli: Ok, this is not a solution. See below #e8b0975

EasyPost/Request.cs
Giorgio-Montagnoli: Hi Andrew,

if you debug this
`new` DateTime(636210240270000000, DateTimeKind.Utc).ToString("yyyy-MM-ddTHH:mm:sszzz");
you'll se just a warning, so you can't run a test that fails unless the test consider a warning as an error.
But, in my humble opinion, you can fix and prevent the warning.

att14: I just don't see how this would happen in normal shipment creation, that is why I am interested in a test case to reproduce the issue using the library.

Giorgio-Montagnoli: In my case, the warning always fire, but the creation always succeed.
If you try to debug the code, the warning fires.

att14: I've fixed this issue, but it requires that you use UTC for these times. Can you merge my changes in and then I'd be willing to merge the other changes you have in the request.

Giorgio-Montagnoli: Ok, I have fixed the conflicts.

PR #71, opened 2017-02-16
Add new shipment fields

adamchester:
- commercial_invoice_signature
 - parties_to_transaction_are_related

cc @jingta


jingta:
These should actually be in the options dictionary, not on the root object. I'll close this PR

PR #87, opened 2017-04-26
Important Fixes to Nuget Feed in General, Adds XML Docs to Nuget

vongillern:
Sorry for the long post, but I think this is important and can make consumption of the client wrapper much nicer for everyone.

This is my very first pull request ever. Not just for your project. Hopefully I don't screw anything up too badly, especially since this changes some critical moving pieces although overall changes are limited. 

I'm halfway expecting that this gets rejected in some form, revised and tried again. 

**Here is what I viewed as wrong and why:**
1. This all started because the .XML documentation (EasyPost.XML) was not being included in the nuget package. This means that while the source code has nice contextual usage information in the <summary> comment tags, no nuget consumer can see it in intellisense. 

2. So, I double clicked on project properties. Clicked on the build tab (it is between "application" and "build events") and I checked the "XML documentation file" checkbox under the Output section while I had the Release configuration active. I did this for all three projects

3. I saw that all three projects (EasyPost, EasyPost35 and EasyPost40) were outputting their contents to /lib/ and /lib/ was being checked in to the repository. This means every build has build artifacts (dlls) included in the respository which makes the repo huge. 

4. The nuspec includes all files in the repository because none are specifically mentioned. That means that every download gets all the source which is not needed at all. And the /lib/ folder which is the main folder used in nuget contains superfluous DLLs (RestSharp, Newtonsoft.Json) which can be referenced by the consuming project. Depending on the order that nuget packages are added to a consuming application, the wrong version number could be referenced by the project and cause runtime binding issues because another nuget package relies on a different version of RestSharp or Newtonsoft.Json. This did in fact happen to me and I thought problem was with Twilio (which also uses RestSharp) instead of the referenced version of Restsharp in my project. 

**In order to fix these general issues I did the following:**

1. Changed the release output path to be "\bin\release\" for all projects
2. I have checked the the XML documenation file checkbox and it is in \bin\release\[PROJECT NAME].xml
3. I have made heavy changes to the nuspec so that there are different dependency groups by target framework (although they're currently all identical)
4. The files that are put into the nupkg are only the files specific to EasyPost for each target framework. As of now this is just the dll and xml file. 
5. Deleted the repository's /lib/ folder as it is no longer the targeted destination for the builds and would cause a lot of confusion.

**Problems that I had**
1. I was having problems signing the assemblies so I just turned off signing.
2. I could not get the test project to build, so I unloaded it. I didn't touch any C# so everything should still be fine.

Errata:
1. As someone who has been doing .net forever, assembly signing is a huge PITA with practically zero upside. I would recommend skipping it and it's associated heartache and confusion.
2. While I do know enough about nuget to know that you don't want to package the entire repository, I'm not super familiar with all of the targeted framework stuff. I feel like there is an easier way to build a small library like this that targets those three frameworks without requiring multiple projects and linked files but no promises. I do feel it is even more likely you can do the nuget packaging more simply/efficiently with the nuspec, but I just don't know how. 

This all should mostly just work though. I hope 👍 

Cheers,
Jon


vongillern:
So as I was digging into some dependency issues I found another related to this PR. The existing nuspec in your repo AND my new nuspec take a dependency on the `RestSharp` package, however the packages.config in the projects rely on `RestSharpSigned`. 

It doesn't screw up anyone right now because the signed version of the RestSharp assembly is being delivered via the EasyPost nuget package. Again, this causes issues when you have other dependencies that rely on RestSharp (in my case, a version of Twilio).

I won't make a revised PR until you guys tell me if you're planning on staying with a signed assembly or not. If you are, each project is referencing the EasyPost.pfx file, but it is not in the respository. HOWEVER, it is being packaged and delivered with with the nuget package (rename the nupkg extension to zip and open it and you'll see it). I don't know the security implications of this.  My guess is that you are keeping the pfx somewhere else and copying it in the build process.

vongillern:
ping

vongillern:
I'm going to try to break this up into multiple PRs so it is a little easier to follow. Should be submitting them tomorrow or Friday

PR #88, opened 2017-05-08
Error parsing bug fix

eliekstein:
I always had problems with parsing error objects from the `HttpException` because included the root name.
Also the error class didn't have all fields.

So I hided the Load method from the base class and went one step into the root to fix it.

### DIFF
--- a/EasyPost/Exception.cs
+++ b/EasyPost/Exception.cs
@@ -4,11 +4,13 @@ namespace EasyPost {
     public class HttpException : Exception {
         public int StatusCode;
         public string Code;
+        public Error Error;
 
         public HttpException(int statusCode, string code, string message)
             : base(message) {
             StatusCode = statusCode;
             Code = code;
+            Error = Error.Load<Error>(message);
         }
     }
 
--- a/EasyPostTest/ErrorTest.cs
+++ b/EasyPostTest/ErrorTest.cs
@@ -12,19 +12,50 @@ namespace EasyPostTest
     public class ErrorTest
     {
         string  error = "{\"error\":{\"code\":\"ADDRESS.VERIFY.FAILURE\",\"message\":\"Unable to verify address.\",\"errors\":[{\"code\":\"E.ADDRESS.NOT_FOUND\",\"field\":\"address\",\"suggestion\":null,\"message\":\"Address not found\"}]}}";
+        private Address address;
 
         [TestInitialize]
         public void Initialize()
         {
             ClientManager.SetCurrent("cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi");
+            address = new Address()
+            {
+                company = "Simpler Postage Inc",
+                street1 = "1645456 Townsend Street",
+                street2 = "Unit 1",
+                city = "San Francisco",
+                state = "CA",
+                country = "US",
+                zip = "94107"
+            };
         }
 
         [TestMethod]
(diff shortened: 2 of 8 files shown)


eliekstein:
@att14 The reason i updated the Error class was because there was more data returned in the error response then in the class.

I did miss explaining the point that parsing the `HttpException.Message` is still required to acquire a proper Error object, I'm going to add it to the Test.

Do you want to have an error object in the `HttpException `class?

EasyPost/Exception.cs
```diff
@@ -4,11 +4,13 @@ namespace EasyPost {
     public class HttpException : Exception {
         public int StatusCode;
         public string Code;
+        public Error Error;
 
         public HttpException(int statusCode, string code, string message)
             : base(message) {
             StatusCode = statusCode;
             Code = code;
+            Error = Error.Load<Error>(message);
```
att14: The signature of this function should change to `public HttpException(int statusCode, string code, string message, List<Error> errors)`. In `EasyPost/Client.cs` the you should call `Error.Load<Error>` on the errors map of the `errors` key. All error objects from EasyPost will be of the following format:

```js
{
  "error": {
    "code": "FOO.BAR",
    "message": "foobar",
    "errors": [
      {
        field: "foo",
        message: "invalid"
      },
      {
        field: "bar",
        message: "invalid"
      }
    ]
  }
}
```

https://www.easypost.com/docs/api#errors
https://www.easypost.com/errors-guide.html

EasyPostTest/ErrorTest.cs
```diff
@@ -32,5 +41,27 @@ public void TestErrorLoad()
...
+
+        [TestMethod]
+        public void TestAddressError()
+        {
+            address.Create();
+            HttpException ex = null;
+            try {
+                address.Verify();
+            }
+            catch (HttpException hex)
```
att14: What is `hex`?

EasyPostTest/ErrorTest.cs
```diff
@@ -32,5 +41,27 @@ public void TestErrorLoad()
...
+            {
+                ex = hex;
+            }
+
+            var sut = Error.Load<Error>(ex.Message);
```
att14: What is `sut`?

EasyPostTest/ErrorTest.cs
eliekstein: That would stand for HttpWebException it's just my habits, i'll standardize it based on your standards.

EasyPostTest/ErrorTest.cs
eliekstein: Same as above, we'll fix this up.

EasyPost/Exception.cs
eliekstein: I get the idea, stay tuned for the update!

j6967chen: The error class should include 'code' & 'suggestion'. 

The following is a sample from api#errors doc. 

` "verifications": {
    "delivery": {
      "success": false,
      "errors": [
        {
          "suggestion": null,
          "code": "E.ADDRESS.NOT_FOUND",
          "field": "address",
          "message": "Address not found"
        },
        {
          "suggestion": null,
          "code": "E.HOUSE_NUMBER.MISSING",
          "field": "street1",
          "message": "House number is missing"
        }
      ],
      "details": {}
    }
  }`

PR #93, opened 2017-05-24
get rates when the list is null or is an empty list

ryn0:
In a situation where you retrieve a shipment and then try to get the lowest rate, the list is empty and it doesn't get the rates which results in an exception. 

Ex:
```
                var shipmentDetail = Shipment.Retrieve("SOME_SHIPMENT_ID");
              
                shipmentDetail.Buy(shipmentDetail.LowestRate(
                                    includeServices: new List<string>() { "First" },
                                    excludeCarriers: new List<string>() { "FedEx" }
                                    ));
```


ryn0:
It appears this issue can be overcome by calling GetRates(); before buying

PR #99, opened 2017-08-01
Update to handle working with multiple apiKeys in a multithreaded environment

technology-shipcaddie-com:
The current implementation of easypost-csharp uses a ClientManager that is a static class, and thus is a Singleton and not thread safe because it stores the apiKey.
Currently, the apiKey gets stored in the ClientManager static class and is then pulled whenever an execute happens.  This works great in a single apiKey environment for both single and multi threaded environments.

This methodology breaks down quickly when in a multi-apiKey environment that is multi-threaded.
The following steps show an example of the problem that happens on busy systems.  Both examples below were happening frequently every day on our system when under load:

**Example 1:**

Thread 1:

- Client 1 sets the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 1");
**NOTE: singleton client Manager is now set to "apiKey 1" for everyone**
- Client 1 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();
- Client 1 gets lowest rate --- var rate = shipment.LowestRate(carriers, services);

Thread 2:
- Client 2 set the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 2");
**NOTE: singleton client Manager is now set to "apiKey 2" for everyone**
- Client 2 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();

Thread 1:
- Client 1 Buys label --- shipment.Buy(rate);
NOTE: this function then throws an error with the following description. "Object reference not set to an instance of an object"  This is because Client 1 is trying to purchase their own shipment with the "apiKey 2" of Client 2 which is in the ClientManager singleton object that the Buy method will pull to make its call.

Thread 2:
- Client 2 goes on and is able to work as if nothing has happened because the apiKey is their own "apiKey 2" value.

**Example 2:**

Thread 1:

- Client 1 sets the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 1");
**NOTE: singleton client Manager is now set to "apiKey 1" for everyone**
- Client 1 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();

Thread 2:
- Client 2 set the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 2");
**NOTE: singleton client Manager is now set to "apiKey 2" for everyone**
- Client 2 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();

Thread 1:
- Client 1 gets lowest rate --- var rate = shipment.LowestRate(carriers, services);
- Client 1 Buys label --- shipment.Buy(rate);
- Client 1 Gets Label from url in shipment object.
NOTE: **This is the worst situation of all.**  Client 1 purchases and gets a label using Client 2's apiKey.  No error message is ever returned or indicated.  All charges will show up on Client 2's account.  Client 1 will not be able to look up the shipment information with the shp_..... key because it is not part of their account.  Client 2 will have an extra shipment appear on their account that they cannot account for.

**This pull request is a solution to this problem.**  It is done in a way that it is a non-breaking change for existing users as there are only optional parameters that they can ignore.  The solution involves adding an OPTIONAL parameter to any method that makes a call to Easypost with the Execute command.  The code can then pass in the explicit apiKey to use for the call and completely bypass the ClientManager singleton object making sure the correct apiKey is used for the action.

### DIFF
--- a/EasyPost.Net35/EasyPost.Net35.csproj
+++ b/EasyPost.Net35/EasyPost.Net35.csproj
@@ -35,7 +35,8 @@
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
   <PropertyGroup>
-    <AssemblyOriginatorKeyFile>EasyPost.pfx</AssemblyOriginatorKeyFile>
+    <AssemblyOriginatorKeyFile>
+    </AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
@@ -193,7 +194,6 @@
     <Compile Include="Tuple.cs" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="EasyPost.pfx" />
     <None Include="packages.config" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
--- a/EasyPost.Net35/EasyPost.Net35.csproj
+++ b/EasyPost.Net35/EasyPost.Net35.csproj
@@ -35,8 +35,7 @@
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
   <PropertyGroup>
-    <AssemblyOriginatorKeyFile>
-    </AssemblyOriginatorKeyFile>
+    <AssemblyOriginatorKeyFile>EasyPost.pfx</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
@@ -194,6 +193,7 @@
     <Compile Include="Tuple.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="EasyPost.pfx" />
     <None Include="packages.config" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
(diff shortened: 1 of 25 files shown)


technology-shipcaddie-com:
NOTE: This only alters the EasyPost project folder.

It does not alter the EasyPost.Net35, EasyPost.Net40, or EasyPostTest project folders.

att14:
Is there any way we can just fix `ClientManager` instead of adding another parameter to these functions?

technology-shipcaddie-com:
The only way I know how to fix it is to have the ClientManager NOT be a singleton(IE: non-static).  That simple fact is what is breaking the multi-threaded, multi-apiKey environment because of the race conditions of who calls the SetCurrent first/second and what order the threads call the methods (Buy, etc) in relation to the calls to SetCurrent as I described above.

One approach to this is maybe moving the ClientManager guts into the resource class and changing them to NON-static.  That way they are only accessible from within an instance of an object, say "Shipment".  All methods within the Shipment(and all others) object that make a call to Request.Execute would need to reference the non-static apiKey from within the Shipment object and pass it to the Request.Execute.

The Execute<T> method within the Request Class would no longer be able to just call ClientManager as it would need to reference the apiKey that was passed in.  This is because the ClientManger is not a safe object.


I am sure the rabbit hole gets deeper from there.

That is why I took the approach I did.  ALL parameters that I passed in, are OPTIONAL.  They are only required when using multiple apiKeys in their system.  Any current user of the EasyPost object does not need to use them as they can continue to use the static ClientManager object.

att14:
I believe https://github.com/EasyPost/easypost-csharp/pull/101 solves your problem.

MazdaRacerBoy:
You could default to using the static key from the ClientManager, but allow it to be overwritten on a per-client basis.  That would keep backwards compatibility with old code, and allow the new code to push in the unique keys.  Just add a default parameter to the method signatures.  If it's null, then pull in the key from the static ClientManager.  If not null, then use it.

PR #101, opened 2017-08-08
make getCurrent thread static

att14:
@technology-shipcaddie-com does this do what you need? https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?view=netframework-4.7


yoelhalb:
This will not work in a Web Server enviroment such as ASP.NET.
See https://stackoverflow.com/questions/4791208/threadstaticattribute-in-asp-net and https://www.hanselman.com/blog/ATaleOfTwoTechniquesTheThreadStaticAttributeAndSystemWebHttpContextCurrentItems.aspx.

Similarly will not it work when dealing with Tasks instead of Threads, or the ThreadPool, or TPL.

PR #115, opened 2018-05-14
Add ability to set pickup_min_datetime

scotthendrickson:
When using DHL eComm for their Metro service they require readyBy to be passed in, I think it's a recent change, and pickup_min_datetime is how that's set via EasyPost. Need it added to the Client Library so that we can do this.


scotthendrickson:
Need to actually test this just in case. I'll reopen once I am sure this won't cause issues.

scotthendrickson:
I'm not certain how to test this on my Mac since I'm using Visual Studio Code and loaded the client library as nuget package which won't let me edit it. I'm going to try and get that figured out but this apparently is a priority now so I'm just going to throw this up first.

PR #124, opened 2018-12-11
- Added support for NetStandard 2.0

twurm:
Added support for .NetStandard 2.0 as an output.  To simplify things further you could remove the NetCoreApp version since any program that targets NetCoreApp20 will work with NetStandard20.  That's up to you, it would be a breaking change obviously since the package names are different.


att14:
I am getting the following error when I try to use this. Any ideas? Can you provide a test solution that uses .Net Standard?

```
System.IO.FileNotFoundException: Could not load file or assembly 'EasyPost.NetStandard20, Version=2.5.1.0, Culture=neutral, PublicKeyToken=46ff7362c295cab3'. The system cannot find the file specified.
```

twurm:
@Justintime50 I can take a look when I have a moment. We no longer use this in our code base.

Justintime50:
@twurm fair enough, thanks for the info! We may look into this separately down the road if you aren't able to take a look. Understandable if you don't want to dedicate the time to it since you no longer use this library.

epclee:
Closing out of date PR.

PR #140, opened 2020-02-14
Create VerificationDetails

rbellow:
Per the EasyPost documentation, there is a `VerificationDetails` object:

Screen Shot 2021-06-28 at 3 12 52 PM

This doesn't exist in this library and should be added.

The companion PR is #141.


Justintime50 (changes requested):
We'll want this file to have the proper extension of `.cs`, can you make this correction and we can take another look?

Justintime50 (approved):

Sarahbear1926 (approved):

jontsai (approved):

PR #143, opened 2020-03-08
batch list [fixes #95]

### LINKED ISSUE #95: Add Support to List Batches
Currently you cannot retrieve a list of batches. This should be added.

att14:
This is complete, however there is a bug with the list endpoint for batches and I don't want to merge this yet.


att14:
This can probably be merged it just won't work correctly until the backend is fixed. I'm closing for now, someone at EasyPost can take it over if they would like.

PR #144, opened 2020-03-08
Remove hard-coded Options class

roehnan:
Dictionary<string, object>, which permits
the use of undocumented (yet widely-used)
shipment options. 

Other strongly-typed client libraries (such 
as Java) do not have an Options class that 
complicates using parameters. 

This Options class was often out of sync 
with the API and it required that the client 
library be updated with every API shipment
options change. The library should be 
loosely couple with the API and this class
prevented that.


roehnan:
Closing pull request as I neglected to update test code. I will update and resubmit the request.

PR #145, opened 2020-03-09
Remove hard-coded Options class (with tests)

roehnan:
Remove hard-coded Options class and use
Dictionary<string, object>, which permits
the use of undocumented (yet widely-used)
shipment options.

Other strongly-typed client libraries (such
as Java) do not have an Options class that
complicates using parameters.

This Options class was often out of sync
with the API and it required that the client
library be updated with every API shipment
options change. The library should be
loosely couple with the API and this class
prevented that.


jontsai:
The code changes in this pull request do not conform to the existing software pattern. Instead, it introduces a new paradigm and removes the `Options` object entirely.

As such, it will break existing integrations for customers anywhere who are using or extending the `Options` object.

PR #155, opened 2021-01-20
add basic github workflow using dotnet 2.2.*

Roguelazer:
will this work? who knows!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: .NET
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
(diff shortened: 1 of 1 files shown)


jontsai:
Is the failing build still a concern? Otherwise, I'm down to merge this to try it out, and iterate from there.

Justintime50:
It's failing because we target 3-4 different SDK's in the build and this GitHub Action is only using 1. We'll need to configure it right to get it to pass.

This is part something I've been chipping away at, didn't work on macOS with Mono but should work on Windows.

Roguelazer:
Now that github supports windows instances for Actions, we should be able to build this with the native SDKs.

epclee:
No longer needed. Closing.

nwithan8:
Handled by #186

Issue #125, opened 2018-12-12
Multiple EasyPost API keys don't work with latest version 2.5.1


taivt0109:
If you are operating with multiple EasyPost API keys, or wish to delegate the construction of the client requests, configure the ClientManager with a delegate at application initialization.

using EasyPost;

ClientManager.SetCurrent(() => new Client(new ClientConfiguration("yourApiKeyHere")));

I reference this document. But it doesn't work with version 2.5.1

i got an error " 'Client' does not contain a constructor that takes 1 arguments "

att14:
I have two tests that should cover this exact situation. https://github.com/EasyPost/easypost-csharp/blob/master/EasyPostTest/ClientManagerTest.cs#L30-L44'

Do you have a more specific example or some modification to the existing tests that would reproduce this behavior?

nicknasirpour:
Hi, I am having the same issue accessing this constructor.

The Client constructor in reference has an "internal" access modifier, which is why you are able to use it within your EasyPostTest project... it is within the same solution/assembly... But we are unable to use it in our solutions. It should be marked as public if you would like us to use to have access.

nicknasirpour:
Here is the commit where the access modifier is changed from public to internal:

https://github.com/EasyPost/easypost-csharp/pull/100/commits/2f8cb663998e93902f0c50303900bd2b3de6b287

Justintime50:
The constructor was corrected in #158. It should also be noted that the way in which this library is written does not allow for thread-safe multi-API key usage (see #98). We are considering a refactor that may allow for this; however, it won't be in the immediate future.

nwithan8:
As of [v4.0.0](https://github.com/EasyPost/easypost-csharp/releases/tag/v4.0.0), this library is instance-based and considered thread-safe.

PR #158, opened 2021-02-24
Add public access to Client delegate constructor

nicknasirpour:
This constructor is listed in the docs as though it should be publicly accessible.

### DIFF
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -13,7 +13,7 @@ namespace EasyPost {
         internal RestClient client;
         internal ClientConfiguration configuration;
 
-        internal Client(ClientConfiguration clientConfiguration) {
+        public Client(ClientConfiguration clientConfiguration) {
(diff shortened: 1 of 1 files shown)


Justintime50:
Looks like this was indeed at [one point public](https://github.com/EasyPost/easypost-csharp/commit/31dbf12b39b03a2a3b151f2838abc925e94fe827#diff-1ae0f58be70a36951459d85d13d43e07ee650333ff967662edd5bd1cf3d2964dL16), I'm not entirely sure the implications of why this change was made.

Justintime50:
This was created in response to issue #125

Justintime50 (approved):
It looks like this was at one point public and has since been made internal. Other users are wanting to use this functionality, I don't see any reason initially why we couldn't put it back to what it was before.

jontsai (approved):

Sarahbear1926 (approved):

PR #159, opened 2021-06-08
docs: removes CreateAndBuy example from README

### LINKED ISSUE #134: Batch.CreateAndBuy does not exist?
The doc overview on [the project home page](https://github.com/EasyPost/easypost-csharp#asynchronous-batch-processing) shows that you can Create and Buy a Batch with one call:
> Batch batch = Batch.CreateAndBuy(new Dictionary<string, object>() {
    { "reference", "MyReference" },
    { "shipments", new List<Dictionary<string, object>>() { shipment } }
});

But the [Batch.cs](https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Batch.cs) class does not have a CreateAndBuy method? Is there a way to do this?

Justintime50:
Over the years, we've had many users write in confused by the `CreateAndBuy` example in our README. The code doesn't exist (maybe it did at one point?) I've provided an example of how to buy a batch (AKA "CreateAndBuy") without the need for us to actually add logic as this functionality is already supported via this library (see #134 for details).

This PR removes the bad `CreateAndBuy` example from the README.

Closes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,37 +171,6 @@ foreach (Form form in shipment.forms) {
 }
 ```
 
-### Asynchronous Batch Processing
-
-The `Batch` object allows you to perform operations on multiple `Shipment`s at once. This includes scheduling a `Pickup`, creating a `ScanForm` and consolidating labels. Operations performed on a `Batch` are asynchronous and take advantage of our [webhoook](https://www.easypost.com/docs/api/csharp#events) infrastructure.
-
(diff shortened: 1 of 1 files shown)


ejachang (approved):

ajnassar (approved):
LGTM

jontsai (approved):

PR #161, opened 2021-07-12
add thread-safety note to README

Roguelazer:


README.md
```diff
@@ -171,6 +171,10 @@ foreach (Form form in shipment.forms) {
 }
 ```
 
+### Warning about Threads
+
+NOTE: The EasyPost .NET client library (in p;articular, the `ClientManager` global object) is not threadsafe; do not attempt to perform requests from multiple threads in parallel. This can be particularly problematic if using multiple API keys; make sure to always use a Mutex, Monitor, or other synchronization method to ensure that concurrent requests do not enter the EasyPost library from different threads.
```
Justintime50, line 176: Correct `p;articular` to `particular`

Justintime50 (changes requested):

Justintime50 (approved):

jontsai (approved):

ejachang (approved):

PR #162, opened 2021-07-13
fix: adds missing VerificationDetails compile targets

Justintime50:
We recently merged in a fix that adds the missing `VerificationDetails` object. This is in a new file. The PR was missing the compile targets to include it in the project. This simply adds the compile targets for the new file so when the project gets built, it's included.


Sarahbear1926 (approved):

jontsai (approved):
LGTM

PR #170, opened 2021-07-21
Add smart rate

hellbinder:
Added smart rate implementation


EasyPost/Shipment.cs
```diff
@@ -128,6 +128,18 @@ public class Shipment : Resource {
             rates = request.Execute<Shipment>().rates;
         }
 
+        public List<SmartRate> GetSmartRates()
```
Justintime50, line 131: @jontsai the interface on the other CLs was `GetSmartrates`, with the `r` being lowercased. I'm assuming we want to follow that here?

Justintime50 (changes requested):
This looks great, let's just lowercase the `r` in `GetSmartrates` (interface) so it matches our other client libraries.

EasyPost/Shipment.cs
hellbinder: renamed to GetSmartrates..

EasyPostTest/ShipmentTest.cs
```diff
@@ -310,5 +310,39 @@ public class ShipmentTest {
...
+        public void TestGetSmartrates()
+        {
+            Shipment shipment = Shipment.Create(parameters);
+            List<Smartrate> smartRateResult = shipment.GetSmartrates();
+            //Make sure shipment id from smartrate is the same as the created one
+            Assert.AreEqual(shipment.rates[0].id, smartRateResult[0].id);
+            Assert.IsNotNull(shipment.rates);
+        }
+
+        //This API call changes each time. Replicated from the python version. Will fail since percentiles are changing each time. Will check for type instead.
```
Justintime50, line 333: Let's change this to a `TODO` comment so we can come back to it later. Something along the lines like we used in the PHP library:

`//TODO: Once we've added `vcr`, assert that the following values actually match an integer`

EasyPostTest/ShipmentTest.cs
```diff
@@ -310,5 +310,39 @@ public class ShipmentTest {
...
+            Shipment shipment = Shipment.Create(parameters);
+            List<Smartrate> smartRateResult = shipment.GetSmartrates();
+            Assert.IsNotNull(smartRateResult);
+        }
+
+        [TestMethod]
+        public void TestGetSmartrates()
+        {
+            Shipment shipment = Shipment.Create(parameters);
+            List<Smartrate> smartRateResult = shipment.GetSmartrates();
```
Justintime50, line 327: `smartRateResult` should probably be `smartrateResult` So that "smartrate" is either capitalized or lowercased whenever used depending so that it feels like a single word

Justintime50 (changes requested):
Real close! A couple small comments

EasyPost/Shipment.cs
```diff
@@ -128,6 +128,18 @@ public class Shipment : Resource {
...
 
+        public List<Smartrate> GetSmartrates()
+        {
+            if (id == null)
+                Create();
+
+            Request request = new Request("v2/shipments/{id}/smartrate");
+            request.AddUrlSegment("id", id);
+            request.RootElement = "result";
+            List<Smartrate> smartRates = request.Execute<List<Smartrate>>();
```
Justintime50, line 139: `smartrates`, lowercased r

EasyPostTest/ShipmentTest.cs
```diff
@@ -310,5 +310,39 @@ public class ShipmentTest {
             ShipmentList nextShipmentList = shipmentList.Next();
             Assert.AreNotEqual(shipmentList.shipments[0].id, nextShipmentList.shipments[0].id);
         }
+
+        //Smart rate
```
Justintime50, line 314: Nit: `Smartrate`

Justintime50 (approved):
Let's go!

jontsai (approved):
LGTM

PR #172, opened 2021-08-03
Add netcore31

hellbinder:
Added .netcore 3.1 project
Added VersionInfo.cs file
Pfx file is now linked on all projects except EasyPost. This removes having to add the file for each project. 
Project files cleanup.


EasyPost/Properties/AssemblyInfo.cs
```diff
@@ -38,7 +38,5 @@
 //
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
-// [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("2.5.1.3")]
-[assembly: AssemblyInformationalVersion("2.5.1.3")]
+
+
```
Justintime50, line 42: Can we kill the double extra line an only have one?

EasyPost/VersionInfo.cs
```diff
@@ -0,0 +1,5 @@
+﻿using System.Reflection;
```
Justintime50, line 1: Can we move all the helpful `VersionAssembly` comments from their individual locations into this file?

```
// Version information for an assembly consists of the following four values:
//
//      Major Version
//      Minor Version
//      Build Number
//      Revision
//
// You can specify all the values or you can default the Build and Revision Numbers
// by using the '*' as shown below:
// [assembly: AssemblyVersion("1.0.*")]
```

EasyPost/EasyPost.csproj
```diff
@@ -92,6 +92,7 @@
     <Compile Include="Pickup.cs" />
     <Compile Include="PostageLabel.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="..\EasyPost\VersionInfo.cs" Link="VersionInfo.cs" />
```
Justintime50, line 95: Why are there 2 references to `VersionInfo.cs` here? One that goes back up a directory and one for the file in this directory? Should there only be 1 or is there a reason for the second?

Justintime50 (changes requested):
Looks great! Let's just clean it up a bit by killing off a couple of the duplicate extra lines at the end of files and moving the `AssemblyVersion` comments from the individual files to the new `VersionInfo` file.

jontsai (approved):

EasyPost/VersionInfo.cs
```diff
@@ -0,0 +1,15 @@
...
+
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("2.6.0")]
+[assembly: AssemblyInformationalVersion("2.6.0")]
+
+// Version information for an assembly consists of the following four values:
```
Justintime50, line 7: Let's add the missing `// [assembly: AssemblyVersion("1.0.*")]` to the end of this comment block and move the entire block comment above line 3 like it appeared in the individual files and then this will be good to go!

Justintime50 (changes requested):

EasyPost/VersionInfo.cs
hellbinder: Sorry about that. All set on this.

Justintime50 (approved):
Thanks for your work on this!

PR #176, opened 2021-09-25
ThreadSafe ability and test fixing.

Quanthir:
Added ability to change ClientManager mode to ThreadStatic so that it can run multi-threaded tasks with different API keys.


Justintime50:
Thanks for the PR!

Can you provide a working example so we can see these changes in action? A little snippet should suffice.

Quanthir:
There is a test case for usage. You can check that. It's also in pull request. ClientManagerTest.cs line 48

PR #181, opened 2021-11-24
Adds tax identifiers

jontsai:


Justintime50:
This actually worked? I didn't think the C# library was JSON encoded and was form encoded instead.

jontsai:
> This actually worked? I didn't think the C# library was JSON encoded and was form encoded instead.

JSON, thanks to https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Resource.cs

epclee (approved):
Looks good, merging.

PR #183, opened 2022-01-19
fix: get the test suite working

Justintime50:
This PR gets the test suite working via C#. We accomplished this by doing the following (note, some of these are temporary fixes):

1.  Moved all the test files to `EasyPost.Tests` so they could run on macOS in addition to Windows
2. Reconstructed the `EasyPost.Tests.csproj` file to use the most up-to-date unit test frameworks compatible with our library
3. Commented out failing dead tests (will rectify soon in a follow up)


nwithan8 (approved):
LGTM, glad we finally got the environment set up.

jchen293 (approved):
lgtm

PR #185, opened 2022-01-20
Include .NET version in User Agent

nwithan8:
This will include the current .NET framework that the user is running our library in in the User Agent made with every API call.

This isn't a perfect implementation, due to how Microsoft has changed how to get this information depending on the .NET framework, and it's difficult to implement a solution that can account (and work) regardless of the version the user is running (i.e. some classes needed to do the registry check for a version in newer .NET versions simply don't exist in older .NET versions, preventing the project from compiling). We can implement a more accurate solution in the future using interfaces, but for now, this will work as a first pass.
  - If the .NET framework is v4.5 or higher (or lower than .NET Core 3.0 - we are dropping .NET Core 2.0 support anyway), it will only report to us as "v4.5+" rather than a specific version. This is mostly fine, since our library only compiles on up to .NET 4.0, although the user themselves could run our library in a project that is .NET v4.5 or higher.
  - Ref: https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-6.0#remarks
  
Also included some basic linting on the Client class during the process of adding .NET version.

### DIFF
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -82,7 +82,7 @@ namespace EasyPost
                  * but we can't get the exact version.
                  * See: https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-6.0#remarks
                  */
-                dotNetVersion = "4.6+ or Core 2.2-";
+                dotNetVersion = "4.6+";
             }
 
             _dotNetVersion = dotNetVersion;
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -1,68 +1,91 @@
 ﻿using Newtonsoft.Json;
 using RestSharp;
-
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Reflection;
 
-namespace EasyPost {
-    public class Client {
-        public string version;
+namespace EasyPost
+{
+    public class Client
+    {
+        private readonly string _libraryVersion;
+        private readonly string _dotNetVersion;
+        private readonly RestClient _restClient;
+        private readonly ClientConfiguration _configuration;
+        private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion}";
 
-        internal RestClient client;
-        internal ClientConfiguration configuration;
+        private RestRequest PrepareRequest(Request request)
+        {
+            var restRequest = (RestRequest)request;
 
-        public Client(ClientConfiguration clientConfiguration) {
-            System.Net.ServicePointManager.SecurityProtocol |= Security.GetProtocol();
-            configuration = clientConfiguration ?? throw new ArgumentNullException("clientConfiguration");
-
-            client = new RestClient(clientConfiguration.ApiBase);
-
-            Assembly assembly = Assembly.GetExecutingAssembly();
-            FileVersionInfo info = FileVersionInfo.GetVersionInfo(assembly.Location);
-            version = info.FileVersion;
-        }
-
-        internal IRestResponse Execute(Request request) {
-            return client.Execute(PrepareRequest(request));
-        }
-
-        internal T Execute<T>(Request request) where T : new() {
-            RestResponse<T> response = (RestResponse<T>)client.Execute<T>(PrepareRequest(request));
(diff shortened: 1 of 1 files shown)


nwithan8:
> We can implement a more accurate solution in the future using interfaces, but for now, this will work as a first pass.

I originally tried an extended implementation that checked the registry (as [Microsoft recommends](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#query-the-registry-using-code)), but this would not work for a number of reasons.

1. The functions needed to query the registry simply don't exist in .NET 3.5, so we'd have to build an interface and add a different implementation for the .NET 3.5 version of the library. This would likely require duplicating the Client class for each version, resulting in a lot of duplicate code.
2. Checking the registry could (and probably would) fail for a variety of reasons:
  - Application is not running on a Windows machine, so there's no registry to query.
  - Application is not running with the permissions needed to query the registry.
3. The final solution is still not exact, and relies on two different methods of checking the registry, two different registry keys to check (depending on 32-bit vs. 64-bit applications), still may not account for 32-bit systems(?), and ultimately falls back to the `Environment.Version` method implemented in this PR.

EasyPost.Tests/BatchTest.cs
```diff
@@ -114,18 +114,5 @@ public class BatchTest {
 
             batch.GenerateScanForm();
         }
-
-        [TestMethod]
```
Justintime50, line 118: This PR was not rebased properly. If it was, this section of code would not be included (commits or code). Please rebase properly before we review again.

Justintime50 (changes requested):

EasyPost.Tests/BatchTest.cs
nwithan8: Corrected.

Justintime50:
Code looks fine, behavior appears as it will remain the same, various variables were renamed (internal/private items) and look to follow the coding convention: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions

Can you list out some possible version strings we'd see coming through with this change? It looks like many will simply be `4.6+ or Core 2.2-` which isn't super helpful in debugging.

nwithan8:
> Code looks fine, behavior appears as it will remain the same, various variables were renamed (internal/private items) and look to follow the coding convention: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions
> 
> Can you list out some possible version strings we'd see coming through with this change? It looks like many will simply be `4.6+ or Core 2.2-` which isn't super helpful in debugging.

For users running our library on a machine running .NET Framework 4.5 or lower (a Windows machine), here are some of the [possible version numbers](https://stackoverflow.com/a/47853580) we may receive using `System.Environment.Version`

Unfortunately, `System.Environment.Version` [will report `4.0.30319.42000`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-6.0#remarks) on any .NET Framework version 4.6 or higher.

As for .NET Core (the cross-platform .NET), versions 2.2 and less [will report will also report `4.0.30319.42000`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-6.0#remarks), so there is no way to discern whether the user is running .NET Framework 4.6+ or .NET Core 2.2-. This has been rectified in [.NET Core 3.1](https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-6.0#:~:text=For%20.NET%20Core%203.x%20and%20.NET%205%2B%2C%20the%20Environment.Version%20property%20returns%20the%20.NET%20runtime%20version%20number.) and [.NET (Core) 5.0](https://github.com/dotnet/runtime/issues/12124#issuecomment-489805891).

With our planned library support (after dropping .NET Core 2.0 and .NET Framework 4.0):
- Someone running our .NET Core 3.1 library requires their system to be minimum .NET Core 3.1, which will report a correct version number. This also applies to our eventual .NET (Core) 5.0 support.
- Someone running our .NET Framework 3.5 library will report a correct version number, unless their system is .NET Framework 4.6+, in which case it will report `4.0.30319.42000`

Justintime50 (approved):
Nice work, thanks for all the clarification!

jchen293 (approved):
lgtm

PR #186, opened 2022-01-20
CI tests

### LINKED ISSUE #154: Automated tests
We currently don't automatically run any tests or do any kind of CI in this client library.

Let's get easypost-csharp working on .NET Core and then add GitHub Actions to run the tests (or at least run a compile). Eventually it'd be great to automate package building and NuGet  upload, too...

nwithan8:
Use GitHub Actions to run CI tests on all the .NET and .NET Core versions we support.

This will attempt to build and run tests on a per-version basis. For each version (i.e. .NET Core 3.1), this will:
- Restore the required NuGet packages
- Build the specific version of the client library and its corresponding test suite version
- Run the test suite against the specific version

This PR also introduces dedicated test suites (i.e. `EasyPost.NetCore31` has a corresponding `EasyPost.Tests.NetCore31` project). All these test suites simply import the global tests (`EasyPost.Tests`), similar to how the library suites work (i.e. `EasyPost.NetCore31` is really just a bunch of links to files in the `EasyPost` project). There is no code duplication here; the different test projects are required because they target different .NET/.NET Core versions to ensure the tests are performed using the correct platform.
- This also makes it easier to perform tests on any version, since there is an explicit corresponding test suite to execute, without the user having to potentially edit the `TargetFramework` information in a project's `.csproj` file to run a test.

Closes #154
Closes #155


nwithan8:
It works, please don't touch it until we add another supported .NET/.NET Core version 🥲

nwithan8:
Closes #155

nwithan8:
Closes #154

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+    strategy:
+      matrix:
+        dotnet-version: [ '35', '40' ]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Setup variables
+        id: test_file
+        run: echo "::set-output name=test_file::EasyPost.Tests.Net${{ matrix.dotnet-version }}"
+      - name: Setup MSBuild
+        if: always()
```
Justintime50, line 21: Do we need the `if: always()` bit here? This should automatically run anyway and shouldn't fail right? We trust Microsoft's action to work every time?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: microsoft/setup-msbuild@v1.1
+      - name: Setup Nuget
+        uses: NuGet/setup-nuget@v1.0.5
```
Justintime50, line 24: # Reviewer Note

This action does not setup the other various Visual Studio tools which is why we need all the separate actions here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Setup VSTest
+        uses: darenm/Setup-VSTest@v1
```
Justintime50, line 26: What does this action do? It's not clear from its README what its purpose is. Looks like below there is a command invoked from this action?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Navigate to project root
```
Justintime50, line 27: GitHub workflows should already be at the project root, AKA the `GITHUB_WORKSPACE` by default? Is there a reason we are changing directories here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: cd $GITHUB_WORKSPACE
+      - name: Restore NuGet Packages
+        run: nuget restore EasyPost.sln
+      - name: Build Solution
+        run: msbuild ${{ steps.test_file.outputs.test_file }}\${{ steps.test_file.outputs.test_file }}.csproj /p:platform="Any CPU" /p:configuration="Test" /p:outputPath="bin/Test" /p:target="Rebuild"
+      - name: Run Tests
+        run: vstest.console.exe ${{ steps.test_file.outputs.test_file }}\bin\Test\${{ steps.test_file.outputs.test_file }}.dll
+        # delete the current DLL to guarantee that the next test will generate and use a new one
```
Justintime50, line 35: Thanks for this inline comment, helps clarify `why` we are doing what we are doing here.

As for using the `JesseTG/rm` action, is there a reason we are pulling in a dependency to effectively do the following?

```
rm -rf ${{ steps.test_file.outputs.test_file }}\bin\Test\${{ steps.test_file.outputs.test_file
```

If we can simply call the `rm` command, I'd rather do that then depending on an entire dependency to do it for us.

EasyPost.Tests.Net35/EasyPost.Tests.Net35.csproj
```diff
@@ -0,0 +1,111 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFramework>net452</TargetFramework>
+
+        <IsPackable>false</IsPackable>
```
Justintime50, line 5: Will this keep the tests from getting tossed in with the rest of the files then? Do we do this on our base test files?

EasyPost.Tests/EasyPost.Tests.csproj
```diff
@@ -1,20 +1,25 @@
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <TargetFramework>net452</TargetFramework>
```
Justintime50, line 3: `net452` doesn't match up with the below `Net35`?

Justintime50 (changes requested):
Overall this is good, though I think there are opportunities to clean it up and have less moving parts.

A few questions to clarify and for my own personal learning. If I left a comment on the Action piece of the code, it applies to both jobs (`NET_Tests` and `NET_Core_Tests`)

.github/workflows/ci.yml
nwithan8: We actually don't need to remove the file at all anymore, since these all run on different runners.

EasyPost.Tests.Net35/EasyPost.Tests.Net35.csproj
nwithan8: What are you referring to here?

.github/workflows/ci.yml
nwithan8: Was simply following the tutorial I found, but you're right, I don't think we ever leave, so this step might be unnecessary.

.github/workflows/ci.yml
nwithan8: Setup Nuget installs nuget.exe, which is needed during the `nuget restore` action later.

.github/workflows/ci.yml
nwithan8: No, this is a hold-over from a previous version of the script. Will remove.

EasyPost.Tests.Net35/EasyPost.Tests.Net35.csproj
Justintime50: Basically, this line is going to exclude tests from the packaged nuget library right?

nwithan8: Correct.

.github/workflows/ci.yml
Justintime50: My question above was attached to the `darenm/Setup-VSTest@` line, what does that one do?

nwithan8: Ah, yes, this imports `vstest.console.exe` (Visual Studio's CLI testing program), which is used to execute the test DLLs later.

.github/workflows/ci.yml
nwithan8: This is to ensure that, i.e. the 4.0 runner continues to run even if the 3.5 runner fails for whatever reason.

Justintime50: To keep our build hours low, we may not want to run all jobs if one fails because we know we'll need to make changes anyway to get it working. Always running every job could eventually become costly. Thoughts?

fiendish:
> Closes #155
> Closes #154

@nwithan8 These tags have to be added in the original PR comment, not latter comments for them to work. I've edited the original message to include them.

.github/workflows/ci.yml
nwithan8: I removed it and, as expected, it halted later tests when one failed. My concern is, the other tests didn't get a chance to run, so we wouldn't know if just one test failed or they all failed because of something we did?
Also, sometimes the failure isn't necessarily something wrong with our code. One time, the CI failed for a platform simply because downloading the NuGet packages to set up the test stalled.

Justintime50:
@nwithan8 looks like a test failed (not due to the removal of the `always()` bit, but an actual transient test failure itself. You could try re-running the CI workflow and see if that clears it up or we may need to tweak an assertion (fee perhaps)?

nwithan8:
> @nwithan8 looks like a test failed (not due to the removal of the `always()` bit, but an actual transient test failure itself. You could try re-running the CI workflow and see if that clears it up or we may need to tweak an assertion (fee perhaps)?

Rerunning the test passed this time, so probably a one-time thing with that particular test.

Justintime50:
Perfect, thanks for re-running them! These transient failures won't be a problem once we introduce VCR.

Justintime50 (approved):

Justintime50 (approved):
Love the consolidated CI jobs!

jchen293 (approved):
lgtm, thanks for combing two workflows into one.

Justintime50:
@nwithan8 while we're here, let's also add the GitHub CI badge to our README like we've done on our other libraries so we have a clear indicator of the status of our CI runs.

PR #187, opened 2022-01-25
Add docstrings

nwithan8:
Most of our methods were documented, but some were missing docstrings and/or had formatting errors. Those have been corrected. Only public methods (those that an end-user would need to call) have been documented, which can be leveraged by doc generation or IDE code tips.

- Added docstrings to all public class methods.
- Fixed spelling errors
- Standardized verbiage and formatting


EasyPost/Client.cs
```diff
@@ -1,91 +1,89 @@
 ﻿using Newtonsoft.Json;
```
Justintime50, line 1: You need to rebase so the user-agent changes aren't included in this PR.

EasyPost/Order.cs
```diff
@@ -89,7 +89,7 @@ public void GetRates()
 
             rates = request.Execute<Order>().rates;
         }
-
+
```
Justintime50, line 92: Let's not include whitespaces.

EasyPost/Request.cs
```diff
@@ -7,7 +7,7 @@
 namespace EasyPost {
     public class Request {
         internal RestRequest restRequest;
-
+
```
Justintime50, line 10: No extra whitespaces

Justintime50 (changes requested):
Good stuff, just a couple extra random whitespaces that need removing and it needs to be rebased. Otherwise good.

EasyPost/Client.cs
nwithan8: Patched in bc436ee

EasyPost/Order.cs
nwithan8: Patched with 90a14e5

EasyPost/Request.cs
nwithan8: Patched with 90a14e5

Justintime50 (approved):

jchen293 (approved):
lgtm

PR #189, opened 2022-01-26
Timeout for http

jchen293:
Add a 30-second default connection timeout to all requests
Add a 60-second default request timeout to all requests
Add setter and getter to allow the user to change the connection and request timeout

### DIFF
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -14,8 +14,11 @@ namespace EasyPost
         private readonly RestClient _restClient;
         private readonly ClientConfiguration _configuration;
         private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion}";
-        private int _connectTimeoutMilliseconds = 30000;
-        private int _requestTimeoutMilliseconds = 60000;
+
+        private readonly int _defaultConnectTimeoutMilliseconds = 30000;
+        private int? _connectTimeoutMilliseconds;
+        private readonly int _defaultRequestTimeoutMilliseconds = 60000;
+        private int? _requestTimeoutMilliseconds;
 
         /// <summary>
         /// Prepare a request for execution by attaching required headers.
@@ -25,7 +28,7 @@ namespace EasyPost
         private RestRequest PrepareRequest(Request request)
         {
             var restRequest = (RestRequest)request;
-            restRequest.Timeout = _requestTimeoutMilliseconds;
+            restRequest.Timeout = getRequestTimeout();
             restRequest.AddHeader("user_agent", UserAgent);
             restRequest.AddHeader("authorization", "Bearer " + _configuration.ApiKey);
             restRequest.AddHeader("content_type", "application/json");
@@ -93,7 +96,7 @@ namespace EasyPost
             _configuration = clientConfiguration ?? throw new ArgumentNullException("clientConfiguration");
 
             _restClient = new RestClient(clientConfiguration.ApiBase);
-            _restClient.Timeout = _connectTimeoutMilliseconds;
+            _restClient.Timeout = getConnectionTimeout();
 
             var assembly = Assembly.GetExecutingAssembly();
             var info = FileVersionInfo.GetVersionInfo(assembly.Location);
@@ -118,7 +121,7 @@ namespace EasyPost
         /// </summary>
         /// <returns> connection timeout in milliseconds.</returns>
         public int getConnectionTimeout(){
-            return _connectTimeoutMilliseconds;
+            return _connectTimeoutMilliseconds ?? _defaultConnectTimeoutMilliseconds;
         }
 
         /// <summary>
@@ -134,7 +137,7 @@ namespace EasyPost
         /// </summary>
         /// <returns> request timeout in milliseconds.</returns>
         public int getRequestTimeout(){
-            return _requestTimeoutMilliseconds;
+            return _requestTimeoutMilliseconds ?? _defaultRequestTimeoutMilliseconds;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Justintime50:
@jchen293 were you able to test this? Try setting the timeout to 1ms each individually and see if the request fails because of a timeout. If so, we know the timeouts are getting set properly (then you can revert them back to what they should be).

jchen293:
> @jchen293 were you able to test this? Try setting the timeout to 1ms each individually and see if the request fails because of a timeout. If so, we know the timeouts are getting set properly (then you can revert them back to what they should be).

@Justintime50 
Yes, I was able to test this by setting the timeout to less than 100ms.

Justintime50 (approved):

nwithan8 (approved):

Justintime50:
@jchen293 you'll need to rebase to fix the conflicts on the CHANGELOG.

Justintime50 (changes requested):
I apologize for realizing so late, but we'll need getters and setters for the request and connection timeouts so users can override these default values if necessary. The default work looks great, let's take it one step further.

nwithan8:
> I apologize for realizing so late, but we'll need getters and setters for the request and connection timeouts so users can override these default values if necessary. The default work looks great, let's take it one step further.

Do we want this? We don't have this in the Java CL.

jchen293:
> > I apologize for realizing so late, but we'll need getters and setters for the request and connection timeouts so users can override these default values if necessary. The default work looks great, let's take it one step further.
> 
> Do we want this? We don't have this in the Java CL.

We just created a ticket

EasyPost/Client.cs
```diff
@@ -108,5 +112,35 @@ public Client(ClientConfiguration clientConfiguration)
...
+        public int getConnectionTimeout(){
+            return _defaultConnectTimeoutMilliseconds;
+        }
+
+        /// <summary>
+        /// Set the connection timeout with a time in milliseconds.
+        /// </summary>
+        /// <param name="connectionTimeoutMilliseconds">connection time in milliseconds.</param>
+        public void setConnectionTimeout(int connectionTimeoutMilliseconds){
+            _defaultConnectTimeoutMilliseconds = connectionTimeoutMilliseconds;
```
Justintime50, line 128: We probably shouldn't set the default connection timeout to something the user sets because it then no longer is the "default", instead (if possible), we should use the user's value only if present and if not, fallback to the default.

EasyPost/Client.cs
```diff
@@ -108,5 +112,35 @@ public Client(ClientConfiguration clientConfiguration)
 
             _dotNetVersion = dotNetVersion;
         }
+
+        /// <summary>
+        /// Return the request time from the client object.
+        /// </summary>
+        public int getConnectionTimeout(){
+            return _defaultConnectTimeoutMilliseconds;
```
Justintime50, line 120: We should probably have an intermediate variable called something like `connectionTimeoutMilliseconds` which is either the user's input or the fallback of the default. If we only ever return the default, this getter will never pick up on the user's input.

Justintime50 (changes requested):

EasyPost/Client.cs
jchen293: Thanks for the feedback, I will rename those two defaults variables

EasyPost/Client.cs
```diff
@@ -14,6 +14,8 @@ public class Client
         private readonly RestClient _restClient;
         private readonly ClientConfiguration _configuration;
         private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion}";
+        private int _connectTimeoutMilliseconds = 30000;
```
nwithan8, line 17:
```suggestion
private int _defaultTimeoutMilliseconds = 3000;
private int _connectTimeoutMillisecond;
```

EasyPost/Client.cs
```diff
@@ -108,5 +112,37 @@ public Client(ClientConfiguration clientConfiguration)
 
             _dotNetVersion = dotNetVersion;
         }
+
+        /// <summary>
+        /// Return the request time from the client object.
+        /// </summary>
+        /// <returns> connection timeout in milliseconds.</returns>
+        public int getConnectionTimeout(){
+            return _connectTimeoutMilliseconds;
```
nwithan8, line 121:
```suggestion
return _connectTimeoutMillseconds ?: _defaultTimeoutMilliseconds;
```

nwithan8: Oops, that's Groovy syntax, my bad. You'll have to add _connectTimeoutMilliseconds in between `?` and `:`

Justintime50: +1 this is along the lines of what I was thinking of. We need to retain the `default` variables and keep them internal and unchanging. We also need to have a variable that's public that users can set to override the default values for timeouts.

Justintime50 (changes requested):

EasyPost/Client.cs
nwithan8: I'm thinking we can do this with some fancy getter/setter magic.

jchen293: I was following how our other library is set up like PHP [here](https://github.com/EasyPost/easypost-php/blob/master/lib/EasyPost/EasyPost.php#L148) and try to avoid additional variables. I will make some changes shortly and add an unit test

EasyPost/Client.cs
```diff
@@ -14,8 +14,11 @@ public class Client
         private readonly RestClient _restClient;
         private readonly ClientConfiguration _configuration;
         private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion}";
-        private int _connectTimeoutMilliseconds = 30000;
-        private int _requestTimeoutMilliseconds = 60000;
+
+        private readonly int _defaultConnectTimeoutMilliseconds = 30000;
+        private int? _connectTimeoutMilliseconds;
```
Justintime50, line 19: By keeping this as a private variable, it means it can only be accessed through the getter/setter correct? Users couldn't access this directly?

Justintime50 (approved):
Great work here, thanks for adding the getters and setters! The unit test is also super helpful.

nwithan8 (approved):

PR #193, opened 2022-01-28
improve the CI build.

jchen293:
Improve the CI build by adding `-restore` in the `msbuild` step because `project.assets.json` could be not found after the NuGet restore
Ignore two unit tests that have multiple live API operations which can lead to timeout or unit test failed in the CI test, this issue will be resolved once we add VCR in the unit test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   # derived from https://dev.to/felipetofoli/github-actions-for-net-full-framework-build-and-test-299h
   NET_Tests:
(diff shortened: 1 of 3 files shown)


Justintime50:
Can we fix the flakey test as a part of this PR? Either tweak the assertion to something less transient or let's skip the test for now so it constantly passes and we can come back to this test later.

Justintime50 (approved):
Consistently passing builds make me happy. Thanks for taking the time to look into this!

nwithan8 (approved):
Great job Jack!

PR #197, opened 2022-02-02
WIP: Use built-in setters and getters

nwithan8:
- GetRates() and GetSmartrates refactored to be consistent, renamed as GenerateRates() and GenerateSmartrates() since functions don't return anything directly, but load data

### DIFF
--- a/EasyPost.Tests/OrderTest.cs
+++ b/EasyPost.Tests/OrderTest.cs
@@ -184,7 +184,7 @@ namespace EasyPost.Tests
         {
             Order order = Order.Create(parameters);
             List<Rate> old = order.rates;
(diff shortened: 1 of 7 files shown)


Justintime50 (changes requested):
I don't want to make this change (though it may be the more appropriate language) because we have the pattern of "get rates" and "get smartrates" across all our libraries and each of their interfaces are named as such. We also had blog posts and the such go out detailing how you could now call "GetSmartrates" to use this new feature (big PR thing) and I wouldn't want to yank that so soon after.

Once we've chosen an interface, we should stick with it and only tweak things behind the scenes which is why naming is hard yet important.

PR #198, opened 2022-02-02
Add support for retrieving all addresses

nwithan8:
Finally, the C# library supports the ability to get all addresses with `Address.All()`

This PR includes:
- The new `Address.All()` function to retrieve all addresses. Can optionally pass in a `Dictionary<string, object>` as parameter. Function returns an `AddressCollection` object.
- Added new `AddressCollection` class, which features:
  - `public List<Batch> addresses`
  - `public bool has_more`
  This is similar to how our other client libraries handle multiple addresses.
- A new `AddressTest.TestRetrieveAll()` test for this feature

Classes, functions and tests have been imported/linked to all projects/versions, and all tests pass.

### DIFF
--- a/EasyPost.Tests/AddressTest.cs
+++ b/EasyPost.Tests/AddressTest.cs
@@ -247,6 +247,18 @@ namespace EasyPost.Tests
         public void TestRetrieveInvalidId() => Address.Retrieve("not-an-id");
 
         [TestMethod]
+        public void TestRetrieveAll()
+        {
+            AddressCollection addressCollection = Address.All(new Dictionary<string, object>());
+            Assert.IsNotNull(addressCollection);
+            if (addressCollection.addresses.Count > 0)
+            {
+                Assert.IsNotNull(addressCollection.addresses[0].id);
+                Assert.AreEqual(addressCollection.addresses[0].id.Substring(0, 4), "adr_");
+            }
+        }
+
+        [TestMethod]
         public void TestVerificationFailure()
         {
             Address address = new Address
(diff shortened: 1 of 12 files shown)


jchen293 (approved):

jchen293:
The code looks good. Can you update the changelog as well?

EasyPost.Tests/AddressTest.cs
```diff
@@ -246,6 +246,18 @@ public void TestInstanceCreateWithVerifications()
         [ExpectedException(typeof(HttpException))]
         public void TestRetrieveInvalidId() => Address.Retrieve("not-an-id");
 
+        [TestMethod]
+        public void TestRetrieveAll()
+        {
+            AddressCollection addressCollection = Address.All();
+            Assert.IsNotNull(addressCollection);
+            if (addressCollection.addresses.Count > 0)
```
Justintime50, line 254: This is fine, could also iterate the returned list and assert against each

Justintime50 (approved):

Justintime50 (changes requested):
Actually I'm going to take another look first thing tomorrow. I think there's more needed here.

EasyPost.Tests/AddressTest.cs
nwithan8: Is this blocking the merge?

Justintime50: Nope, if I stamp it means we can proceed but any feedback regardless of who left it is highly encouraged to include 😉 I won't stop progress for little nit picks though. Some of these are learnings we can share as a team for the future.

Justintime50 (approved):

EasyPost.Tests/AddressTest.cs
nwithan8: Resolved in 7e048a5

Justintime50 (approved):

jchen293 (approved):

PR #202, opened 2022-02-03
Add support for re-rating a shipment

nwithan8:
Finally, the C# library supports the ability to re-rate a shipment with `shipment.RefreshRates()`

This PR includes:
- The new `RefreshRates()` method to refresh the rates for a given shipment. Can optionally pass in a `Dictionary<string, object>` as parameter. Function replaces existing `rates` on shipment object, does not return anything.
- Added new `ResourceNotCreated` exception (extends standard `Exception`). This is used in the tests for this method (and can hopefully be used in the future). Thrown if the resource has not been created server-side.
- `RefreshRates()` throws `ResourceNotCreated` if the shipment object has not been created server-side (the shipment's `ID` is `null`)
- New `ShipmentTest.TestRerateShipment()` and `ShipmentTest.TestRerateShipmentWithoutCreate()` tests for this feature

All tests pass.

### DIFF
--- a/EasyPost/Exception.cs
+++ b/EasyPost/Exception.cs
@@ -52,12 +52,4 @@ namespace EasyPost
     public class ClientNotConfigured : Exception
     {
     }
-
-    [Serializable]
-    public class ResourceNotCreated : Exception
-    {
-        public ResourceNotCreated(string message) : base(message)
-        {
-        }
-    }
 }
--- a/EasyPost/Exception.cs
+++ b/EasyPost/Exception.cs
@@ -52,4 +52,12 @@ namespace EasyPost
     public class ClientNotConfigured : Exception
     {
     }
+
+    [Serializable]
+    public class ResourceNotCreated : Exception
+    {
+        public ResourceNotCreated(string message) : base(message)
+        {
+        }
+    }
 }
--- a/EasyPost/Shipment.cs
+++ b/EasyPost/Shipment.cs
@@ -135,6 +135,28 @@ namespace EasyPost
         }
 
         /// <summary>
+        ///     Refresh the rates for this Shipment.
+        /// </summary>
+        /// <param name="parameters">Optional dictionary of parameters for the API request.</param>
+        public void RefreshRates(Dictionary<string, object> parameters = null)
+        {
+            if (id == null)
+            {
+                throw new ResourceNotCreated("Shipment must be created before refreshing rates. Try running `shipment.Create()` first.");
+            }
+
+            Request request = new Request("shipments/{id}/rerate", Method.POST);
+            request.AddUrlSegment("id", id);
+            if (parameters != null)
+            {
+                request.AddBody(parameters);
+            }
+
+            Shipment _shipment = request.Execute<Shipment>();
+            rates = _shipment.rates;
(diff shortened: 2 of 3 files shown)


EasyPost/Shipment.cs
```diff
@@ -134,6 +134,29 @@ public void GetRates()
             rates = request.Execute<Shipment>().rates;
         }
 
+        /// <summary>
+        ///     Refresh the rates for this Shipment.
+        /// </summary>
+        /// <param name="parameters">Optional dictionary of parameters for the API request.</param>
+        /// <exception cref="ResourceAlreadyCreated">Shipment has not be created server-side yet.</exception>
+        public void RefreshRates(Dictionary<string, object> parameters = null)
```
Justintime50, line 142: Let's call this `RegenerateRates`. We unfortunately do not have uniformity of names for this method across libraries, but the majority of them use `regenerate` so let's try to keep it as close as possible.

EasyPost/Shipment.cs
```diff
@@ -134,6 +134,29 @@ public void GetRates()
...
+        {
+            if (id == null)
+            {
+                throw new ResourceNotCreated("Shipment must be created before refreshing rates. Try running `shipment.Create()` first.");
```
Justintime50, line 146: Nice touch, though it looks like the what we do on other methods is simply create the shipment for them via `Create();`, could we do that here to stay consistent?

Justintime50 (changes requested):

EasyPost/Shipment.cs
```diff
@@ -134,6 +134,29 @@ public void GetRates()
...
+            }
+
+            Request request = new Request("shipments/{id}/rerate", Method.POST);
+            request.AddUrlSegment("id", id);
+            if (parameters != null)
+            {
+                request.AddBody(parameters);
+            }
+
+            Shipment _shipment = request.Execute<Shipment>();
```
jchen293, line 156: This variable is not declared as private, why does it start underscore?

nwithan8: Oh, that's just something I do for throwaway variables. C# complains about it though. Removed.

EasyPost/Shipment.cs
nwithan8: Resolved in 71c3033

EasyPost/Shipment.cs
nwithan8: Resolved in 71c3033

EasyPost/Exception.cs
```diff
@@ -52,4 +52,12 @@ public class ResourceAlreadyCreated : Exception
     public class ClientNotConfigured : Exception
     {
     }
+
+    [Serializable]
+    public class ResourceNotCreated : Exception
```
jchen293, line 57: Is this being used in other places? Correct me if I am wrong.

Justintime50: Yeah, let's remove this bit if it's not in use.

Justintime50 (changes requested):

Justintime50 (approved):

jchen293 (approved):

PR #203, opened 2022-02-04
Add support for updating user's brand

nwithan8:
Finally, the C# library supports the ability to get update your brand with `user.UpdateBrand()`

This PR includes:
- The new `user.UpdateBrand()` function to update the associated user's brand. Pass in a `Dictionary<string, object>` as parameter. Function returns a `Brand` object.
- Added new `Brand` class, which features all necessary properties.
- A new `UserTest.TestUpdateBrand()` test for this feature

Classes, functions and tests have been imported/linked to all projects/versions, and all tests pass.

### DIFF
--- a/EasyPost.Tests/UserTest.cs
+++ b/EasyPost.Tests/UserTest.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using System.Collections.Generic;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace EasyPost.Tests
 {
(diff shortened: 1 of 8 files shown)


EasyPost.Tests/UserTest.cs
```diff
@@ -18,6 +19,24 @@ public void TestRetrieveSelf()
             Assert.AreEqual(user.id, user2.id);
         }
 
+        [TestMethod]
+        public void TestUpdateBrand()
+        {
+            User user = User.Retrieve();
+            Assert.IsNotNull(user.id);
+
+            const string color = "#000000";
```
Justintime50, line 28: Let's use a color that's super random so we know this went through (AKA not a default white or black).

EasyPost.Tests/UserTest.cs
```diff
@@ -18,6 +19,24 @@ public void TestRetrieveSelf()
...
+            Dictionary<string, object> details = new Dictionary<string, object>()
+            {
+                { "color", color }
+            };
+            Brand brand = user.UpdateBrand(parameters: details);
+            Assert.IsNotNull(brand);
+            Assert.IsNotNull(brand.user_id);
```
Justintime50, line 35: This second assertion is inherently useless as we already asserted the object isn't null - objects will always come back with IDs on them.

Justintime50 (approved):

EasyPost.Tests/UserTest.cs
nwithan8: Resolved in [8d459a7](https://github.com/EasyPost/easypost-csharp/pull/203/commits/8d459a71ab9649ee007be4f7891e1ca9c06b371e)

EasyPost.Tests/UserTest.cs
nwithan8: Resolved in [8d459a7](https://github.com/EasyPost/easypost-csharp/pull/203/commits/8d459a71ab9649ee007be4f7891e1ca9c06b371e)

Justintime50 (approved):

PR #205, opened 2022-02-04
Add support for Insurance objects

nwithan8:
Finally, our C# library supports Insurance!

This PR includes:
- A new `Insurance` class, with `Create()`, `Retrieve()`, `Refresh()` and `All()` functions
- A new `InsuranceCollection` class required for the `Insurance.All()` function
- Tests for all functionalities

### DIFF
--- a/EasyPost.Tests/InsuranceTest.cs
+++ b/EasyPost.Tests/InsuranceTest.cs
@@ -1,0 +1,128 @@
+using System.Collections.Generic;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace EasyPost.Tests
+{
+    [TestClass]
+    public class InsuranceTest
+    {
+        private Insurance _insurance { get; set; }
+        private Dictionary<string, object> InsuranceData { get; set; }
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            ClientManager.SetCurrent("NvBX2hFF44SVvTPtYjF0zQ");
+            Address fromAddress = Address.Create(new Dictionary<string, object>
+            {
+                {
+                    "company", "Simpler Postage Inc"
+                },
+                {
+                    "street1", "164 Townsend St"
+                },
+                {
+                    "street2", "Unit 1"
+                },
+                {
+                    "city", "San Francisco"
+                },
+                {
+                    "state", "CA"
+                },
+                {
+                    "zip", "94107"
+                },
+                {
+                    "country", "US"
+                }
+            });
+            Address toAddress = Address.Create(new Dictionary<string, object>()
+            {
+                {
+                    "company", "The White House"
+                },
+                {
+                    "street1", "1600 Pennsylvania Avenue NW"
+                },
+                {
+                    "street2", "DC 20500"
+                },
+                {
+                    "city", "Washington"
+                },
+                {
+                    "state", "DC"
+                },
+                {
+                    "zip", "20500"
+                },
+                {
+                    "country", "US"
+                }
+            });
+            Tracker tracker = Tracker.Create(trackingCode: "EZ1000000001", carrier: "USPS");
+
+            InsuranceData = new Dictionary<string, object>
+            {
+                {
+                    "to_address", toAddress
+                },
+                {
+                    "from_address", fromAddress
+                },
+                {
+                    "tracker", tracker
+                },
+                {
+                    "amount", (float)100
+                },
+                {
+                    "currency", "USD"
+                }
+            };
+            _insurance = new Insurance
+            {
+                from_address = toAddress,
+                to_address = fromAddress,
+                tracker = tracker,
+                amount = (float)100.00,
+                provider = "USPS"
+            };
+        }
+
+        [TestMethod]
+        public void TestCreateAndRetrieve()
+        {
+            Insurance insurance = Insurance.Create(InsuranceData);
+            Assert.IsNotNull(insurance.id);
+            Assert.AreEqual(insurance.from_address, _insurance.from_address);
+
+            Insurance retrieved = Insurance.Retrieve(insurance.id);
+            Assert.AreEqual(insurance.id, retrieved.id);
+        }
+
+        [TestMethod]
+        public void TestCreateInstance()
+        {
+            _insurance.Create();
+            Assert.IsNotNull(_insurance.id);
+        }
+
+        [TestMethod]
+        public void TestRetrieveAll()
+        {
+            InsuranceCollection insuranceCollection = Insurance.All();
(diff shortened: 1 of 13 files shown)


EasyPost.Tests/InsuranceTest.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace EasyPost.Tests
+{
+    [TestClass]
+    public class InsuranceTest
+    {
+        private Dictionary<string, object> _fromAddressData { get; set; }
+        private Insurance _insurance { get; set; }
+        private Dictionary<string, object> _parcelData { get; set; }
+        private Dictionary<string, object> _toAddressData { get; set; }
```
Justintime50, line 12: Does the linter want these ordered alphabetically? If not, it'd make more sense to have from/to address next to each other, though I assume they're this way for the linter and that's fine.

EasyPost.Tests/InsuranceTest.cs
```diff
@@ -0,0 +1,145 @@
...
+                    "amount", (float)100
+                },
+                {
+                    "carrier", "USPS"
+                }
+            };
+        }
+
+        [TestMethod]
+        public void TestCreateAndRetrieve()
```
Justintime50, line 119: Longterm, tests should only do one thing (test one thing) - We can leave this for now but when we circle back, there should be a standalone `create` test, a standalone `retrieve` test (which may still need to create an object, but it shouldn't assert against the created one, only the retrieved one), and then an `all` test aside.

EasyPost/Tracker.cs
```diff
@@ -102,7 +102,12 @@ public static Tracker Retrieve(string id)
         public static bool CreateList(Dictionary<string, object> param)
         {
             Request request = new Request("trackers/create_list", RestSharp.Method.POST);
-            request.AddBody(new Dictionary<string, object>() { { "trackers", param } });
+            request.AddBody(new Dictionary<string, object>
+            {
+                {
+                    "trackers", param
```
Justintime50, line 108: `param` should be pluralized.

Justintime50 (approved):

EasyPost.Tests/InsuranceTest.cs
nwithan8: Almost every one of our test files has a `CreateAndRetrieve` method, so we'll need to rework that for the entire library and all tests.

EasyPost/Tracker.cs
nwithan8: Resolved in [e01e5b4](https://github.com/EasyPost/easypost-csharp/pull/205/commits/e01e5b4a785997e20d948c2c0fccb4d2a0f0747b)

EasyPost.Tests/InsuranceTest.cs
```diff
@@ -0,0 +1,145 @@
...
+namespace EasyPost.Tests
+{
+    [TestClass]
+    public class InsuranceTest
+    {
+        private Dictionary<string, object> _fromAddressData { get; set; }
+        private Insurance _insurance { get; set; }
+        private Dictionary<string, object> _parcelData { get; set; }
+        private Dictionary<string, object> _toAddressData { get; set; }
+        private Dictionary<string, object> InsuranceData { get; set; }
```
jchen293, line 13: **InsuranceData** is a private variable, should the variable start with an underscore and lowercase?

EasyPost.Tests/InsuranceTest.cs
```diff
@@ -0,0 +1,145 @@
...
+                    "to_address", _toAddressData
+                },
+                {
+                    "from_address", _fromAddressData
+                },
+                {
+                    "tracking_code", shipment.tracking_code
+                },
+                {
+                    "amount", (float)100
```
jchen293, line 110: Why are we converting **int** to **float** here? Can we just define it as float?

nwithan8: Without the cast, it thinks it's an `int`. With decimals, it thinks it's a `double`

EasyPost.Tests/InsuranceTest.cs
nwithan8: Good point, resolved in [76da6a1](https://github.com/EasyPost/easypost-csharp/pull/205/commits/76da6a1c7800ac4df98f8c230458ee78540999bc)

EasyPost.Tests/InsuranceTest.cs
nwithan8: Resolved by using `decimal` type in [76da6a1](https://github.com/EasyPost/easypost-csharp/pull/205/commits/76da6a1c7800ac4df98f8c230458ee78540999bc)

EasyPost.Tests/InsuranceTest.cs
jchen293: Why are the private variables all start with uppercase now?

I am following the naming convention from the Microsoft doc: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions

EasyPost/Insurance.cs
```diff
@@ -0,0 +1,116 @@
+using System.Collections.Generic;
+using RestSharp;
+
+namespace EasyPost
+{
+    public class Insurance : Resource
+    {
+        public decimal amount { get; set; }
```
Justintime50, line 8: This is a string (we use string fields almost exclusively with a few exceptions).

Justintime50 (changes requested):

EasyPost.Tests/InsuranceTest.cs
nwithan8: These are properties (setters and getters), not fields.

EasyPost/Insurance.cs
nwithan8: In the Java library, which I modeled this after, `amount` is a number, not a string: https://github.com/EasyPost/easypost-java/blob/master/src/main/java/com/easypost/model/Insurance.java#L22

EasyPost.Tests/InsuranceTest.cs
jchen293: thanks for the clarification.

EasyPost/Insurance.cs
Justintime50: Unfortunately the API returns a string even if it was setup wrong in Java.

nwithan8: Resolved in [f696919](https://github.com/EasyPost/easypost-csharp/pull/205/commits/f696919c16d57f48faeb4f74bedd73f684112b54)

Justintime50 (approved):

jchen293 (approved):

PR #206, opened 2022-02-08
Fix `GetRates()` function

nwithan8:
This PR:
- Removes the `GetRates()` method from the `Shipment` class, as rates are loaded when the shipment is created server-side. Use `shipment.rates` to retrieve loaded rates. Use `shipment.RegenerateRates()` to refresh rates.

### DIFF
--- a/EasyPost.Net/Order.cs
+++ b/EasyPost.Net/Order.cs
@@ -85,7 +85,7 @@ namespace EasyPost
         /// <summary>
         ///     Populate the rates property for this Order.
         /// </summary>
-        public void RegenerateRates(Dictionary<string, object> parameters = null)
+        public void GetRates()
         {
             if (id == null)
             {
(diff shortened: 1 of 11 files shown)


Justintime50:
Let's leave this in draft mode just until we've done the minor version bump on this library. Since this is a breaking change, we'll need to include this in our 2nd batch of updates to follow shortly.

Justintime50:
@nwithan8 let's rebase this and mark it ready for review.

Justintime50:
@nwithan8 can we fix tests here?

nwithan8:
Technically, this is running against the new tests, rather than the old test suite.

EasyPost.Net/Order.cs
```diff
@@ -85,7 +85,7 @@ public void Create()
         /// <summary>
         ///     Populate the rates property for this Order.
         /// </summary>
-        public void GetRates()
+        public void RegenerateRates(Dictionary<string, object> parameters = null)
```
Justintime50, line 88: You can't regenerate rates for an order, only a shipment. You'll want to put this back to what it was in both locations.

Justintime50 (changes requested):

nwithan8:
It seems this patch accidentally snuck itself into `master` via another PR (likely the test suite). This PR is no longer necessary.

PR #207, opened 2022-02-08
remove Rating class.

jchen293:
This PR removes the **Rating** class because the old **` rating/v1/rates`** endpoint isn't documented and only supported on this library via the Rating.Create() function. Rating class is currently the only extra class that we have across our seven client libraries. This is going to be a breaking change if any user is still using this outdated endpoint.


Justintime50 (approved):

nwithan8 (approved):

Justintime50:
@jchen293 please rebase this and ping me when it's ready to merge.

PR #209, opened 2022-02-10
Add/drop version support, restructure parent folders

nwithan8:
EDIT: Preserving the original PR description as-is below for educational purposes, but this PR has changed course slightly since it was first opened.

Per internal discussions, addressing version support is necessary before we can move forward with other improvements to this library (i.e. RestSharp upgrade, adding VCR support). The time has come for us to change what versions we support.

This PR:
- **Drop support for the deprecated .NET (Framework) 4.0.** The only remaining explicit version of the legacy .NET Framework we support is .NET (Framework) 3.5, which is currently in LTS. Those using newer versions of .NET (Framework) will be able to utilize the .NET (Framework) 3.5 version of this library with no issue.
- **Drop support for the deprecated .NET Core 2.0**
- **Adds support for .NET 5.0**
- **Adds support for .NET 6.0**
- **Adds test suites for all versions**

As a result of these additions and drops, we now support .NET Core 3.1, .NET 5.0 and .NET 6.0, which are part of the modern, cross-platform .NET/.NET Core framework, and .NET Framework 3.5, which is part of the legacy, Windows-only .NET Framework framework.

Future development tasks, such as upgrading RestSharp to v107, can only be done on the .NET/.NET Core versions due to both support and changing syntax. This PR introduces a split-parent folder structure, creating two groups of versions.

`EasyPost.Net` targets .NET Core 3.1, .NET 5.0 and .NET 6.0, while `EasyPost.NetFramework` targets .NET Framework 3.5. This restructure allows us to only have two "projects", rather than a single project for each version that merely import parent files anyway. In short, this is much cleaner.

While `EasyPost.Net` and `EasyPost.NetFramework` can (and some do) target multiple frameworks in one project, the test suites are still explicitly single-target, allowing us to run tests with a specific environment, i.e. `EasyPost.Net` in a .NET 5.0 environment.

For the tests, `EasyPost.Tests.NetCore31`, `EasyPost.Tests.Net50` and `EasyPost.Tests.Net60` import their files from `EasyPost.Tests.Net`, while `EasyPost.Tests.NetFramework35` imports its files from `EasyPost.Tests.NetFramework`. The reason for the two parent test projects (`EasyPost.Tests.Net`, `EasyPost.Tests.NetFramework`) is to allow us to add tests for one platform (i.e. .NET/.NET Core) that won't be available in the other platform, such as if we add a new feature that for whatever reason isn't available in our .NET Framework version.

----
ORIGINAL:

A lot of our upcoming work, including `RestSharp` v107 upgrade and VCR support, only work in our `.NET Core` versions.
Current support:
| `.NET` version       | `RestSharp` v107 | VCR (via `Scotch`) |
|--------------------|----------------|------------------|
| `.NET Framework 3.5` | N              | N                |
| `.NET Framework 4.0` | N              | N                |
| `.NET Core 2.0`      | -              | -                |
| `.NET Core 3.1`      | Y              | Y                |

Future support:
| .NET version       | RestSharp v107 | VCR (via Scotch) |
|--------------------|----------------|------------------|
| `.NET Framework 3.5` | N              | N                |
| `.NET Core 3.1`      | Y              | Y                |
| `.NET 5.0`           | Y              | Y                |
| `.NET 6.0`           | Y              | Y                |

As you can see, currently the only library compatible with these planned upgrades are our `.NET Core 3.1` version (`.NET Core 2.0` is also compatible, but we will be dropping support for that version very soon; we don't even have a test suite for `.NET Core 2.0` established).

For the time being, this PR merely separates out the `.NET Core 3.1` files on their own. Normally, the actual file for any given class or test would exist in one parent directory (`EasyPost` or `EasyPost.Tests`), and simply be linked into the various individual version projects. Because there are actual differences in the code files (i.e. `RestSharp` v107 uses different cases for its enums, VCR requires annotations), we can't simply update the parent file without breaking compatibility for the other library versions.

Instead, the `.NET Core 3.1` project features its own copies of the parent files. Currently, there are no changes made to these files, but future work will be able to modify these files specifically without interfering with the other versions (which still all utilize the parent files). This does not affect the functionality of the code at all, but does introduce unfortunate code duplication (see below).

**All PRs regarding `RestSharp` v107 and VCR support will be based on this branch until this branch is merged into `master.`**

Only one of our four versions currently reference these new files. Following our eventual version support addition/removals, three of our four versions will reference these files (our `.NET Framework 3.5` version will then be the only one referencing files different from the other versions).

Going forward, due to the incompatibilities that will be created by the `RestSharp` v107 and VCR upgrades, we will have **two** different versions of each class file, one supporting `.NET Framework` (`3.5` always, `4.0` for now) and one supporting `.NET/.NET Core` (`3.1` always, `5.0` and `6.0` eventually)

During the version swapping, I would advise we clearly indicate the two parent projects (`EasyPost.NetFramework` and `EasyPost.Net`), and have each version project import from the appropriate parent. For now, however, `EasyPost` continues to be the only parent. The `.NET Core 3.1`-specific versions of those files reside in `EasyPost.NetCore31`


Justintime50:
For my understanding, this PR simply duplicates the file/folder structure but makes no changes. In a PR to follow shortly, these two version sets will start to diverge based on the syntax required for each version?

EasyPost.Net/Address.cs
```diff
@@ -62,7 +62,7 @@ public void Create(List<string> verifications = null, List<string> strictVerific
         /// <summary>
         ///     Verify an address.
         /// </summary>
-        /// <returns>EasyPost.Address instance. Check message for verification failures.</returns>
+        /// <returns>EasyPost.Net.Address instance. Check message for verification failures.</returns>
```
Justintime50, line 65: I think all these return comment need to be reverted. Because of us fixing the namespace, it's still just an `EasyPost.Address` instance. This goes for every file in this PR.

Justintime50 (dismissed):

EasyPost.Net/Address.cs
nwithan8: Resolved in [ba993b2](https://github.com/EasyPost/easypost-csharp/pull/209/commits/ba993b21ecde5d188894a9bcd4f13bc8ddd00f1a)

EasyPost.Net/Event.cs
```diff
@@ -18,9 +18,9 @@ public class Event : Resource
         public DateTime? updated_at { get; set; }
 
         /// <summary>
-        ///     Resend the last Event for a specific EasyPost object instance.
+        ///     Resend the last Event for a specific EasyPost.Net object instance.
```
Justintime50, line 21: These comment update needs to be removed

EasyPost.Net/Insurance.cs
```diff
@@ -67,7 +67,7 @@ public static InsuranceCollection All(Dictionary<string, object> parameters = nu
         ///     Dictionary containing parameters to create the insurance with. Valid pairs:
         ///     * {"to_address", Address} The actual destination of the package to be insured
         ///     * {"from_address", Address} The actual origin of the package to be insured
-        ///     * {"tracking_code", string} The tracking code associated with the non-EasyPost-purchased package you'd like to insure
+        ///     * {"tracking_code", string} The tracking code associated with the non-EasyPost.Net-purchased package you'd like to insure
```
Justintime50, line 70: This needs to be reverted.

EasyPost.Net/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("EasyPost.Net")]
```
Justintime50, line 8: Does the Assembly need to simply be `EasyPost` or do these need to be this way?

Justintime50 (dismissed):

EasyPost.NetFramework/Client.cs
```diff
@@ -34,10 +34,10 @@ public int RequestTimeoutMilliseconds
             set => _requestTimeoutMilliseconds = value;
         }
 
-        private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion}";
+        private string UserAgent => $"EasyPost.NetFramework/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion}";
```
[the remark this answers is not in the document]
nwithan8: Resolved in [3b9751f](https://github.com/EasyPost/easypost-csharp/pull/209/commits/3b9751ffe837360356709749f90f25a116898726)

EasyPost.NetFramework/Event.cs
```diff
@@ -0,0 +1,73 @@
...
+        public List<string> pending_urls { get; set; }
+        public Dictionary<string, object> previous_attributes { get; set; }
+        public Dictionary<string, object> result { get; set; }
+        public string status { get; set; }
+        public DateTime? updated_at { get; set; }
+
+        /// <summary>
+        ///     Resend the last Event for a specific EasyPost.NetFramework object instance.
+        /// </summary>
+        /// <param name="id">String representing an EasyPost.NetFramework object instance.</param>
```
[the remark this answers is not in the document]
nwithan8: Resolved in [3b9751f](https://github.com/EasyPost/easypost-csharp/pull/209/commits/3b9751ffe837360356709749f90f25a116898726)

EasyPost.Net/Event.cs
nwithan8: Resolved in [3b9751f](https://github.com/EasyPost/easypost-csharp/pull/209/commits/3b9751ffe837360356709749f90f25a116898726)

EasyPost.Net/EasyPost.Net.csproj
```diff
@@ -0,0 +1,54 @@
...
+    <WarningLevel>4</WarningLevel>
+    <DocumentationFile>..\lib\net\EasyPost.Net.XML</DocumentationFile>
+    <NoWarn>1701;1702;1591;1570</NoWarn>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Test' ">
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
```
Justintime50, line 37: This is dead code now.

Justintime50 (dismissed):

EasyPost.Net/EasyPost.Net.csproj
nwithan8: Resolved in [50391d3](https://github.com/EasyPost/easypost-csharp/pull/209/commits/50391d3b4a6bbb7f699b8326350bc8be5afd02e9)

Justintime50 (approved):

EasyPost.NetFramework/EasyPost.NetFramework.csproj
```diff
@@ -26,45 +27,36 @@
...
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed">
+      <HintPath>..\packages\Newtonsoft.Json.11.0.2\lib\net35\Newtonsoft.Json.dll</HintPath>
+      <Private>True</Private>
     </Reference>
-    <Reference Include="RestSharp, Version=106.13.0.0, Culture=neutral, PublicKeyToken=598062e77f915f75, processorArchitecture=MSIL">
-      <HintPath>..\packages\RestSharp.106.13.0\lib\net452\RestSharp.dll</HintPath>
+    <Reference Include="RestSharp, Version=100.0.0.0, Culture=neutral, PublicKeyToken=598062e77f915f75, processorArchitecture=MSIL">
```
jchen293, line 57: Just curious why did you switch the version to 100.0.0.0?

jchen293, line 57: Just curious why did you switch the version to 100.0.0.0?

jchen293 (approved):

EasyPost.NetFramework/EasyPost.NetFramework.csproj
Justintime50: I too am curious on this... 

Also, don't we want to pin the versions of deps? I suppose that can be done separately.

Justintime50: Any particular reason why it's not more explicitly listed as what it really is (RestSharpSigned) and we have to alias it this way?

nwithan8: Technically, RestSharpSigned v105 comes with the RestSharp v100 DLL inside it, and that's the RestSharp that we're using for this project. So the way it is configured currently is necessary.
image

PR #219, opened 2022-02-28
Pull v2 Branch into Master

Justintime50:
We released a patch to the User-Agent header assembly info that resides on the v2 branch but not on master. This brings master up-to-date.


nwithan8:
This has already been patched in the `tests` branch that utilizes the new dual-parent file structure.

Justintime50:
@nwithan8 ok, sounds good, I'll close this then.

PR #222, opened 2022-03-03
Generate code coverage reports

nwithan8:
Now that we have unit tests built out, let's see how complete they are.

This PR includes:
- A bash script to generate test coverage reports for each library version
- Instructions in the README about how this works


jchen293 (approved):

generate_test_reports.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+tests=(NetCore31 Net50 Net60 NetFramework35)
+
+for test in "${tests[@]}"; do
+    test_folder="EasyPost.Tests.$test"
+    cd "$test_folder"
+    dotnet test --collect:"XPlat Code Coverage"
+    reportgenerator -reports:TestResults/*/coverage.cobertura.xml -targetdir:../coveragereport/$test -reporttypes:Html
```
Justintime50, line 9: This will fail if `dotnet` or `reportgenerator` are not installed. We should add a check prior to this running that checks if they are installed and error if not. We should also call it out in the new section you're adding to the README.

Please also run `shellcheck` on this file (you'll need to download it), there are various items that need correcting to be compliant with POSIX sh.

Justintime50 (changes requested):

generate_test_reports.sh
nwithan8: Resolved in [75bac9b](https://github.com/EasyPost/easypost-csharp/pull/222/commits/75bac9bec61a02b4011d5e788bbb8a51b84f059d)

Justintime50 (approved):
This is fine for now. I'd like to see a better cross-platform solution long-term like a Makefile maybe. Shell scripts don't work out of the box on Windows.

PR #225, opened 2022-03-04
WIP: Improve Error de/serialization

nwithan8:
The way we currently parse error data from JSON into a C# object is rather janky. It's also technically incomplete, as the "errors" key (the suggestions of what to fix) don't deserialize properly.

Utilizing the new JSON de/serialization utilities, let's improve this.

This PR:
- Clarifies that an `HttpException` is an `ApiException`, so the end user knows it's related to the EasyPost API rather than a more generic networking issue
- Redesigns the `Error` class, splitting up suggestions into a separate, more explicit subclass, rather than reusing the `Error` class for two different elements of an error.
- Fix suggestion parsing, so users can actually see the suggestions returned by the EasyPost API

This PR also switches VCR to `Replay` for Refund tests; this was missed as part of the PR that added the Refund class.

**This is a WIP, and only currently implemented for the .NET version of the library.** Following feedback, this will be added to the .NET Framework version as well.


nwithan8:
Closing for now, will address API errors holistically in a later sprint.

Issue #98, opened 2017-08-01
does not work in mult-threaded, multi-apiKey environments


technology-shipcaddie-com:
The current implementation of easypost-csharp uses a ClientManager that is a static class, and thus is a Singleton and not thread safe because it stores the apiKey.
Currently, the apiKey gets stored in the ClientManager static class and is then pulled whenever an execute happens.  This works great in a single apiKey environment for both single and multi threaded environments.

This methodology breaks down quickly when in a multi-apiKey environment that is multi-threaded.
The following steps show an example of the problem that happens on busy systems.  Both examples below were happening frequently every day on our system when under load:

**Example 1:**

Thread 1:

- Client 1 sets the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 1");
**NOTE: singleton client Manager is now set to "apiKey 1" for everyone**
- Client 1 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();
- Client 1 gets lowest rate --- var rate = shipment.LowestRate(carriers, services);

Thread 2:
- Client 2 set the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 2");
**NOTE: singleton client Manager is now set to "apiKey 2" for everyone**
- Client 2 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();

Thread 1:
- Client 1 Buys label --- shipment.Buy(rate);
NOTE: this function then throws an error with the following description. "Object reference not set to an instance of an object"  This is because Client 1 is trying to purchase their own shipment with the "apiKey 2" of Client 2 which is in the ClientManager singleton object that the Buy method will pull to make its call.

Thread 2:
- Client 2 goes on and is able to work as if nothing has happened because the apiKey is their own "apiKey 2" value.

**Example 2:**

Thread 1:

- Client 1 sets the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 1");
**NOTE: singleton client Manager is now set to "apiKey 1" for everyone**
- Client 1 creates shipment object and populates it (not shown) --- EasyPost.Shipment shipment = new EasyPost.Shipment();

Thread 2:
- Client 2 set the Client Manager api key --- EasyPost.ClientManager.SetCurrent("apiKey 2");
**NOTE: singleton client Manager is now set to "apiKey 2" for everyone**
[...]

PR #228, opened 2022-03-09
WIP: Thread Manager

nwithan8:
In response to #98 , one proposed solution is to track the different API keys for different threads in a multi-threaded environment.

This is a proof of concept, implementing a ThreadManager, which stores and tracks thread-API key pairs.

To use it, users would first enable multithreading support by `ClientManager.IsMultithreaded = true`, and then register a thread with `ThreadManager.RegisterThread(thread/threadId, apiKey)`. When loading a `Client` for a request, if multithreading is enabled, the `ClientManager` will poll the `ThreadManager` for the API key corresponding to the currently running thread.

### DIFF
--- a/EasyPost.Net/ClientManager.cs
+++ b/EasyPost.Net/ClientManager.cs
@@ -7,7 +7,22 @@ namespace EasyPost
     /// </summary>
     public static class ClientManager
     {
-        private static Func<Client> getCurrent;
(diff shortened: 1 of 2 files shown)


nwithan8:
No longer needed because we are going async in https://github.com/EasyPost/easypost-csharp/pull/235

Issue #151, opened 2020-09-02
VB.Net error Create is ambiguous


chargerback-ron:
I'm using VB.Net to develop an EasyPost solution using the EasyPost.Net35.dll. When compiling I get the following error:

> error BC31429: 'Create' is ambiguous because multiple kinds of members with this name exist in class 'EasyPost.Address'.

According to documentation on MSDN, this error means that the source for the DLL was written in C#, which is case sensitive, so create and Create are two different things, whereas VB is case insensitive, so create and Create are the same thing. I looked through the code in this repo and I don't see a case difference in Address for Create. However, this repo seems to be pretty old; it says last updated 2014?

Justintime50:
Hey @chargerback-ron, we have released newer versions since that are not reflected on this GitHub page (see: https://www.nuget.org/packages/EasyPost-Official/).

As for your question, I can try digging around. Maybe @roehnan can shed some light on this?

chargerback-ron:
Thank you @Justintime50 I have verified that I am in fact using the latest version of the DLL from NuGet.

chargerback-ron:
@Justintime50 any update on my issue? I'm stalled in my EasyPost implementation right now. Thank you!

Justintime50:
I'll be honest, C# isn't my forte; however, in my research, I found the following that may be useful. I'm not entirely sure if this is an issue with the C# client library or the way it's being implemented on your side. Can you share a code snippet or gist here that we can take a look at?

Useful links (assuming this is a problem with the implementation of this library and not the actual library itself):

- https://visualstudiomagazine.com/articles/2016/06/01/conflicts-in-c-and-visual-basic.aspx
-  https://stackoverflow.com/questions/3491801/vb-net-ambiguous-class-name-error-when-no-ambiguity-exists/3660954#3660954

Let me know if either of these help or you are able to share a bit more context that we can look at. Thanks!

chargerback-ron:
Be happy to! It's quite easy to reproduce the error. Just try to compile this and you will get the error on the line FromAddress.Create(). It's definitely an error with the client library. The DLL is marked as "Language Neutral" but it is not if this simple code cannot compile. The Reflection link you sent looks promising; I'll let you know if it works.

        Dim FromAddress As New Address

        FromAddress.name = ShipperContact
        FromAddress.company = ShipperName
        FromAddress.street1 = ShipperAddress
        FromAddress.street2 = ShipperAddress2
        FromAddress.city = ShipperCity
[...]

PR #230, opened 2022-03-10
WIP: Removing non-static `Create` methods

nwithan8:
Per #151 , Visual Basic cannot tell the difference between a static and non-static version of the same method signature. This is an issue because we have both static and non-static `Create` methods in a handful of our classes.

The non-static methods, however, are for creating/recreating an object in case its `id` is blank, which logically does not make much sense. Why are we checking, i.e. when calling `myOrder.Buy()` whether `myOrder` exists? Of course it does.

The methods were likely implemented for a single-altered structure, accounting for if the `id` was somehow blank/not returned by the API. There are better ways to handle this. For now, I have removed the `id` check, but if it is truly necessary, we should have an exception thrown, rather have the library attempt to regenerate the object. In that case, we can rollback part of these changes and add a new `ResourceNotCreated` expection.

By removing these non-static `Create` methods, there should be no more ambiguity for Visual Basic.

### DIFF
--- a/EasyPost.Net/Address.cs
+++ b/EasyPost.Net/Address.cs
@@ -55,45 +55,11 @@ namespace EasyPost
         public string zip { get; set; }
 
         /// <summary>
-        ///     Create this Address.
-        /// </summary>
-        /// <exception cref="ResourceAlreadyCreated">Address already has an id.</exception>
-        public void Create() => Create(verify, verify_strict);
-
-        /// <summary>
-        ///     Create this Address.
-        /// </summary>
-        /// <param name="verifications">
-        ///     A list of verifications to perform on the address.
-        ///     Possible items are "delivery" and "zip4".
-        /// </param>
-        /// <param name="strictVerifications">
-        ///     A list of verifications to perform on the address.
-        ///     Will cause an HttpException to be raised if unsuccessful.
-        ///     Possible items are "delivery" and "zip4".
-        /// </param>
-        /// <exception cref="ResourceAlreadyCreated">Address already has an id.</exception>
-        public void Create(List<string> verifications = null, List<string> strictVerifications = null)
-        {
-            if (id != null)
-            {
-                throw new ResourceAlreadyCreated();
-            }
-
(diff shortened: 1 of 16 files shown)


EasyPost.Net/Address.cs
```diff
@@ -54,46 +54,12 @@ public class Address : Resource
...
-            Merge(SendCreate(AsDictionary(), verifications, strictVerifications));
-        }
-
         /// <summary>
         ///     Verify an address.
         /// </summary>
         /// <returns>EasyPost.Address instance. Check message for verification failures.</returns>
         public void Verify(string carrier = null)
         {
-            if (id == null)
```
Justintime50, line 92: We should definitely throw an error if no ID is available. You can check the verbiage from other libs. We need the ID to properly build the URL

Justintime50 (changes requested):

EasyPost.Net/Address.cs
nwithan8: Resolved in [4ca0652](https://github.com/EasyPost/easypost-csharp/pull/230/commits/4ca0652c98489543c9e6be57360d4bd3b59d7567)

nwithan8:
@Justintime50 per our discussions, I have added an extra test for the Visual Basic test suite. There's no need to recreate the entire test suite for the F# and Visual Basic variants.

There are two tests in the Visual Basic test suite:
- A basic test that is expected to throw a ClientNotConfigured error since there is no API key set. As long as it throws the expected error, and nothing else, we can assume Visual Basic is able to import our C# code properly.
- A basic test to create an Address. Again, this will produce the ClientNotConfigured error, but as long as no other errors are thrown, we can consider it a success. Address creation was one of a few spots where the competing named functions was causing compile errors. As long as this unit test pass, we can assume that issue has been resolved.

In fact, the test suite actually running is largely unnecessary. Just by the nature of the project (featuring formerly-problematic code) compiling, we can confirm that our C# code can be imported successfully into a Visual Basic or F# environment.

Justintime50 (approved):
Nice work!

jchen293 (approved):

PR #232, opened 2022-03-14
Fix issues with the port to v107. Need singleton RestClient and cannot use GetAwaiter().GetResult()!

kendallb:
Fix issues with the port to v107. As mentioned here:

https://restsharp.dev/v107/#restclient-lifecycle

The RestClient needs to be a singleton instance now, not an instance per request so we introduce a RestClient cache to cache them per API endpoint. 

Also since RestSharp v107 removed all the Sync functions from the API, it is incorrect to call an AsyncFunction and return the result using GetAwaiter().GetResult() as that WILL deadlock at some point. To do this correctly, you need to use a wrapper to run the functions Async and wait for the result. The wrapper in question is derived from the code in Rebus:

https://github.com/rebus-org/Rebus

### DIFF
--- a/EasyPost.Net/AsyncHelpers.cs
+++ b/EasyPost.Net/AsyncHelpers.cs
@@ -1,0 +1,136 @@
+﻿/*
+ * Taken from Rebus
+ */
+
+using System;
+using System.Collections.Concurrent;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace EasyPost
+{
+    public static class AsyncHelpers
+    {
+        /// <summary>
+        /// Executes a task synchronously on the calling thread by installing a temporary synchronization context that queues continuations
+        /// </summary>
+        /// <param name="task">Callback for asynchronous task to run</param>
+        public static void RunSync(
+            Func<Task> task)
+        {
+            var currentContext = SynchronizationContext.Current;
+            var customContext = new CustomSynchronizationContext(task);
+            try {
+                SynchronizationContext.SetSynchronizationContext(customContext);
+                customContext.Run();
+            } finally {
+                SynchronizationContext.SetSynchronizationContext(currentContext);
+            }
+        }
+
+        /// <summary>
+        /// Executes a task synchronously on the calling thread by installing a temporary synchronization context that queues continuations
+        /// </summary>
+        /// <param name="task">Callback for asynchronous task to run</param>
+        /// <typeparam name="T">Return type for the task</typeparam>
+        /// <returns>Return value from the task</returns>
+        public static T RunSync<T>(
+            Func<Task<T>> task)
+        {
+            T result = default;
+            RunSync(async () => {
+                result = await task();
+            });
+            return result;
+        }
+
+        /// <summary>
+        /// Synchronization context that can be "pumped" in order to have it execute continuations posted back to it
+        /// </summary>
+        private class CustomSynchronizationContext : SynchronizationContext
+        {
+            private readonly ConcurrentQueue<Tuple<SendOrPostCallback, object>> _items = new ConcurrentQueue<Tuple<SendOrPostCallback, object>>();
+            private readonly AutoResetEvent _workItemsWaiting = new AutoResetEvent(false);
+            private readonly Func<Task> _task;
+            private ExceptionDispatchInfo _caughtException;
+            private bool _done;
+
+            /// <summary>
+            /// Constructor for the custom context
+            /// </summary>
+            /// <param name="task"></param>
+            /// <exception cref="ArgumentNullException"></exception>
+            public CustomSynchronizationContext(
+                Func<Task> task)
+            {
+                _task = task ?? throw new ArgumentNullException(nameof(task), "Please remember to pass a Task to be executed");
+            }
+
+            /// <summary>
+            /// When overridden in a derived class, dispatches an asynchronous message to a synchronization context.
+            /// </summary>
+            /// <param name="function">Callback function</param>
+            /// <param name="state">Callback state</param>
+            public override void Post(
+                SendOrPostCallback function,
+                object state)
+            {
+                _items.Enqueue(Tuple.Create(function, state));
+                _workItemsWaiting.Set();
+            }
+
+            /// <summary>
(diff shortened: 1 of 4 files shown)


Justintime50:
Thanks for submitting this! Once tests/linting is corrected, we can take a closer look.

kendallb:
It's entirely not clear what is missing from the linting perspective. And I don't now why the NetFramework35 tests are failing as I did not change anything there.

Actually come to think of it I am pretty sure your NetFramework35 tests are busted across the board. Those tests are not using your VCR tool to play back the JSON, but making requests to a sandbox account and I think that account is now dead. I use the same keys for testing my Async version of the library, and all those tests are failing now also. So I think you guys need to fix that?

kendallb:
Actually all the NetFramework tests are failing due to SSL:

"The request was aborted: Could not create SSL/TLS secure channel."

kendallb:
Ok, I have fixed the SSL issue as we need to set this up in the TestStuite class as you require TLS 1.2 now for that website:

// Allow TLS 1.2 for the tests to work
System.Net.ServicePointManager.SecurityProtocol |= System.Net.SecurityProtocolType.Tls12;

However now of course I get the expected API key error:

"This api key is no longer active. Please use a different api key or reactivate this key."

So the API Key used for testing those frameworks is no longer active so you guys need to re-activate it again on your end for those tests to pass. I am not sure it's feasible to re-write those tests to use the VCR support (although I do wonder what the point of the VCR is - it lets your tests pass, but does nothing to make sure that it all still works to your real site).

kendallb:
Also Net50, Net60 and NetCore31 all pass for me (easy as they only use the VCR). Not sure why it reported them as failed before.

Justintime50:
The API keys we have are valid and are stored in GitHub's Secret Store, unfortunately though [these are not accessible by 3rd party contributors](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow). 

VCR allows us to not do exactly this ^ allowing us to store the requests/responses and replay them without the need for live API calls which gets difficult on a CI context and across various dev setups and contributors. Unfortunately, NetFramework3.5 (as written) doesn't allow us to take advantage of VCR which is why it's the only version without. As to the concern that it doesn't properly test anything - this is true to a point. Typical practice here would require the developer to remove the cassettes and re-record them to make a live API call.

We do indeed require TLS1.2, though I'm unaware of any changes we've made in regards to that in the last day. We have other PRs that have been in motion today and the last few days that didn't need this tweak, though thank you for adding it.

To move this forward, we'll review it as a team and then most likely recreate this PR (exactly as it is) from one of our accounts to ensure those API calls go through and the test suite passes, alternatively we may look into ways to circumvent this blocked test on CI. We'll keep you posted with what we determine the next step is!

kendallb:
Ok thanks. Yes in the old days the API keys were accessible and worked (so I was able to run my own tests in my own Async library against them). I cannot currently run the unit tests successfully on that library either, due to the API keys no longer working. Is there a document that describes how I could set up those tests to run against my own sandbox account?

And yes, I would recommend you block the NetFramework35 tests from blocking a merge due to the API key issues. Or have someone port the VCR to it :)

Justintime50:
I think that makes a fantastic topic for a blog post we may publish soon! As to having a document for how to test, there isn't really one at the moment. You can use your own API keys to test. The test suite expects env vars as `EASYPOST_TEST_API_KEY` and `EASYPOST_PROD_API_KEY`. Once set, most of those tests should work for you with a few exceptions (altering a couple of the fixture IDs found in the `Fixture.cs` file - 1) `UspsCarrierAccountId` and 2) `ChildUserId` to be the IDs of your own objects associated with your API key.)

Hopefully that helps get you started. We are constantly looking to improve our process, documentation, and libraries so any feedback is greatly appreciated.

kendallb:
Once thing that would be really nice would be to get official support for my Async version of the library. I did two things with that library when I forked it all those years ago:

1. Add support for Task Based Async (easy enough to add to your existing library without any breaking changes)
2. Got rid of the nasty dictionary functionality so the user can deal with native strongly typed POCO classes and not need to both with stuffing values into dictionaries.

The caveat is that I am not sure how to blend in support for the non-dictionary API. Perhaps it could be an entirely new client in the same library? But it is easy nicer to build code to my library than yours :) Similar to the library I wrote for ShipEngine as well which is based on their Swagger spec.

nwithan8:
> Once thing that would be really nice would be to get official support for my Async version of the library. I did two things with that library when I forked it all those years ago:
> 
> 1. Add support for Task Based Async (easy enough to add to your existing library without any breaking changes)
> 2. Got rid of the nasty dictionary functionality so the user can deal with native strongly typed POCO classes and not need to both with stuffing values into dictionaries.
> 
> The caveat is that I am not sure how to blend in support for the non-dictionary API. Perhaps it could be an entirely new client in the same library? But it is easy nicer to build code to my library than yours :) Similar to the library I wrote for ShipEngine as well which is based on their Swagger spec.

We've actually been having internal conversations in the past few days about exactly this. Overhauling our library to be async by default is on our roadmap, as is reworking it to use an instance of a client, rather than singletons, as it's been creating some issues for our users (both things I noticed you changed in your fork 😄 ). As for the POCO classes, that's also on my personal shortlist, but isn't specifically something we've looked at. Perhaps we can do a hybrid approach, offering both POCO classes and the existing dictionary methods.

Thanks for your contributions!

nwithan8:
Thank you for your contribution.

We have just released v3.0.0 of our library, which officially introduces, among several other things, asynchronicity into our library. All functions are now asynchronous to prevent deadlocks, stalled threads and lengthy, unneccessary wait times for API calls.

Slowly but surely, we are merging in features from your fork into our official library. Next on our roadmap is exploring replacing our singletons with instances.

For the time being, I am going to close this PR, as we have diverged from the PR quite a bit.

Thank you again.

kendallb:
Sounds good. But if you are still using a RestClient per request you will want to use my changes to use a single instance or your users can run out of resources.

kendallb:
Yeah the code still allocates a client per request right here which is the other thing I fixed:

https://github.com/EasyPost/easypost-csharp/blob/eef15b6ab7e7f2ae363b1a0725ede04ad2224cae/EasyPost/ClientManager.cs#L16

PR #233, opened 2022-03-14
fix: Improve VCR for unit tests

nwithan8:
- Replace built-in Scotch with submodule
- VCR Replay switches to Record automatically if cassette not found by default
- Bump Newtonsoft JSON dependency to latest


EasyPost.Net/EasyPost.Net.csproj
```diff
@@ -49,7 +49,7 @@
     </None>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="RestSharp" Version="107.3.0" />
```
Justintime50, line 53: These should probably be locked to their respective major versions instead of simply declared (will avoid people from getting into another restsharp not compatible issue.

Justintime50 (dismissed):
Code looks great, can we just pin the versions of these deps?

EasyPost.Net/EasyPost.Net.csproj
nwithan8: Resolved in [8a00135](https://github.com/EasyPost/easypost-csharp/pull/233/commits/8a0013587d461400f1f4f6db8f1691751fd30328)

Justintime50 (approved):

jchen293 (approved):

PR #234, opened 2022-03-16
[feat] Support for `columns` and `additional_columns` on Report creation

nwithan8:
Our API now supports specifying `columns` and `additional_columns` when [creating reports](https://www.easypost.com/docs/api#create-a-report).

This PR:
- Adds support for specifying `columns` and `additional_columns` (both lists of strings) in the `parameters` dictionary passed into `Report.Create(parameters)`
- Adds unit tests for this feature

Unfortunately, since the API response does not indicate the columns used for a report, the only way to ensure the report actually includes the columns requested would be to make a live API call, download the generated CSV report and parse its column headers. That is a complex unit test with many points of failure and prevents us from using VCR.

Instead, we simply assume that if no error is thrown during the API call to generate and retrieve the report, that the report contains the correct columns. If it doesn't, that constitutes a server-side error that is out of our control.

This has been end-to-end tested locally. The generated report included the requested columns as expected.

### DIFF
--- a/EasyPost.Tests/ReportTest.cs
+++ b/EasyPost.Tests/ReportTest.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
@@ -27,10 +26,16 @@ namespace EasyPost.Tests.Net
             });
         }
 
-        private static async Task TestCreateReport(string cassetteName, string reportType, string idPrefix)
+        private static async Task<Report> CreateAdvancedReport(string reportType, Dictionary<string, object> parameters)
+        {
+            parameters["start_date"] = Fixture.ReportStartDate;
+            parameters["end_date"] = Fixture.ReportEndDate;
+            return await Report.Create(reportType, parameters);
+        }
+
+        private static async Task TestCreateBasicReport(string cassetteName, string reportType, string idPrefix)
         {
             VCR.Replay(cassetteName);
-
             Report report = await CreateBasicReport(reportType);
 
             Assert.IsInstanceOfType(report, typeof(Report));
@@ -41,12 +46,11 @@ namespace EasyPost.Tests.Net
         {
             VCR.Replay(cassetteName);
 
-
             Report report = await CreateBasicReport(reportType);
 
             Report retrievedReport = await Report.Retrieve(report.id);
 
-            Assert.IsInstanceOfType(report, typeof(Report));
+            Assert.IsInstanceOfType(retrievedReport, typeof(Report));
             Assert.AreEqual(report.start_date, retrievedReport.start_date);
             Assert.AreEqual(report.end_date, retrievedReport.end_date);
         }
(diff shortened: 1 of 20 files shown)


jchen293:
Code looks good, can you add another unit test that only has input param `columns`? This was brought up by @Justintime50  from this comment: https://github.com/EasyPost/easypost-java/pull/108#discussion_r831376161

jchen293 (dismissed):

EasyPost.Net/Report.cs
```diff
@@ -39,16 +39,45 @@ public class Report : Resource
...
+
+                // removing from params since it's already been added in querystring
+                parameters.Remove("additional_columns");
+            }
+
             request.AddQueryString(parameters ?? new Dictionary<string, object>());
 
-            return request.Execute<Report>();
+            Report report = request.Execute<Report>();
+            return report;
```
jchen293, line 80:
```suggestion
            return request.Execute<Report>();
```

EasyPost.NetFramework/Report.cs
```diff
@@ -39,16 +39,42 @@ public class Report : Resource
...
+                    request.AddParameter("additional_columns[]", column, ParameterType.QueryString);
+                }
+
+                parameters.Remove("additional_columns");
+            }
             request.AddQueryString(parameters ?? new Dictionary<string, object>());
 
-            return request.Execute<Report>();
+            Report report = request.Execute<Report>();
+            return report;
```
jchen293, line 77:
```suggestion
            return request.Execute<Report>();
```

EasyPost.Tests.Net/ReportTest.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Net;
+using System.Threading;
```
jchen293, line 4: Nit: this can be removed since its not being used in anywhere.

jchen293 (dismissed):
lgtm, just some suggestions.

EasyPost.Tests.Net/ReportTest.cs
nwithan8: It will be when this gets merged with the async stuff, so I'll keep it for now.

nwithan8:
NOTE: I'd like to get the `async` branch merged first, since it has more changes and would therefore be harder it on this branch rather than the other way around.

nwithan8:
Putting in draft mode until `async` is merged for rebase

jchen293 (approved):

EasyPost.Tests/ReportTest.cs
```diff
@@ -27,11 +26,20 @@ private static async Task<Report> CreateBasicReport(string reportType)
...
+        private static async Task<Report> CreateAdvancedReport(string reportType, Dictionary<string, object> parameters)
         {
-            VCR.Replay(cassetteName);
+            parameters["start_date"] = Fixture.ReportStartDate;
+            parameters["end_date"] = Fixture.ReportEndDate;
+            return await Report.Create(reportType, parameters);
+        }
 
-            Report report = await CreateBasicReport(reportType);
+#nullable enable
```
Justintime50, line 36: This indentation doesn't look right?

EasyPost.Tests/ReportTest.cs
```diff
@@ -27,11 +26,20 @@ private static async Task<Report> CreateBasicReport(string reportType)
...
+        private static async Task TestCreateReport(string cassetteName, string reportType, string idPrefix, Dictionary<string, object>? parameters = null)
+        {
+            VCR.Replay(cassetteName);
+            Report report = parameters == null
+                ? await CreateBasicReport(reportType)
+                : await CreateAdvancedReport(reportType, parameters);
```
Justintime50, line 42: This test will only ever test a basic report creation (no columns) so why have the conditional advanced params here?

EasyPost.Tests/ReportTest.cs
```diff
@@ -41,12 +49,13 @@ private static async Task TestRetrieveReport(string cassetteName, string reportT
...
 
-
             Report report = await CreateBasicReport(reportType);
 
             Report retrievedReport = await Report.Retrieve(report.id);
 
-            Assert.IsInstanceOfType(report, typeof(Report));
+            Assert.IsInstanceOfType(retrievedReport, typeof(Report));
+            // Some reports take a long time to generate, so the URL won't be available immediately
+            // Assert.IsNotNull(retrievedReport.url);
```
Justintime50, line 58: We don't test against the URL, remove it completely and have this match the other libs.

EasyPost.Tests/cassettes/report/create_payment_log_report.json
```diff
@@ -1,50 +1,51 @@
...
+        "Content-Type": "application/json; charset=utf-8",
+        "Content-Length": "51"
+      },
       "Method": "POST",
-      "URI": "https://api.easypost.com/v2/reports/payment_log?start_date=2022-02-01&end_date=2022-02-03",
       "RequestHeaders": {
         "Accept": "application/json",
         "Authorization": "********",
         "content_type": "application/json",
         "User-Agent": "EasyPost/v2,CSharpClient/2.8.1,.NET/5.0.13"
```
Justintime50, line 15: Shouldn't this user agent say `2.8.2`? Why is it not using the most recent version? Leads me to believe that the header bug fix never made it to the master branch (this is why we must keep all of these things in separate PRs, pulling them into an already existing one leads to this confusion and if done wrong could lead to certain changes getting missed).

Justintime50 (changes requested):

EasyPost.NetFramework/Report.cs
```diff
@@ -39,16 +39,42 @@ public class Report : Resource
...
+            parameters = parameters ?? new Dictionary<string, object>();
+
+            if (parameters.ContainsKey("columns"))
+            {
+                foreach (string column in (List<string>)parameters["columns"])
+                {
+                    request.AddParameter("columns[]", column, ParameterType.QueryString);
+                }
+
+                parameters.Remove("columns");
```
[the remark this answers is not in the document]
nwithan8: This file no longer exists after rebase.

EasyPost.NetFramework/Report.cs
```diff
@@ -39,16 +39,42 @@ public class Report : Resource
...
+            }
+
+            if (parameters.ContainsKey("additional_columns"))
+            {
+                foreach (string column in (List<string>)parameters["additional_columns"])
+                {
+                    request.AddParameter("additional_columns[]", column, ParameterType.QueryString);
+                }
+
+                parameters.Remove("additional_columns");
```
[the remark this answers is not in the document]
nwithan8: This file no longer exists after rebase.

EasyPost.Tests/ReportTest.cs
nwithan8: Because :)

EasyPost.Tests/cassettes/report/create_payment_log_report.json
nwithan8: I'll merge that in first

Justintime50 (approved):

jchen293 (approved):

Issue #66, opened 2017-01-02
No Async support?


kendallb:
This library currently does not have any support for Async network calls? I might modify the code for the library to make it async myself, but I am curious why async support was not added from the start?

att14:
See #42

kendallb:
My new Async library is now complete and I have it as a fork on Github here:

https://github.com/kendallb/easypost-async-csharp

Note that I renamed the project since the changes are quite vast. I re-implemented all the API functions to hang off a single EasyPostClient with a corresponding IEasyPostClient interface so everything is instance based only. This allows me to properly mock all the EasyPost functions in our code and works nicely in a dependency injection environment like ours. 

Since this library is targeting only .Net 4.5 and later I made the entire library Async and did not implement any non-async versions. Any code can use the async functions in a non-async manner easily enough (Unit Tests are written that way). I also took the liberty of adding documentation to everything inline to help me when using Intellisense and consuming the library. 

I also did not like how some the API code was using dictionaries for parameter passing because it makes it much harder to refactor, so all the code was re-implemented to use instance only API's and the dictionaries are kept 100% internal to the API. With that change I also decided to change the naming convention for all the classes to match normal C# style CamelCase rather than the json_case style that is passed over the write. The code to convert a resource object to a dictionary automatically maps all the property names from CamelCase to camel_case, and the RestSharp JSON deserializer takes care of mapping it the other way with the incoming results. 

Finally due to the way exceptions work with async functions I removed the exception support from the code and the functions won't throw exceptions anymore. Rather for anything that is a first class EastPost object that is returned over the wire, if there was an error we deserialize the error object and stuff it into the response returned to the caller. So rather than catching exceptions the caller simply knows they will always get a valid result and can just check for the presence of result.RequestError and if present all the error message details are included in that object. 

I think the API is way cleaner now  and it will suit our purposes nicely. I hope you find it useful and perhaps would consider making it an official version of the API for others looking for C# Async support and/or the ability to use EasyPost in a DI environment?

PR #235, opened 2022-03-17
Migrate to async for .NET versions

nwithan8:
This has been a long time coming, but we're finally making our library asynchronous.

It's best practice in C# (and most programming languages, in fact) to make I/O bound tasks, such as network requests, [execute asynchronously](https://docs.microsoft.com/en-us/dotnet/csharp/async) so as to not lock up the main thread of an application. Our library previously executed HTTP requests synchronously, which needless blocked the application for several seconds, even the API call had already completed.

For example, in our own testing, we found that the `Shipment.Create()` function was taking almost 8 seconds to execute because the runtime was effectively padding the execution to give the HTTP call plenty of time to finish. When we migrated this function to asynchronous, the function returned the response from the HTTP call as soon as it was available, which was approximately 350 milliseconds. Another test, `Address.Create()` dropped from 2 seconds to ~400 milliseconds.

This PR:
- Migrates all functions in the .NET versions (Core 3.1, 5.0 and 6.0 currently) to asynchronous functions.
- <del>Our .NET Framework 3.5 version unfortunately cannot be migrated to asynchronous, as asynchronous functionality was not introduced until [.NET Framework 4.5](https://devblogs.microsoft.com/dotnet/async-in-4-5-worth-the-await/)</del>
- New .NET Standard 2.0 version is also asynchronous.
- Migrates all unit tests to asynchronous functions
- Fixed/enforce nullability of parameters in .NET

There's a joke that "async spreads like a virus", because when you make a function async/await, you have to make the function that called that function async/await, and the function that called that function... until all your functions are async or you reach a function that doesn't return anything.

Those implementing our now-asynchronous library may have varying levels of asynchronousity already in their applications, so this may be easier for some to implement than others. I recommend, for those unfamiliar, reading up on how asynchronous code works in C#, and how to [implement it into your final solution](https://exceptionnotfound.net/async-await-in-asp-net-csharp-ultimate-guide/).

If, for whatever reason, you must continue to run our API client library synchronously, you can use the built-in `.GetAwaiter().GetResult()` function to have any asynchronous function run synchronously. **This is [unadvised](https://www.nikouusitalo.com/blog/why-is-getawaiter-getresult-bad-in-c/), and will result in severe performance degradation.**


Justintime50:
Will review this once we adjust the build matrix for Standard.

EasyPost/Client.cs
```diff
@@ -51,7 +52,7 @@ public Client(ClientConfiguration clientConfiguration, HttpClient? customHttpCli
             {
                 Assembly assembly = typeof(Client).Assembly;
                 FileVersionInfo info = FileVersionInfo.GetVersionInfo(assembly.Location);
-                _libraryVersion = info.FileVersion;
+                _libraryVersion = info.FileVersion ?? "Unknown";
```
Justintime50, line 55: How did this sneak in or was it added? Rebase gone wrong somewhere? I thought this was already in the repo.

EasyPost/IResource.cs
```diff
@@ -4,7 +4,7 @@ namespace EasyPost
 {
     public interface IResource
     {
-        Dictionary<string, object> AsDictionary();
+        Dictionary<string, object?> AsDictionary();
```
Justintime50, line 7: What's with the proliferation of all these question marks throughout this PR? I thought this was only to make things async?

EasyPost/Insurance.cs
```diff
@@ -98,33 +99,20 @@ public static Insurance Create(Dictionary<string, object> parameters)
...
-        public static Insurance Retrieve(string id)
+        public static async Task<Insurance> Retrieve(string id)
         {
             Request request = new Request("insurances/{id}");
             request.AddUrlSegment("id", id);
 
-            return request.Execute<Insurance>();
-        }
-
-        private static Insurance SendCreate(Dictionary<string, object> parameters)
```
Justintime50, line 117: Please call the removal of this out in the CHANGELOG.

EasyPost/Order.cs
```diff
@@ -127,25 +128,12 @@ public static Order Create(Dictionary<string, object> parameters)
...
-        public static Order Retrieve(string id)
+        public static async Task<Order> Retrieve(string id)
         {
             Request request = new Request("orders/{id}");
             request.AddUrlSegment("id", id);
 
-            return request.Execute<Order>();
-        }
-
-        private static Order SendCreate(Dictionary<string, object> parameters)
```
Justintime50, line 138: Please call the removal of this out in the CHANGELOG.

EasyPost/Tracker.cs
```diff
@@ -102,31 +103,30 @@ public static TrackerCollection All(Dictionary<string, object> parameters = null
...
             request.AddBody(new Dictionary<string, object>
             {
                 {
                     "trackers", parameters
                 }
             });
-            request.Execute<Tracker>();
+            return await request.Execute();
             // This endpoint does not return a response so we return true here
-            return true;
```
Justintime50, line 129: Why are we switching this?

EasyPost/Insurance.cs
nwithan8: It was internal, so no one ever saw it. Dead code.

EasyPost/Order.cs
nwithan8: It was internal, so no one ever saw it. Dead code.

EasyPost/Tracker.cs
nwithan8: Instead of always getting `true` back from the function, we now actually get a boolean based on whether the API call was successful or not.

EasyPost/IResource.cs
nwithan8: The `?`  in C# means nullable. The IDE is constantly complaining that we are setting `null` as a default value to non-nullable values. So now, the values are nullable. More explicit to the end user, less likely to throw a runtime error.

EasyPost/Client.cs
nwithan8: I believe I added that in earlier, but it's just a fallback in case the .FileVersion function fails (returns null, but `_libraryVersion` can't have a null value assigned to it)

jmalatia:
Being that you guys are finally giving this C# library some much needed attention, you really should address:

- [Follow C# Naming styles](https://github.com/EasyPost/easypost-csharp/issues/66#issuecomment-270287354)... use PascalCase for naming classes and public properties
- Get rid of dictionaries on the public side of the API
- DI friendly

Many of the same things echoed for years... see this post https://github.com/EasyPost/easypost-csharp/issues/66#issuecomment-270287354 from 2017

see this library: https://github.com/kendallb/easypost-async-csharp

EasyPost/Tracker.cs
```diff
@@ -102,31 +103,30 @@ public static TrackerCollection All(Dictionary<string, object> parameters = null
...
-            Request request = new Request("trackers/create_list", RestSharp.Method.Post);
+            Request request = new Request("trackers/create_list", Method.Post);
             request.AddBody(new Dictionary<string, object>
             {
                 {
                     "trackers", parameters
                 }
             });
-            request.Execute<Tracker>();
+            return await request.Execute();
```
jchen293, line 128: Quick question, should this be like `return await request.Execute<Tracker>();`?

EasyPost.Tests/ShipmentTest.cs
```diff
@@ -14,11 +15,11 @@ public void Initialize()
         }
 
         [TestMethod]
-        public void TestCreate()
+        public async Task TestCreate()
         {
-            VCR.Replay("create");
+            VCR.Record("create");
```
jchen293, line 20: Why was this changed from `Replay` to `Record. I don't have much knowledge on scotch VCR yet, but this change seems inconsistent with other unit tests.

jchen293 (commented):
Code looks good, have you re-record cassettes locally to see if all unit tests pass? @nwithan8

EasyPost/Tracker.cs
nwithan8: No, this API endpoint doesn't return any data. `request.Execute()` returns true/false based on whether the API returned an error status code, while `request.Execute<T>()` returns a T-type object.

EasyPost.Tests/ShipmentTest.cs
nwithan8: Resolved in [13dec76](https://github.com/EasyPost/easypost-csharp/pull/235/commits/13dec7625f11be9e9d01d625178bc7e0fd723410)

nwithan8:
> Code looks good, have you re-record cassettes locally to see if all unit tests pass? @nwithan8

I just deleted and re-recorded all cassettes locally and "everything" works (two batch tests failed because of the timing issue, as expected. Since those cassettes couldn't be finished, I didn't push new cassettes to the repo)

jchen293 (approved):
nice work

PR #236, opened 2022-03-17
Drop .NET Framework v3.5, add .NET Standard 2.0

nwithan8:
Turns out our .NET Framework v3.5 support is really dragging us down
- We have to use a different version of RestSharp, which, because of syntactical changes, means we have to maintain duplicate files with only small differences between them.
- Because we can't use the newer RestSharp, we can't use the Scotch VCR, so all our unit tests have to make live API calls.
- .NET Framework is a Windows-only framework, so we can't run unit tests locally on a Mac; we have to send it up to GitHub Actions to see if tests pass.
- .NET Framework v3.5 doesn't have asynchronous support, so all API calls have to be done synchronously, which leads to long response times and locked threads.

All in all, it's a huge headache.

Thankfully, there is a solution (sort of): .NET Standard 2.0

Microsoft, in their ever-consistent .NET development, put out [.NET Standard](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-0/) a few years ago, which ultimately became the bridge between the old, Windows-only .NET Framework and the new, cross-platform .NET/.NET Core frameworks.

Long story short, we can ensure that our library will work on .NET Framework v4.6.1+ systems if we make our library target .NET Standard 2.0.

Yes, we would be dropping support for .NET Framework v3.5 (which is in long-term support until 2029) and every .NET Framework versions before v4.6.1 (which have [all been deprecated](https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework) at this point, so you shouldn't be using them if you still are).

On the positive side, we can once again have one set of parent files rather than two, with only slight modifications needed to account for .NET Standard 2.0 compatibility (mostly in regards to dynamic objects via JSON de/serialization; this has been fixed as part of this PR). We can use RestSharp v107 across the board. We can use VCR for all our tests. We can make the entire library asynchronous.

This PR:
- Removes the .NET Framework v3.5 files and test suite
- Moves `EasyPost.Net` to `EasyPost`, adds .NET Standard 2.0 to target frameworks
- Adds a .NET Standard 2.0 test suite (runs tests in a [.NET Framework v4.6.1 environment](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-0/))
- Makes minor changes to custom JSON serialization and unit tests to account for .NET Standard 2.0 compatibility

GitHub CI for this PR may fail initially, as Scotch currently does not support .NET Standard 2.0. Once Scotch is updated, these tests should be able to run.


Justintime50:
While I think on this, can we get Scotch ready by adding support for Net Standard (without yet removing Net Framework)? Then pull that into this PR so we have passing builds prior to review?

nwithan8:
> While I think on this, can we get Scotch ready by adding support for Net Standard (without yet removing Net Framework)? Then pull that into this PR so we have passing builds prior to review?

https://github.com/EasyPost/scotch/pull/3

EasyPost.sln
```diff
@@ -69,6 +51,14 @@ Global
 		{B05C7F02-568D-4587-91FF-D27B4DF5ED16}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{B05C7F02-568D-4587-91FF-D27B4DF5ED16}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{B05C7F02-568D-4587-91FF-D27B4DF5ED16}.Release|Any CPU.Build.0 = Release|Any CPU
+		{1B44D741-4EC2-4085-BF6E-8A81D330D77E}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
Justintime50, line 54: What are all these and why do they seem to change with every PR?

EasyPost/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("EasyPost")]
+[assembly: AssemblyDescription("EasyPost .NET/.NET Core Shipping API Client Library for .NET https://easypost.com/docs")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("EasyPost")]
+[assembly: AssemblyProduct("EasyPost")]
+[assembly: AssemblyCopyright("Copyright © 2022")]
```
Justintime50, line 13: This should match the LICENSE probably.

Justintime50 (commented):
I feel comfortable making this change. We'll maintain a `v2` branch and backport necessary bug fixes for those that don't want to make the switch to v3 with async/await and a new matrix of built binaries.

@nwithan8 please pull in the new Scotch version and fix the build and I'll take a look.

EasyPost.sln
nwithan8: It's the Debug and Release configurations for the new .NET Standard project

Justintime50: Ah, it's changed so much because we keep tweaking the platforms we'll use haha.

EasyPost/Properties/AssemblyInfo.cs
nwithan8: Resolved in [17ae1c2](https://github.com/EasyPost/easypost-csharp/pull/236/commits/17ae1c2e7f7dab18144c13c992d365649e6985d2)

EasyPost.sln
nwithan8: Not necessarily, it's the project name. But yes, it's harmless.

Justintime50 (approved):

jchen293 (approved):

PR #238, opened 2022-03-24
Fix API query vs body parameters

nwithan8:
As we recently learned when implementing the new columns features for report creation, all non-GET requests against our API should be sending parameters via the request body and not as query parameters.

A number of POST requests in the C# library are sending parameters in the query string rather than in the request body.

This PR makes the following modifications:
- Most non-GET requests use body parameters now rather than query parameters (`Address.Create` will be addressed in a later PR)
- Attempting to add query parameters to a non-GET request will throw an error
- Re-recorded all cassettes


Justintime50:
@nwithan8 is this ready for re-review then?

EasyPost/Request.cs
```diff
@@ -53,9 +53,31 @@ public void AddParameter(string name, string? value, ParameterType type)
         /// <param name="parameters">Dictionary of key-value pairs for creating URL query parameters.</param>
         public void AddQueryString(IDictionary<string, object> parameters)
         {
+            if (_restRequest.Method != Method.Get)
+            {
+                throw new ArgumentException("Query string parameters can only be added to GET requests.");
```
Justintime50, line 58: Please add a checkbox of all langs to our internal task and check off C# is done due to this so we don't forget.

Justintime50 (approved):

jchen293 (approved):

PR #243, opened 2022-03-29
Compare objects rather than IDs in unit tests

nwithan8:
For a number of our unit tests, we're simply comparing the IDs of the original and test object, rather than comparing the whole objects. This PR introduces that functionality.

- New overloaded function for evaluating equality of Resource objects
- New function to convert Resource objects to JSON representation
- Unit tests now assert whole-object equality rather than just IDs (sans special cases where the two objects have known differences; ID comparison only in those situations)


Justintime50 (approved):
Nice!

jchen293 (approved):

PR #244, opened 2022-04-04
docs: simplify and clean up README

Justintime50:
Simplifies and cleans up the README as part of our client library README improvement efforts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ namespace shipments
 {
     class createShipment
     {
-        static void Main()
+        static async void Main()
         {
             EasyPost.ClientManager.SetCurrent(Environment.GetEnvironmentVariable("EASYPOST_API_KEY"));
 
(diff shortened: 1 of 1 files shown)


jchen293 (approved):

README.md
```diff
@@ -1,33 +1,83 @@
...
+                { "weight", 10 },
+            }
+
+            Shipment shipment = Shipment.Create(new Dictionary<string, object>() {
+                { "from_address", fromAddress },
+                { "to_address", toAddress },
+                { "parcel", parcel },
+            });
+
+            shipment.Buy(shipment.LowestRate());
```
nwithan8, line 69: This needs to be async

nwithan8 (changes requested):

README.md
Justintime50: Perhaps I pull it out of a class then and only have it be a barebones example, or do you mean the calls to the functions need to be async? Either way, care to provide suggestions on the review of what's needed where?

Justintime50: I believe I've added async/await where necessary. Ready for a double check.

nwithan8 (approved):

PR #254, opened 2022-04-15
Combine .NET platform test projects into one project

nwithan8:
# Description

We have a separate project for each platform that we support (i.e. `EasyPost.Tests.Net50, EasyPost.Tests.NetStandard20`), but the source code in each of these projects is the exact same (they all simply link to the source code files in the parent test project `EasyPost.Tests`). This creates a lot of visual clutter in our repo for no real reason.

Instead, we can have one test project, `EasyPost.Tests` that targets all our supported framework versions. This will still produce the same end result (an individual DLL for each supported platform) with far less clutter.

NOTE: Visual Basic and F# tests are still in separate projects, since those are different languages.

- Combine .NET platform test projects into one project
- Update GitHub CI to work with new structure

# Testing

Local testing
- All expected DLLs are still created
- Running unit tests still executes all tests in all versions
- When running a specific test, the IDE may offer you the choice of which (supported) framework you'd like to use to run the test.

# Pull Request Type


nwithan8:
The names of each test have changed (lowercase rather than title case), so the required tests list on GitHub will need to be altered.

.github/workflows/ci.yml
```diff
@@ -22,15 +22,11 @@ jobs:
       EASYPOST_PROD_API_KEY: ${{ secrets.EASYPOST_PROD_API_KEY }}
     strategy:
       matrix:
-        platform: [ 'NetStandard20', 'NetCore31', 'Net50', 'Net60' ]
+        platform: [ 'netstandard2.0', 'netcoreapp3.1', 'net5.0', 'net6.0' ]
```
Justintime50, line 25: Do we need to change the casing here? Would prefer not to bug systems if we don't need to.

Justintime50 (commented):
So this change will still run all the test suites just without the need to maintain separate test projects for each?

.github/workflows/ci.yml
nwithan8: Right now, I'm not seeing a way. By default, `msbuild` builds a structure like this:
image

And I can't find a way to override the path. If I indicate a new folder path, it only makes a single DLL file in the indicated path. That won't work for us, because we explicitly want to test each platform version.

nwithan8: Alright, I found a workaround to keep us from having to change the names of the required tests

Justintime50 (approved):
Thanks for the helpful notes!

jchen293 (approved):

PR #257, opened 2022-04-20
Make building requests more abstract

nwithan8:
# Description

While trying to block developers from sending query parameters for non-GET/DELETE, we discovered that the process of building a request in C# is more manual that we'd like, and introduces more points of failure (e.g. the need for the query parameter check is because the developer is adding query parameters manually upstream).

This PR overhauls the Request functionality, abstracting away the building of the request, making it simpler to add parameters and less prone to user error.
- User calls `AddParameter(s)` on a `Request` object. Internally, parameters are added as query or body parameters depending on method.
- Explicitly indicate method for all requests

# Testing

Re-run cassettes successfully

# Pull Request Type


Justintime50 (approved):
This is MUCH cleaner, nicely done!

jchen293 (approved):

PR #263, opened 2022-05-06
feat: adds lowest_rate to more objects

nwithan8:
# Description

- Adds `LowestRate` function to `Order` and `Pickup`, improves function on `Shipment`
- Adds `myShipment.LowestSmartrate()` function
- Adds `SmartrateAccuracy` enums so users don't have to know the JSON key name
- Adds new `PickupRate` class (explicit Rate class for pickups) to match Java and Go

# Testing

New unit tests and new recorded cassettes

# Pull Request Type


EasyPost.Tests/PickupTest.cs
```diff
@@ -81,5 +81,34 @@ public async Task TestCancel()
...
+            Assert.AreEqual("USPS", lowestRate.carrier);
+
+            // test lowest rate with service filter (should error due to bad service)
+            List<string> services = new List<string>
+            {
+                "BAD_SERVICE"
+            };
+            Assert.ThrowsException<FilterFailure>(() => pickup.LowestRate(null, services, null, null));
+
+
```
Justintime50, line 105: Nit: No double line breaks

EasyPost.Tests/ShipmentTest.cs
```diff
@@ -73,7 +88,7 @@ public async Task TestBuy()
             _vcr.SetUpTest("buy");
```
Justintime50, line 90: We should remove all these double line breaks throughout

Justintime50, line 90: We should remove all these double line breaks throughout

EasyPost.Tests/ShipmentTest.cs
```diff
@@ -37,7 +52,7 @@ public async Task TestRetrieve()
             _vcr.SetUpTest("retrieve");
```
Justintime50, line 54: Nit: No double line breaks

Justintime50, line 54: Nit: No double line breaks

EasyPost.Tests/ShipmentTest.cs
```diff
@@ -230,5 +267,56 @@ public async Task TestCreateWithIds()
...
+            Smartrate lowestSmartrate = await shipment.LowestSmartrate(1, SmartrateAccuracy.Percentile90);
+            Assert.AreEqual("Priority", lowestSmartrate.service);
+            Assert.AreEqual(7.37, lowestSmartrate.rate);
+            Assert.AreEqual("USPS", lowestSmartrate.carrier);
+
+            // test lowest smartrate with invalid filters (should error due to strict delivery_days)
+            await Assert.ThrowsExceptionAsync<FilterFailure>(async () => await shipment.LowestSmartrate(0, SmartrateAccuracy.Percentile90));
+
+            // test lowest smartrate with invalid filters (should error due to bad delivery_accuracy)
+            // this test is not needed in the C# CL because it uses enums for the accuracy (can't pass in an incorrect value)
```
Justintime50, line 319: Right, we want to test that your enums work is the point. This will guard against enums getting removed/changed in the future by someone without context. Can we add a tests that tries to use an invalid enum?

Justintime50, line 319: Right, we want to test that your enums work is the point. This will guard against enums getting removed/changed in the future by someone without context.

EasyPost/Exception.cs
```diff
@@ -66,5 +66,16 @@ public override string Message
...
+        {
+        }
+    }
+
+    [Serializable]
+    public class FilterFailure : Exception
+    {
+        public FilterFailure(string message) : base(message)
+        {
+        }
```
Justintime50, line 79: What's all this?

Justintime50, line 79: What's all this?

EasyPost/ClientManager.cs
```diff
@@ -35,7 +35,7 @@ internal static Client Build()
         {
             if (getCurrent == null)
             {
-                throw new ClientNotConfigured();
+                throw new ClientNotConfigured("No function set to retrieve the current client.");
```
Justintime50, line 38: What's this?

Justintime50, line 38: What's this?

EasyPost/TimeInTransit.cs
```diff
@@ -18,5 +18,24 @@ public class TimeInTransit
         public int? percentile_97 { get; set; }
         [JsonProperty("percentile_99")]
         public int? percentile_99 { get; set; }
+
```
jchen293, line 21: Add comments on what this function does.

EasyPost/Shipment.cs
```diff
@@ -324,8 +309,6 @@ public static async Task<Shipment> Retrieve(string id)
             return await request.Execute<Shipment>();
         }
 
-        private static void FilterRates(ref List<Rate> rates, Func<Rate, bool> filter) => rates = rates.Where(filter).ToList();
```
Justintime50, line 327: Hmm... what was this here for?

Also why removal?

Justintime50, line 327: Hmm... what was this here for?

EasyPost/Utilities/Enumeration.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace EasyPost.Utilities
+{
+    public abstract class Enumeration : IComparable
```
Justintime50, line 8: We definitely need tests for this

Justintime50, line 8: We definitely need tests for this

EasyPost/Utilities/Rates.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <summary>
+        ///     Get the lowest rate from a list of rates
+        /// </summary>
+        /// <param name="rates">List of rates to parse.</param>
+        /// <param name="includeCarriers">Carriers to include in the filter.</param>
+        /// <param name="includeServices">Services to include in the filter.</param>
+        /// <param name="excludeCarriers">Carriers to exclude in the filter.</param>
+        /// <param name="excludeServices">Services to exclude in the filter.</param>
+        /// <returns>Lowest rate matching the filter.</returns>
+        public static Rate GetLowestObjectRate(List<Rate> rates, List<string>? includeCarriers = null, List<string>? includeServices = null, List<string>? excludeCarriers = null, List<string>? excludeServices = null)
```
Justintime50, line 18: This looks a bit more verbose than the original, what changed between the two? As this is a highly trafficked function, we'd like to keep the behavior the same as before but just make it into a utility.

Justintime50, line 18: This looks a bit more verbose than the original, what changed between the two? As this is a highly trafficked function, we'd like to keep the behavior the same as before but just make it into a utility.

Justintime50 (changes requested):

EasyPost/Utilities/Enumeration.cs
nwithan8: What would a test for an enum look like?

EasyPost.Tests/ShipmentTest.cs
nwithan8: There's no such thing as an invalid enum?

EasyPost/Exception.cs
nwithan8: FilterFailure is a custom exception thrown specifically in the filtering functions, easier to specifically target, more descriptive to the end user about what the exact problem was (rather than a generic Exception)

EasyPost/ClientManager.cs
nwithan8: Oops, that was just a cleanup thing that accidentally slipped in here. I guess ClientNotConfigured was never throwing an error message before, just the exception.

EasyPost/Shipment.cs
nwithan8: It was being used by the old `LowestRate` function when that function was inside/only used by the Shipment class.
Since that logic has been moved to a utility file, this is no longer needed.

EasyPost/Utilities/Rates.cs
nwithan8: Nothing really changed, the method signature (for Shipment) is the same as it was, so it won't require any rewriting on the end-users part: https://github.com/EasyPost/easypost-csharp/blob/c69f20c8e37e8647f7669a4f48feda9f6defbd61/EasyPost/Shipment.cs#L200

Logically, it's still doing the same stuff, in the same order.
1. Setting up local variables
1. Filtering carriers and services
1. Looping through the remaining rates, looking for the lowest one.
1. Returning the lowest rate, or throwing an exception if nothing matches the criteria.

nwithan8:
A recent commit for formatting touched a majority of the code files, my apologies. I will roll this back and re-introduce it after the PR has been approved.

EasyPost.Tests/ShipmentTest.cs
nwithan8: Looking at the link you provided (thank you, by the way), it seems like that exception is thrown because a value cannot be cast into the enum type. This error would be thrown in the same code block as the attempted cast, which, in our case, would be above our library, in the end-user's source code. So there's nothing we need to account for in our library, as the error would be thrown before the enum even entered our functions.

As a POC, I did some basic testing in our library, and I am getting compiler errors when attempting to abuse the enum (which, to be honest, are really just static, final variables rather than true enums).

image

I'll explain the failures, in order:
1. The integer `123` cannot be cast to a `SmartrateAccuracy` object, compilation will fail.
1. The string "string" cannot be cast to a `SmartrateAccuracy` object, compilation will fail.
1. User cannot declare a new `SmartrateAccuracy` object, since the constructor is marked private, and therefore inaccessible to end-users
1. Uses the `SmartrateAccuracy` "enum" correctly, no errors thrown.

Please let me know if there are any other edge-cases you'd like me to attempt.

Since it seems any abuse of the "enum" results in a compilation failure, there is no real way to write a unit test for this, since the unit test would fail to compile and run.

Justintime50: This is sufficient. Thank you for digging in. This is ultimately what I wanted to know. In Python, we wouldn't catch this till runtime where we'd want a test to ensure it worked whereas in C# it'll error at compile time so to your point - no test other than these end to end tests here are needed and we can move on.

EasyPost/Pickup.cs
```diff
@@ -33,14 +34,32 @@ public class Pickup : Resource
...
+        /// </summary>
+        /// <returns>List of Rate objects.</returns>
+        internal List<Rate> Rates
+        {
+            get
+            {
+                List<Rate> rates = new List<Rate>();
+                foreach (PickupRate pickupRate in pickup_rates)
+                {
+                    rates.Add(pickupRate);
```
jchen293, line 56: Just a question, where do you convert the `pickupRate` to `rate`?

jchen293 (approved):

EasyPost/Pickup.cs
nwithan8: `PickupRate` is a subclass of `Rate`, so I can put a `PickupRate` object into a list of `Rate` objects without needing to cast.

EasyPost.Tests/PickupTest.cs
```diff
@@ -15,71 +15,99 @@ public void Initialize()
             _vcr = new TestUtils.VCR("pickup");
         }
 
-        private static async Task<Pickup> CreateBasicPickup()
```
Justintime50, line 18: I'd like to keep the CRUD order of tests like we have in all other test files. Can we put the order back to:

1. create
2. retrieve
3. all
4. update
5. delete

EasyPost.Tests/ShipmentTest.cs
```diff
@@ -73,42 +44,145 @@ public async Task TestBuy()
...
-            Shipment shipment = await Shipment.Create(Fixture.FullShipment);
+            await shipment.GenerateLabel("ZPL");
 
-            await shipment.RegenerateRates();
+            Assert.IsNotNull(shipment.postage_label.label_zpl_url);
+        }
 
-            List<Rate> rates = shipment.rates;
+        [TestMethod]
+        public async Task TestCreate()
```
Justintime50, line 67: Can we keep the order here too? Moving functions around isn't ideal because it blows up the lines to review and harder to determine what actually changed vs what got moved.

Justintime50 (changes requested):
Code mostly looks fine; however, we shouldn't be moving functions around for reasons detailed in another comment here. Can we put the ordering of functions back so it's easier to review what changed here?

EasyPost.Tests/ShipmentTest.cs
nwithan8: I have reverted the rearrangement manually.

One of the pre-commit checks on my IDE is to "rearrange code", which helps keep code in line with our style guide, in terms of in what order they appear. This was recently enabled on my end so I didn't have to remember to do this manually. Clearly, already our code base has strayed from the style guide.

Currently, per our style guide, code is arranged by access level, by static/non-static, and alphabetically.

After this PR is merged, I'm going to run a system-wide lint of this code base and open a separate PR. There will be no code or functionality changes in it, just rearrangement to fall in line with our style guide. And with the pre-commit check, moving forward, things should never need to be re-ordered again.

EasyPost/ClientManager.cs
Justintime50: Did we still need this? Sounds like you were going to remove it?

Justintime50 (approved):

PR #264, opened 2022-05-06
[WIP] Instance-based rewrite

nwithan8:
# Description

This is a rewrite of the C# client library to be instance-based (user creates an API client with a specific API key, then uses that client to make API calls via microservices).

This is modeled after Stripe's .NET client library (microservices), and allows users to have multiple instances of an API client (each with different credentials and settings), rather than relying on a singleton Client. This should resolve the multi-threading issues that have plagued this library (#98). ~~This client structure is also extendable, allowing for different sub-client types with access to different services (i.e. a `V2Client` for normal operations, and a `BetaClient` for access to beta features)~~ This has been reworked to maintain one `Client`, and instead utilize attributes to determine API version compatibility for the microservices.

The rewrite also locks down the library as much as possible, making many classes `internal` or `protected` to avoid unwanted access by end-users, as well as to clear up confusion as to what functions and classes to call.

The rewrite also fundamentally re-orders the relationship between a `Resource`, a `Client` and a `Request`, simplifying the logic and making it as extendible as possible. Developers implementing a new function in any `Service` will often only need to write two or three lines of new code moving forward, as much of the process of building an HTTP request and processing a response has been abstracted away. Common HTTP calls, such as POST create new object, GET retrieve existing object by ID, DELETE remove existing object by ID, GET list existing objects and PATCH update an existing object by ID, have been built in as reusable methods into the `Service` interface.

NOTE: This is still a WIP, and I will do my best to maintain this branch in parallel with `master`, including adding all new features, until this can be considered for adoption.

Updated/ported features:
- Add VCR NuGet package (0.3.1)
- Use strict matching and record/replay logic in VCR
- Use censoring in cassettes
- Beta client added
- EndShipper added
- Lowest rate support added
- Billing added
- Replace PUT with PATCH where needed

# Testing

All cassettes have been re-recorded and re-run, confirmed working.

# Pull Request Type


nwithan8:
Switched from MSTest to XUnit for test class design (still using MSTest for assertions for now), as XUnit is the only testing package that allows for inherited classes (allowing to use the new UnitTest base class): https://www.lambdatest.com/blog/nunit-vs-xunit-vs-mstest/

nwithan8:
This has been discussed and agreed upon internally. Will be re-created piece-by-piece. Closing.

PR #267, opened 2022-05-17
add if statement to make sure an API key is required before making li…

jchen293:
# Description

Add if statement to check if there is an API key before making any live API calls.

# Testing

E2E test with no API key, an error is thrown.
Screen Shot 2022-05-17 at 5 16 18 PM

# Pull Request Type

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,8 @@ jobs:
     # derived from https://dev.to/felipetofoli/github-actions-for-net-full-framework-build-and-test-299h
     runs-on: windows-2022
     env:
-      EASYPOST_TEST_API_KEY: ${{ 123 }}
-      EASYPOST_PROD_API_KEY: ${{ 123 }}
+      EASYPOST_TEST_API_KEY: "123"
+      EASYPOST_PROD_API_KEY: "123"
     strategy:
       matrix:
         name: [ 'NetStandard20', 'NetCore31', 'Net50', 'Net60' ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,8 @@ jobs:
     # derived from https://dev.to/felipetofoli/github-actions-for-net-full-framework-build-and-test-299h
     runs-on: windows-2022
     env:
-      EASYPOST_TEST_API_KEY: ${{ secrets.EASYPOST_TEST_API_KEY }}
-      EASYPOST_PROD_API_KEY: ${{ secrets.EASYPOST_PROD_API_KEY }}
+      EASYPOST_TEST_API_KEY: ${{ 123 }}
+      EASYPOST_PROD_API_KEY: ${{ 123 }}
     strategy:
       matrix:
         name: [ 'NetStandard20', 'NetCore31', 'Net50', 'Net60' ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Next Release
+Requests will now fail fast with an error if an API key is not provided instead of making a live API call with no key.
+
 ## v3.0.0
 
 Upgrading major versions of this project? Refer to the [Upgrade Guide](UPGRADE_GUIDE.md).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 # Next Release
+
 Requests will now fail fast with an error if an API key is not provided instead of making a live API call with no key.
 
 ## v3.0.0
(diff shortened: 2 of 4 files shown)


Justintime50:
Code looks good. We able to add a simple unit test to make sure this works too? I meant to do this on the others and will add it in the future.

nwithan8 (changes requested):
Technically, since the ApiKey and ApiBase properties are public, people could edit them outside of the constructor. I was going to address this on my own, but since we're doing it here, can you replace `public` with `internal` on the `ApiKey` and `ApiBase` properties? That will block end-users from incorrectly accessing the properties, which still making them accessible to other aspects of the library where they're needed.

jchen293:
> Technically, since the ApiKey and ApiBase properties are public, people could edit them outside of the constructor. I was going to address this on my own, but since we're doing it here, can you replace `public` with `internal` on the `ApiKey` and `ApiBase` properties? That will block end-users from incorrectly accessing the properties, which still making them accessible to other aspects of the library where they're needed.

Good point, thanks for pointing this out!

.github/workflows/ci.yml
```diff
@@ -18,8 +18,8 @@ jobs:
     # derived from https://dev.to/felipetofoli/github-actions-for-net-full-framework-build-and-test-299h
     runs-on: windows-2022
     env:
-      EASYPOST_TEST_API_KEY: ${{ secrets.EASYPOST_TEST_API_KEY }}
-      EASYPOST_PROD_API_KEY: ${{ secrets.EASYPOST_PROD_API_KEY }}
+      EASYPOST_TEST_API_KEY: ${{ 123 }}
+      EASYPOST_PROD_API_KEY: ${{ 123 }}
```
Justintime50, line 22:
```suggestion
      EASYPOST_TEST_API_KEY: "123"
      EASYPOST_PROD_API_KEY: "123"
```

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Next Release
+Requests will now fail fast with an error if an API key is not provided instead of making a live API call with no key.
```
Justintime50, line 4: This needs a line break after the header
```suggestion
## Next Release

Requests will now fail fast with an error if an API key is not provided instead of making a live API call with no key.
```

EasyPost.Tests/ErrorTest.cs
```diff
@@ -31,5 +32,21 @@ public async Task TestError()
                 Assert.IsTrue(error.Errors.Count == 2);
             }
         }
+
+        [TestMethod]
+        public async Task TestNoApiKey()
+        {
+            _vcr = new TestUtils.VCR("no_api_key");
+            _vcr.SetUpTest("empty", "");
+
```
Justintime50, line 41: Hmmm... we shouldn't need to invoke VCR here at all. We should really be making a new client with an empty API key.

EasyPost/ClientConfiguration.cs
```diff
@@ -10,11 +12,11 @@ public class ClientConfiguration
         /// <summary>
         ///     The API base URI to use on a per-request basis.
         /// </summary>
-        public string ApiBase { get; set; }
+        internal string ApiBase { get; set; }
```
Justintime50, line 15: Why were these made internal, this is a breaking change

Justintime50 (changes requested):

EasyPost/ClientConfiguration.cs
jchen293: I don't think this is a breaking change, we are making it `internal` to avoid users modifying the value. The place that `ClientConfiguration()` is being called is in this line: https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/ClientManager.cs#L16

EasyPost.Tests/ErrorTest.cs
jchen293: addressed

EasyPost/ClientConfiguration.cs
nwithan8: Correct, this will not introduce any breaking changes.

Justintime50: Hmm, if it's public there's a possibility users could be accessing these no? And if so, making it internal would suddenly make them unaccessible

nwithan8: True, but the only `ClientConfiguration` object in the library is stored as a property under `Client`, where it's private and inaccessible. So, while the ClientConfiguration's ApiKey could be publicly accessed and modified after construction, a Client's ClientConfiguration (and by extension, it's ApiKey) could not be accessed or modified after construction.

This is eliminating a possible edge case that I'm fairly confident no end-user has exploited yet.

nwithan8: I also need to modify the accessibility levels of these various properties as a part of my implementation of the Beta client anyway, so this conversation is ultimately moot at this point.

Justintime50 (approved):

nwithan8 (approved):

PR #271, opened 2022-05-23
[WIP] Warn users about accessing internally-intended functions

nwithan8:
# Description

Since many of our methods are `public`, it seems some end-users are utilizing functions that were intended for internal-only use. This PR adds deprecation notices to those methods to warn end-users to not utilize these functions. After enough time has passed with the deprecation notices, many of these functions should be switched to `internal`, `protected` or `private` access levels to make them inaccessible to end-users.

- Delete unnecessary IResource interface
- Mark non end-user functions as obsolete, change access level in the future

# Testing

All unit tests pass

# Pull Request Type


EasyPost/Resource.cs
```diff
@@ -64,11 +67,11 @@ public void Merge(object source)
 
             switch (value)
             {
-                case IResource resource:
+                case Resource resource:
```
Justintime50, line 70: What's with the `IResource` to `Resource` change?

nwithan8: IResource was an interface, I'm assuming added for reusability way-back-when, but it's not being used anywhere except here, and since it's an interface, it's interfering with the ability to change the access limit on some functions.

TL;DR old unnecessary code that was removed.

Justintime50 (approved):

jchen293 (approved):

PR #275, opened 2022-06-03
Add billing functions

nwithan8:
# Description

- Add billing functions, unit tests and cassettes
- Fix consistency of adding parameters to requests

# Testing

New cassettes added, all tests pass

# Pull Request Type


jchen293 (changes requested):

EasyPost/PaymentMethod.cs
```diff
@@ -0,0 +1,135 @@
...
+            return await Delete(paymentMethod.id);
+        }
+
+        /// <summary>
+        ///     Fund a credit card.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="paymentMethodId">ID of payment method to fund.</param>
+        /// <returns>An EasyPost.CreditCardFund instance.</returns>
+        public static async Task<CreditCardFund> Fund(string amount, string paymentMethodId)
```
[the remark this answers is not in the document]
nwithan8: Yeah, I'll admit this was somewhat purposeful since, now that I can see how this was designed, I'm not a huge fan of how this was structured.

But I'll concede.

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,124 @@
...
+            return paymentMethod;
+        }
+
+        /// <summary>
+        ///     Fund a credit card.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="paymentMethodId">ID of payment method to fund.</param>
+        /// <returns>An EasyPost.CreditCardFund instance.</returns>
+        public static async Task<CreditCardFund> Fund(string amount, string paymentMethodId)
```
jchen293, line 74: we should make this function private so user does not access this method directly, thought?

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,124 @@
...
+                    paymentMethod = paymentMethods.secondary_payment_method;
+                    break;
+                default:
+                    break;
+            }
+
+            if (paymentMethod == null)
+            {
+                throw new Exception("The chosen payment method is not a valid method. Please try again.");
+            }
```
jchen293, line 63: This works, but the pattern is little different than the other six CLs as we also check the `paymentMethodId` prefix is `card_`

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,124 @@
...
+                {
+                    "amount", amount
+                }
+            });
+
+            return await request.Execute<CreditCardFund>();
+        }
+
+        /// <summary>
+        ///     Fund a credit card.
```
jchen293, line 88: Let's use the docstring from Python CL, the docstring here should be more explicit. 
```suggestion
        ///     Fund your EasyPost wallet by charging your primary or secondary card on file.
```

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <returns>An EasyPost.CreditCardFund instance.</returns>
+        public static async Task<CreditCardFund> Fund(string amount, Priority priority)
+        {
+            CreditCard paymentMethod = await GetPaymentMethodByPriority(priority);
+
+            return await Fund(amount, paymentMethod.id);
+        }
+
+        /// <summary>
+        ///     Delete a credit card.
```
jchen293, line 101:
```suggestion
        ///     Delete a credit card by ID.
```

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            return await request.Execute();
+        }
+
+        /// <summary>
+        ///     Delete a credit card.
+        /// </summary>
+        /// <param name="priority">Which payment method to delete.</param>
+        /// <returns>Whether the request was successful or not.</returns>
+        public static async Task<bool> Delete(Priority priority)
```
jchen293, line 117: Why do we support deleting a credit card by its priority? This wasn't in the documentation

EasyPost.Tests/PaymentMethodTest.cs
```diff
@@ -0,0 +1,52 @@
...
+        // Skipping due to the lack of an available real credit card in tests.
+        public async Task TestFund()
+        {
+            _vcr.SetUpTest("fund");
+
+            CreditCardFund funding = await CreditCard.Fund("20", CreditCard.Priority.Primary);
+
+            Assert.IsNotNull(funding);
+            Assert.IsInstanceOfType(funding, typeof(CreditCardFund));
+        }
```
jchen293, line 50: these two unit tests can be removed since we moved them to `CreditCardTest.cs`

jchen293 (changes requested):
We shouldn't record cassette for `CreditCard.delete()`

EasyPost/CreditCard.cs
nwithan8: Oh, because there's two functions with different params? I don't know, might be helpful?

EasyPost/CreditCard.cs
nwithan8: Yeah, I added it thinking it would be a helpful "shortcut".
@Justintime50 care to weigh in?

EasyPost/CreditCard.cs
nwithan8: Closing, this relates to https://github.com/EasyPost/easypost-csharp/pull/275#discussion_r889364816

EasyPost/CreditCard.cs
nwithan8: Gotcha, because in the future it could potentially be something other than a card (i.e. a bank account) and something might break.

EasyPost/CreditCard.cs
Justintime50: Although I like the idea, the other 6 libs are already done and not built this way. For now, let’s do it by ID. This will also keep it consistent with how all our other functions for everything work (provide an ID). I’m a fan of revisiting potentially though. Worst case scenario is someone using this simply retrieves their payment methods and makes a second call to delete doing so programmatically.

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        /// <summary>
+        ///     Fund a credit card.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="paymentMethodId">ID of payment method to fund.</param>
+        /// <returns>An EasyPost.CreditCardFund instance.</returns>
+        private static async Task<CreditCardFund> Fund(string amount, string paymentMethodId)
+        {
+            Request request = new Request($"credit_cards/{paymentMethodId}/charge", Method.Post);
```
jchen293, line 76:
```suggestion
            Request request = new Request($"credit_cards/{paymentMethodId}/charges", Method.Post);
```

with the `s`

jchen293 (changes requested):

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,112 @@
...
+        public string exp_month { get; set; }
+        [JsonProperty("exp_year")]
+        public string exp_year { get; set; }
+        [JsonProperty("brand")]
+        public string brand { get; set; }
+
+        /// <summary>
+        ///     Payment method priority
+        /// </summary>
+        public enum Priority
```
Justintime50, line 31: I'd prefer we kept variable names consistent with the other libs. `PrimaryOrSecondary`? Same goes for the param below.

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,112 @@
...
+                case Priority.Secondary:
+                    paymentMethod = paymentMethods.secondary_payment_method;
+                    break;
+                default:
+                    break;
+            }
+
+            if (paymentMethod == null || !paymentMethod.id.StartsWith("card_"))
+            {
+                throw new Exception("The chosen payment method is not a valid method. Please try again.");
```
Justintime50, line 62: Same here, consistency is important.
```suggestion
                throw new Exception("The chosen payment method is not a credit card. Please try again.");
```

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,112 @@
...
+            return await request.Execute<CreditCardFund>();
+        }
+
+        /// <summary>
+        ///     Fund a credit card.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="priority">Which type of payment method to fund.</param>
+        /// <returns>An EasyPost.CreditCardFund instance.</returns>
+        public static async Task<CreditCardFund> Fund(string amount, Priority priority)
```
Justintime50, line 93: Renaming this particular param (since it's the public one) may be the better one to rename from `priority`?

EasyPost/CreditCard.cs
```diff
@@ -0,0 +1,112 @@
...
+            return paymentMethod;
+        }
+
+        /// <summary>
+        ///     Fund a credit card.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="paymentMethodId">ID of payment method to fund.</param>
+        /// <returns>An EasyPost.CreditCardFund instance.</returns>
+        private static async Task<CreditCardFund> Fund(string amount, string paymentMethodId)
```
Justintime50, line 74: I don't feel having two competing interfaces for the same thing is a good idea. There should typically only be one rock solid way of doing something. Since the pattern we chose was for the user to pass an amount and then which of the two payment methods on file they want to use, we should keep that consistent and remove this version of the function that accepts an ID completely.

EasyPost/Request.cs
```diff
@@ -16,11 +16,11 @@ public class Request
         private readonly Dictionary<string, object> _urlSegments;
         public string? RootElement { get; set; }
 
-        public Request(string iResource, Method method, Dictionary<string, object>? parameters = null)
```
Justintime50, line 19: What prompted this change? Is there a downside to removing this?

Justintime50 (dismissed):

EasyPost/CreditCard.cs
nwithan8: So `PrimaryOrSecondary.Secondary`?

EasyPost/CreditCard.cs
nwithan8: This method is private, so is not available to the end user to call directly.

EasyPost/Request.cs
nwithan8: No downside. `.AddParameters()` is the preferred way to add parameters to a `Request` object, but I did add the ability to pass params to the constructor when refactoring this a few weeks ago. Noticed our library was about half-half on which it used, so decided to remove the confusion and make only one way to do it.

EasyPost.Tests/CreditCardTest.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            _vcr = new TestUtils.VCR("credit_card", TestUtils.ApiKey.Production);
+        }
+
+        [Ignore]
+        [TestMethod]
+        // When recording, can only record in one framework, then manually copy cassette to other frameworks.
```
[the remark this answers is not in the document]
Justintime50: Ah. Per the other libs and decision, we won't record this function. Test it the once to make sure it works and we'll circle back to this.

Thanks for explaining this.

EasyPost/CreditCard.cs
Justintime50: Ah, I missed that.

EasyPost/CreditCard.cs
nwithan8: Changed

Justintime50 (dismissed):
There still exists a bunch of empty `delete` cassettes for PaymentMethod?

PR #278, opened 2022-06-13
Clean up build warnings

nwithan8:
# Description

- Suppress nullability warnings
- Suppress obsolete function warnings

# Testing

All unit tests run as expected

Remaining errors:
image

# Pull Request Type

### DIFF
--- a/EasyPost/EasyPost.csproj
+++ b/EasyPost/EasyPost.csproj
@@ -22,6 +22,8 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <!-- Nullability and other warnings to suppress -->
+    <NoWarn>CS8618;CS8603;CS8602;CS8765;CS8767;CS0659;CS0618</NoWarn>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|AnyCPU'">
     <DebugType>pdbonly</DebugType>
@@ -31,7 +33,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <DocumentationFile>..\lib\net\EasyPost.XML</DocumentationFile>
-    <NoWarn>1701;1702;1591;1570</NoWarn>
+    <NoWarn>1701;1702;1591;1570;CS8618;CS8603;CS8602;CS8765;CS8767;CS0659;CS0618</NoWarn>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
     <SignAssembly>false</SignAssembly>
--- a/EasyPost/EasyPost.csproj
+++ b/EasyPost/EasyPost.csproj
@@ -23,6 +23,13 @@
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
     <!-- Nullability and other warnings to suppress -->
+    <!-- CS8618: Non-nullable field not initialized -->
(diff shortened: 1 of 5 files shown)


EasyPost/Address.cs
```diff
@@ -8,21 +8,21 @@ namespace EasyPost
     public class Address : Base.Address
     {
         [JsonProperty("carrier_facility")]
-        public string carrier_facility { get; set; }
+        public string carrier_facility { get; set; } = null!;
```
Justintime50, line 11: I thought we were not going to do nullability now and instead ignore them?

nwithan8: This is how you ignore them. The `!` suppresses the warning.

Justintime50: Mmm, I had something in mind more like a one or two line config tweak to globally ignore vs adding fluff to hundreds of lines. Can we instead do that?

nwithan8: Updated

EasyPost/EasyPost.csproj
```diff
@@ -22,6 +22,8 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <!-- Nullability and other warnings to suppress -->
+    <NoWarn>CS8618;CS8603;CS8602;CS8765;CS8767;CS0659;CS0618</NoWarn>
```
Justintime50, line 26: I'd suggest adding comments for what each of these are (can group `CS8618;CS8603;CS8602;CS8765;CS8767` in a single comment stating nullability).

I would like to see internal tasks to fix the three explicit items here we are suppressing (nullability, overrides, and obsoletes). I don't want these a year from now to still be ignored.

EasyPost/Utilities/Rates.cs
```diff
@@ -64,12 +64,13 @@ public static Rate GetLowestObjectRate(List<Rate> rates, List<string>? includeCa
                 float rateValue = float.Parse(rate.rate);
                 float lowestRateValue = float.Parse(lowestRate.rate);
 
-                // if this rate is lower than the lowest rate, set it to this rate
-                if (rateValue < lowestRateValue)
+                // if this rate is higher than or the same as the lowest rate, skip it
+                if (lowestRateValue <= rateValue)
```
Justintime50, line 68: Please do not alter this as it will break someones integration. With a particular carrier mix, customers have come to rely on the behavior of "last lowest rate gets returned". It's a terrible thing to rely on, but I promise you someone will right in about this if we alter it (even if this is a smart altering to do less OR equals).

EasyPost/Utilities/Rates.cs
```diff
@@ -107,12 +108,13 @@ public static Smartrate GetLowestShipmentSmartrate(List<Smartrate> smartrates, i
                     continue;
                 }
 
-                // if this smartrate is lower than the lowest smartrate, set it to this smartrate
-                if (smartrate.rate < lowestSmartrate.rate)
+                // if this smartrate is higher or the same as the lowest smartrate, skip it
+                if (lowestSmartrate.rate <= smartrate.rate)
```
Justintime50, line 112: Same comment as above.

Justintime50 (dismissed):

Justintime50 (commented):
There's still 1 IDE warning in the `GetLowestShipmentSmartrate` function for the lowestSmartrate being null.

There also appear to be 4 warnings remaining when the solution is built related to CS8600.

Justintime50 (approved):
And just like that, everything is beautiful.

Please create internal tasks to fix these items that are suppressed.

jchen293 (approved):

PR #280, opened 2022-06-27
Support for webhook secrets

nwithan8:
# Description

- Allow passing parameters during webhook update
- ~~New unit test to test webhook secret (disabled, see inline comment)~~
- ~~New webhook service used for webhook unit testing and verification~~
- Fix some docstrings

# Testing

- ~~Re-recorded webhook cassettes to use new webhook service~~
- All tests pass

# Pull Request Type


Justintime50 (commented):
Seems fine.

My only point of feedback is the testing here is too involved. Unit tests are meant to be fast and small. We don't need to spin up a whole service here. We should end to end test this to make sure it works, but otherwise a `webhook_secret` is no different than a normal parameter and doesn't necessitate its own unit test because it is not its own "unit". If you've verified that this works with your setup here, I'd argue that we don't need a separate unit test for this or the webhook service that got added here as it greatly complicates the test suite and is unnecessary - secret testing would then be verified and ensured at the API level since all this lib does is pass a string.

nwithan8:
@Justintime50 The `.gitattribute` change to ignore cassettes doesn't seem to be working, if you want to take a look at it.

jchen293 (approved):

Justintime50 (approved):

PR #282, opened 2022-06-29
removed CreditCardFund

ryannguyen16:
# Description
removed the  CreditCardFund class and edited the test class to work as well.

# Testing

# Pull Request Type


EasyPost.Tests/CreditCardTest.cs
```diff
@@ -36,10 +36,10 @@ public async Task TestFund()
         {
             _vcr.SetUpTest("fund");
 
-            CreditCardFund funding = await CreditCard.Fund("20", CreditCard.Priority.Primary);
+            bool funding = await CreditCard.Fund("20000", CreditCard.Priority.Primary);
 
             Assert.IsNotNull(funding);
```
nwithan8, line 41: Delete this line.

Rename `funding` to `success` in this unit test to be more explicit about what it is.

nwithan8 (approved):

EasyPost/CreditCard.cs
```diff
@@ -81,7 +81,7 @@ private static async Task<CreditCardFund> Fund(string amount, string paymentMeth
                 }
             });
 
-            return await request.Execute<CreditCardFund>();
+            return await request.Execute();
```
Justintime50, line 84: @nwithan8 wouldn't we want to return the bool only after executing instead of returning the request itself.

nwithan8: `Request.Execute()` (without a class type) doesn't attempt to serialize the JSON response into any object. Instead, it returns a boolean based on whether an error was thrown during the HTTP request:
image

So, the result of `await request.Execute()` will be `true` if the API call was successful, `false` if it wasn't (a 4XX or 5XX was thrown)

nwithan8: This was added a while ago because a lot of our set-and-forget API calls (mostly DELETE calls) were hard-coded to return `true`, even if an HTTP error occurred during the API call.
Now, the return value is based on whether the API call was successful.

Justintime50: I remember now, thanks!

Justintime50 (approved):

PR #283, opened 2022-06-29
Capture OS details in User-Agent

nwithan8:
# Description

- Capture OS details in User-Agent
- Removed outdated, unused X-Client-User-Agent header
- Improve user agent detail calculation logic

# Testing

- Implementation works on all supported versions of .NET. Examples below (notice the different `.NET` value):
image

image

image

image

- All tests pass as expected

# Notes
- The legacy .NET Framework does not have pretty version numbers (will look something like 4.0.xxxx.xxxx), but it thankfully sticks out compared to the other .NET/.NET Core versions (which follow the X.Y.Z pattern)
- In the screenshots above, the `OSVersion` for .NET 5 and .NET 6 report "12.0.1", while .NET Core 3.1 and .NET Framework report "21.1.0.0". This is due to a change [introduced in .NET 5](https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=net-6.0#remarks) that improved the accuracy of OS version reporting. To make sense of this, .NET Framework and .NET Core 3.1 are seeing "Darwin 21.1.0.0", while .NET 5 and .NET 6 are seeing "MacOS 12.0.1". These are [the same thing](https://en.wikipedia.org/wiki/MacOS_Monterey#Release_history) ultimately.

# Pull Request Type

### DIFF
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -49,7 +49,7 @@ namespace EasyPost
 
             _libraryVersion = RuntimeInfo.ApplicationInfo.ApplicationVersion;
             _dotNetVersion = RuntimeInfo.ApplicationInfo.DotNetVersion;
-            _osDetails = RuntimeInfo.OperationSystemInfo.UserAgentOsDetails;
+            _osDetails = $"OS/{RuntimeInfo.OperationSystemInfo.Name} OSVersion/{RuntimeInfo.OperationSystemInfo.Version} OSArch/{RuntimeInfo.OperationSystemInfo.Architecture}";
 
             RestClientOptions clientOptions = new RestClientOptions
             {
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -17,7 +17,9 @@ namespace EasyPost
 
         private readonly string _dotNetVersion;
         private readonly string _libraryVersion;
-        private readonly string _osDetails;
+        private readonly string _osArch;
+        private readonly string _osName;
+        private readonly string _osVersion;
 
         private readonly RestClient _restClient;
         private int? _connectTimeoutMilliseconds;
@@ -35,7 +37,7 @@ namespace EasyPost
             set => _requestTimeoutMilliseconds = value;
         }
 
-        private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion} {_osDetails}";
+        private string UserAgent => $"EasyPost/v2 CSharpClient/{_libraryVersion} .NET/{_dotNetVersion} OS/{_osName} OSVersion/{_osVersion} OSArch/{_osArch}";
 
         /// <summary>
         ///     Constructor for the EasyPost client.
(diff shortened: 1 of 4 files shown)


EasyPost/Utilities/RuntimeInfo.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            /// <summary>
+            ///     Get the architecture of the operating system.
+            /// </summary>
+            /// <returns>Architecture of the operating system.</returns>
+            internal static string Architecture
+            {
+                get
+                {
+                    // sorry, ARM users
```
Justintime50, line 122: Question: Why are we not supporting ARM? Why not just pass through whatever C# says instead of a string?

EasyPost/Utilities/RuntimeInfo.cs
```diff
@@ -0,0 +1,128 @@
...
+#if NET45
+                    switch (OperatingSystem.Platform)
+                    {
+                        case PlatformID.Win32S:
+                        case PlatformID.Win32Windows:
+                        case PlatformID.Win32NT:
+                        case PlatformID.WinCE:
+                            return "Windows";
+                        case PlatformID.Xbox:
+                            return "Xbox"; // yes, really
```
Justintime50, line 79: Issue: No.

EasyPost/Utilities/RuntimeInfo.cs
```diff
@@ -0,0 +1,128 @@
...
+                get { return Environment.OSVersion; }
+            }
+
+            /// <summary>
+            ///     Get user-agent pre-formatted operating system details.
+            /// </summary>
+            /// <returns>Operating system details pre-formatted for the User-Agent header.</returns>
+            internal static string UserAgentOsDetails
+            {
+                get { return $"OS/{Name} OSVersion/{Version} OSArch/{Architecture}"; }
```
Justintime50, line 59: suggestion: I would rather build these directly on the string where the others are getting added instead of abstract these into a function. It makes it more difficult to determine what the final user-agent header will be when reading the code.

EasyPost/Utilities/RuntimeInfo.cs
nwithan8: Tweaked

EasyPost/Utilities/RuntimeInfo.cs
nwithan8: No fun.

EasyPost/Utilities/RuntimeInfo.cs
nwithan8: Resolved

jchen293 (approved):

EasyPost/Client.cs
```diff
@@ -48,18 +47,9 @@ public Client(ClientConfiguration clientConfiguration, HttpClient? customHttpCli
...
-            }
-            catch (Exception)
-            {
-                _libraryVersion = "Unknown";
-            }
-
-            _dotNetVersion = Environment.Version.ToString();
+            _libraryVersion = RuntimeInfo.ApplicationInfo.ApplicationVersion;
+            _dotNetVersion = RuntimeInfo.ApplicationInfo.DotNetVersion;
+            _osDetails = $"OS/{RuntimeInfo.OperationSystemInfo.Name} OSVersion/{RuntimeInfo.OperationSystemInfo.Version} OSArch/{RuntimeInfo.OperationSystemInfo.Architecture}";
```
Justintime50, line 52: suggestion: what I meant was inline on line 38. Having each individual piece be modular so we can easily adjust if needed. So assigning them to a var like the lib version and then building each delimited piece of that string one by one so the engineer looking at it knows exactly what's included without needing to go elsewhere.

Justintime50 (approved):

Justintime50:
CHANGELOG entry needed too

Justintime50 (approved):

PR #286, opened 2022-07-01
Support for bank accounts in billing

nwithan8:
# Description

- Support for bank account in billing options
- Move PaymentMethod and CreditCard functions to Billing
- New PaymentMethodObject that can be either a credit card or bank account
- Move and update unit tests

Per our offline conversation, the new `PaymentMethodObject` (open to renames) has all the properties to accept either a credit card or a bank account JSON from the server (since we don't know which we're going to get). The `Type` property on it returns an enum to the user based on the ID prefix.

# Testing

- Unit tests updated and unified under `PaymentMethodTest` file
- All tests pass as expected

# Pull Request Type


EasyPost/Billing.cs
```diff
@@ -0,0 +1,93 @@
...
+        ///     Fund your wallet from a specific payment method.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="primaryOrSecondary">Which type of payment method to use to fund.</param>
+        /// <returns>True if successful, false otherwise.</returns>
+        public static async Task<bool> FundWallet(string amount, PaymentMethod.Priority primaryOrSecondary)
+        {
+            PaymentMethodObject paymentMethod = await GetPaymentMethodByPriority(primaryOrSecondary);
+
+            Request request = new Request($"credit_cards/{paymentMethod.id}/charges", Method.Post);
```
jchen293, line 32:
```suggestion
            Request request = new Request($"this can be either bank account or CC/{paymentMethod.id}/charges", Method.Post);
```

You need to add logic to check if the payment method is a `bank account` or a `credit card`, and update the request URL

EasyPost/Billing.cs
```diff
@@ -0,0 +1,93 @@
...
+    public static class Billing
+    {
+        /// <summary>
+        ///     Delete a payment method.
+        /// </summary>
+        /// <param name="paymentMethodId">ID of the payment method to delete.</param>
+        /// <returns>Whether the request was successful or not.</returns>
+        public static async Task<bool> DeletePaymentMethod(string paymentMethodId)
+        {
+            Request request = new Request($"credit_cards/{paymentMethodId}", Method.Delete);
```
jchen293, line 17:
```suggestion
            Request request = new Request($"this can be either bank account or CC/{paymentMethodId}", Method.Delete);
```

You need to check if the delete payment method is CC or bank account

EasyPost/Billing.cs
```diff
@@ -0,0 +1,93 @@
...
+                    paymentMethod = paymentMethods.secondary_payment_method;
+                    break;
+                default:
+                    break;
+            }
+
+            if (paymentMethod == null || !paymentMethod.id.StartsWith("card_"))
+            {
+                throw new Exception("The chosen payment method is not a credit card. Please try again.");
+            }
```
jchen293, line 88: What if the payment method is bank account that start with `bank_`

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,100 @@
...
+                if (id.StartsWith("bank_"))
+                {
+                    return PaymentMethodType.BankAccount;
+                }
+
+                return null;
             }
+        }
+
+        [JsonProperty("verified")]
```
jchen293, line 88: Why is this JsonProperty isolated from above?

EasyPost/Billing.cs
Justintime50, line 17: Issue: The whole point of consolidating these is to do a conditional check on what payment method was passed in. If a user passes in a BankAccount ID here, this will fail because it needs to be routed to the `bank_accounts` endpoint instead.

We should check what kind of ID is passed and change the URL accordingly

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,100 @@
...
-        public static async Task<PaymentMethod> All()
+        public PaymentMethodType? Type
         {
-            Request request = new Request("payment_methods", Method.Get);
+            get
+            {
+                if (id == null)
+                {
+                    return null;
+                }
```
jchen293, line 72: This can be removed

EasyPost/Billing.cs
Justintime50, line 32: Issue: The whole point of consolidating these is to do a conditional check on what payment method was passed in. If a user passes in a BankAccount ID here, this will fail because it needs to be routed to the `bank_accounts` endpoint instead.

We should check what kind of ID is passed and change the URL accordingly

jchen293 (dismissed):

Justintime50 (dismissed):

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,100 @@
...
         [JsonProperty("primary_payment_method")]
-        public CreditCard primary_payment_method { get; set; }
+        public PaymentMethodObject primary_payment_method { get; set; }
         [JsonProperty("secondary_payment_method")]
-        public CreditCard secondary_payment_method { get; set; }
+        public PaymentMethodObject secondary_payment_method { get; set; }
+    }
+
+    /// <summary>
+    ///     Represents a credit card or a bank account.
```
Justintime50, line 31: Issue: I'd like an extra comment in here describing the **why** we have to do this

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,100 @@
...
+        [JsonProperty("exp_year")]
+        public string exp_year { get; set; }
+        [JsonProperty("id")]
+        public string id { get; set; }
+        [JsonProperty("last4")]
+        public string last4 { get; set; }
+        [JsonProperty("name")]
+        public string name { get; set; }
+        [JsonProperty("object")]
+        public string Object { get; set; }
```
Justintime50, line 60: Issue: Let's add inline comments here describing which one each belongs to

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,100 @@
...
 
-            return paymentMethod;
+        /// <summary>
+        ///     Delete this payment method from the user's account.
+        /// </summary>
+        /// <returns>True if successful, false otherwise.</returns>
+        public async Task<bool> Delete()
+        {
+            return await Billing.DeletePaymentMethod(id);
         }
```
Justintime50, line 98: Issue: Let's remove this (for now) and when we revisit data flow, we can approach this holistically

nwithan8: Resolved

EasyPost/PaymentMethod.cs
nwithan8: This safeguards in case somehow the user has created a PaymentMethodObject object without initializing it.

jchen293: I mean you have a return statement that return `null` down below, this if-statement is redundant

EasyPost/Billing.cs
nwithan8: Resolved

EasyPost/PaymentMethod.cs
nwithan8: I think if I tried to call `StartsWith` on a `null` value, it would throw an error, so this prevents that.

EasyPost/Billing.cs
nwithan8: Resolved.

EasyPost/PaymentMethod.cs
nwithan8: Resolved.

EasyPost.Tests/PaymentMethodTest.cs
```diff
@@ -19,10 +19,39 @@ public async Task TestAll()
...
+        public async Task TestDeleteCreditCard()
+        {
+            _vcr.SetUpTest("delete_credit_card");
+
+            PaymentMethod paymentMethod = await Billing.RetrievePaymentMethods();
+            PaymentMethodObject creditCard = paymentMethod.primary_payment_method;
+
+            bool success = await creditCard.Delete();
+            // ^ instance method, can also do below
+            // bool success = await Billing.DeletePaymentMethod(creditCard.id);
```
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/PaymentMethod.cs
nwithan8: Resolved.

EasyPost/PaymentMethod.cs
jchen293: Should we move this to the top?

build.bat
```diff
@@ -0,0 +1,85 @@
+:: This script will build a .NET project in Release mode, sign the generated DLLs with a provided PFX certificate,
```
jchen293, line 1: why is this file here?

EasyPost.Tests/PaymentMethodTest.cs
```diff
@@ -14,12 +14,41 @@ public void Initialize()
             _vcr = new TestUtils.VCR("payment_method", TestUtils.ApiKey.Production);
```
jchen293, line 14: This test file need to be changed to `BillingTest.cs`

build.bat
nwithan8: Rebase gone bad, fixed.

EasyPost/PaymentMethod.cs
nwithan8: The getter/setter properties are sorted alphabetically, so the `Type` property is being organized above this, unfortunately. I believe our linter will complain if I were to force it to the top.

EasyPost.Tests/PaymentMethodTest.cs
nwithan8: Resolved.

EasyPost/Billing.cs
```diff
@@ -0,0 +1,116 @@
...
+            return await FundWallet(amount, paymentMethod);
+        }
+
+        /// <summary>
+        ///     Fund your wallet from a specific payment method.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="paymentMethodObject">Payment method to use to fund the wallet.</param>
+        /// <returns>True if successful, false otherwise.</returns>
+        public static async Task<bool> FundWallet(string amount, PaymentMethodObject paymentMethodObject)
```
Justintime50, line 53: Issue: We should be abstracting away the need for the user to pass us a whole payment method. Instead, this function should accept `primaryOrSecondary` and adjust the endpoint and logic accordingly.

Justintime50 (changes requested):

EasyPost/Billing.cs
```diff
@@ -0,0 +1,116 @@
...
+        ///     Delete a payment method.
+        /// </summary>
+        /// <param name="paymentMethodObject">Payment method to delete.</param>
+        /// <returns>Whether the request was successful or not.</returns>
+        public static async Task<bool> DeletePaymentMethod(PaymentMethodObject paymentMethodObject)
+        {
+            Request request = new Request($"{paymentMethodObject.Endpoint}/{paymentMethodObject.id}", Method.Delete);
+
+            return await request.Execute();
+        }
```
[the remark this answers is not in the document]
nwithan8: Removed

EasyPost/Billing.cs
nwithan8: Removed

EasyPost/Billing.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        /// <summary>
+        ///     Fund your wallet from a specific payment method.
+        /// </summary>
+        /// <param name="amount">Amount to fund.</param>
+        /// <param name="priority">Which type of payment method to use to fund the wallet.</param>
+        /// <returns>True if successful, false otherwise.</returns>
+        public static async Task<bool> FundWallet(string amount, PaymentMethod.Priority priority)
```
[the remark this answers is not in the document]
nwithan8: Resolved

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,119 @@
...
         [JsonProperty("secondary_payment_method")]
-        public CreditCard secondary_payment_method { get; set; }
+        public PaymentMethodObject secondary_payment_method { get; set; }
+    }
+
+    /// <summary>
+    ///     Represents a credit card or a bank account.
+    ///     Have to collect all possible properties since we can't anticipate which type of payment method the API will return.
+    /// </summary>
+    public class PaymentMethodObject : Resource
```
Justintime50, line 34: Thought: I was looking over Go this morning and we actually do the same thing there on a couple objects so there is precedent to do this.

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,119 @@
...
 
-            if (paymentMethod.id == null)
-            {
-                throw new Exception("Billing has not been setup for this user. Please add a payment method.");
+                if (id.StartsWith("bank_"))
+                {
+                    return PaymentMethodType.BankAccount;
+                }
+
+                return null;
```
Justintime50, line 95: Issue: Only one return statement please instead of 4 in one function. To Jack's point, I feel there is potential for some refactoring here to clean it up a bit.

EasyPost/PaymentMethod.cs
Justintime50: Oh my... please look into fixing this. I don't want to propagate bad patterns like this.

Justintime50 (dismissed):
Code is good, one block needs a bit of work and I'd like us to look into the wonky linter error.

EasyPost/PaymentMethod.cs
nwithan8: Type-strictness FTW

EasyPost/PaymentMethod.cs
nwithan8: There's nothing we can do, since it's all getter/setters

EasyPost/PaymentMethod.cs
nwithan8: I'm confused why this is giving you pause. This is effectively a switch statement, returning a value based on what the ID prefix is: `CreditCard` if it's a credit card, `BankAccount` if it's a bank account or `null` if the ID either is not set or is not one of the two expected ID types.

EasyPost/PaymentMethod.cs
nwithan8: Resolved using [region markers ](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/directives/region-directive) to keep elements grouped together, even during linting

Justintime50: Looove this! Thanks for going the extra mile to correct this.

EasyPost/PaymentMethod.cs
```diff
@@ -1,38 +1,127 @@
...
+                }
+
+                return Type.EndPoint;
             }
+        }
+
+        public class PaymentMethodType : Enumeration
+        {
+            public static readonly PaymentMethodType BankAccount = new PaymentMethodType(2, "bank_accounts");
+            public static readonly PaymentMethodType CreditCard = new PaymentMethodType(1, "credit_cards");
```
jchen293, line 80: Why is the id set up `1` and `2`? 

Also, did you E2E the test?

Justintime50 (approved):

jchen293 (approved):

PR #289, opened 2022-07-01
Makefile and build scripts

nwithan8:
# Description
- New Makefile with common commands
- Split build.bat into multiple smaller Batch scripts (allow to be run individually or via make)
- Fix compile tests in F# and VB unit tests

# Testing
- All Unix-compatible scripts work.
- Will test on Windows machine.

# Pull Request Type

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,13 +14,6 @@ jobs:
       - uses: actions/checkout@v3
       - name: Check dotnet Style
         run: dotnet-format --check --exclude /
-  Validate_Scripts:
-    runs-on: windows-2022
-    steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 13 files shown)


Justintime50 (commented):
Windows doesn't have access to `make` out of the box whereas many unix systems do. You'll want to install `make` on Windows and ensure these commands are cross compatible, otherwise there isn't much worth in adding these in: https://community.chocolatey.org/packages/make

I'll defer reviewing till Windows support has been confirmed as I'm sure there are a couple tweaks left.

Justintime50 (commented):
All is fine; however, I'd prefer that `check_scripts` is renamed to something like `lint_scripts` so it is more easier for searching, understanding, and uses the same language we are accustomed to.

PR #308, opened 2022-08-03
chore: unifies Makefile and adds release target

Justintime50:
# Description

Unified the Makefile like we have in other libs, adds a release target.

# Pull Request Type

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,46 @@
 help:
 	@cat Makefile | grep '^## ' --color=never | cut -c4- | sed -e "`printf 's/ - /\t- /;'`" | column -s "`printf '\t'`" -t
 
+## build - Build the project in Debug mode
+build:
+	dotnet msbuild -property:Configuration="Debug" -target:Rebuild -restore
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -2,6 +2,46 @@
 help:
 	@cat Makefile | grep '^## ' --color=never | cut -c4- | sed -e "`printf 's/ - /\t- /;'`" | column -s "`printf '\t'`" -t
 
+## build - Build the project in Release mode
+build:
+	dotnet msbuild -property:Configuration="Debug" -target:Rebuild -restore
```
nwithan8, line 7: Debug and Release are flipped

nwithan8 (changes requested):

Makefile
Justintime50: The config itself was intentional. The command `build` should be the generic debug since people would be running that much more than a production build which should require a separate target; however, the comments are definitely off. Fixed.

nwithan8 (approved):

jchen293 (approved):

PR #309, opened 2022-08-05
Thread-safe rearchitecture

### LINKED ISSUE #98: does not work in mult-threaded, multi-apiKey environments

nwithan8:
# Description

- Services (static functions) called via a property of a Client object
- Non-static functions called via an instance of an object
- Files re-organized for better readability
- All unit tests reworked to account for new structure, all based on base UnitTest (using XUnit rather than MSTest)

Closes #98 

# Testing

- All unit tests pass
- All cassettes re-recorded

# Pull Request Type


Justintime50:
This PR needs to be put up against the `v4` branch, not master.

EasyPost/Utilities/General.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <param name="action">Action to do for each loop.
+        /// Action receives a (int, T) tuple containing the index and the current element in the loop as a parameter</param>
+        /// <typeparam name="T">Type of input.</typeparam>
+        public static void Each<T>(this IEnumerable<T> ie, Action<int, T> action)
+        {
+            int i = 0;
+            foreach (var e in ie) action(i++, e);
+        }
+    }
+}
```
Justintime50, line 21: Issue: Let's remove this unused code.

EasyPost.Tests/AddressTest.cs
```diff
@@ -1,147 +1,127 @@
...
-
-            Assert.IsInstanceOfType(address, typeof(Address));
-            Assert.IsTrue(address.id.StartsWith("adr_"));
-            Assert.AreEqual("388 Townsend St", address.street1);
         }
 
-        [TestMethod]
-        public async Task TestCreateVerify()
+        [Fact]
+        public async Task TestAll()
```
Justintime50, line 15: Question: Can we keep these CRUD ordered so they match the other libs? Otherwise there will be a bit more searching to be done to find stuff. 

This is true of the order of all tests.

EasyPost.Tests/AddressTest.cs
```diff
@@ -1,147 +1,127 @@
...
+                }
+            });
 
-            Address address = await Address.Create(addressData);
+            List<Address> addresses = addressCollection.addresses;
 
-            Assert.IsInstanceOfType(address, typeof(Address));
-            Assert.IsTrue(address.id.StartsWith("adr_"));
-            Assert.AreEqual("417 MONTGOMERY ST FL 5", address.street1);
+            Assert.True(addresses.Count <= Fixture.PageSize);
```
Justintime50, line 28: Issue: It appears some of your assertions have disappeared (please assert it's `true`):

`Assert.IsNotNull(addressCollection.has_more);`

I think maybe this is true of many of the tests. Please 1:1 double check every test, I'd rather not dig deep to make sure every assertion matches what was there before, that should implicitly be the expectation that the test assertions perfectly match what is on master and what is making it into your PR.

Justintime50 (changes requested):
I will need to review each file individually over the next few days, I left a few high level items that need addressing. Please address anything else you find while looking prior to pinging for review again at which point I'll do a real review.

Beta client stuff is still TBD.

EasyPost/Models/API/Address.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        #region JSON Properties
+
+        [JsonProperty("city")]
+        public string city { get; set; }
+        [JsonProperty("company")]
+        public string company { get; set; }
+        [JsonProperty("country")]
+        public string country { get; set; }
+
```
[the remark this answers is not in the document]
nwithan8: Resolved

EasyPost/_base/EasyPostService.cs
```diff
@@ -0,0 +1,38 @@
...
+    public abstract class EasyPostService : WithClient, IEasyPostService
+    {
+        internal EasyPostService(EasyPostClient client)
+        {
+            Client = client;
+        }
+
+        protected async Task<T> Create<T>(string url, Dictionary<string, object?>? parameters = null, string? rootElement = null, ApiVersion? apiVersion = null) where T : class, new() => await Request<T>(Method.Post, url, parameters, rootElement, apiVersion);
+
+        protected async Task CreateBlind(string url, Dictionary<string, object?>? parameters = null, ApiVersion? apiVersion = null) => await Request(Method.Post, url, parameters, apiVersion);
```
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/Utilities/General.cs
nwithan8: Resolved.

README.md
```diff
@@ -32,7 +32,7 @@ namespace example
         {
             EasyPost.ClientManager.SetCurrent(Environment.GetEnvironmentVariable("EASYPOST_API_KEY"));
 
-            Dictionary<string, object> fromAddress = new Dictionary<string, object>() {
+            Dictionary<string, object?> fromAddress = new Dictionary<string, object?>() {
```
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost.Tests/AddressTest.cs
nwithan8: Resolved using annotations and file layout rules. These will automatically be sorted and remain sorted during the code formatting process moving forward.

EasyPost.Tests/AddressTest.cs
nwithan8: Resolved.

nwithan8:
Good news, in my journey to fix an odd bug for .NET Standard build process, all unit tests now take 1-2 seconds to run. Of course, there's still the several-minute process of building everything prior to running unit tests, but hey, progress.

nwithan8:
@Justintime50 This is ready for review, pending your decision on how to handle beta features.

nwithan8:
NOTE: We'll need to move EndShipper to GA either in this PR or a future PR, once it's enabled server-side.

EasyPost.Tests/AddressTest.cs
```diff
@@ -1,147 +1,139 @@
...
 
-            AddressCollection addressCollection = await Address.All(new Dictionary<string, object>
+            AddressCollection addressCollection = await Client.Address.All(new Dictionary<string, object>
             {
                 {
                     "page_size", Fixture.PageSize
                 }
             });
-
+            Assert.True(addressCollection.HasMore);
```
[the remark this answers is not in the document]
nwithan8: Updated

EasyPost.Tests/AddressTest.cs
```diff
@@ -1,147 +1,139 @@
...
-        [TestMethod]
+        [Fact]
+        [CrudOperations.Create]
         public async Task TestCreateVerify()
         {
-            _vcr.SetUpTest("create_verify");
+            UseVCR("create_verify");
 
-            Dictionary<string, object> addressData = Fixture.IncorrectAddressToVerify;
-            addressData.Add("verify", true);
```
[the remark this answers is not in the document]
nwithan8: It is currently, we can address this when we pull in the fixtures.

EasyPost.Tests/CarrierAccountTest.cs
```diff
@@ -1,131 +1,133 @@
...
         }
 
-        [TestMethod]
+        [Fact]
+        [CrudOperations.Update]
         public async Task TestUpdate()
         {
-            _vcr.SetUpTest("update");
+            UseVCR("update");
+
```
Justintime50, line 88: Nit: Extra line unneeded

EasyPost.Tests/CarrierAccountTest.cs
```diff
@@ -1,131 +1,133 @@
...
+            UseVCR("delete");
 
-            Assert.IsInstanceOfType(carrierAccount, typeof(CarrierAccount));
-            Assert.IsTrue(carrierAccount.id.StartsWith("ca_"));
-            Assert.AreEqual(testDescription, carrierAccount.description);
+            CarrierAccount carrierAccount = await CreateBasicCarrierAccount();
+
+            await carrierAccount.Delete();
+
+            // TODO: Assert something
```
Justintime50, line 118: Issue: Let's assert something now. Whatever this returns at present, and we can update the assertions later when we fix deleting. A test with no assertion is not a test.

EasyPost.Tests/EasyPost.Tests.csproj
```diff
@@ -9,16 +9,20 @@
...
-    <PackageReference Include="MSTest.TestFramework" Version="2.2.8"/>
-    <PackageReference Include="MSTest.TestAdapter" Version="2.2.8"/>
+    <PackageReference Include="EasyVCR" Version="0.4.0"/>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Include="coverlet.collector" Version="1.2.0" />
+    <PackageReference Include="Microsoft.VisualStudio.TestPlatform.ObjectModel" Version="14.0.0" />
+    <PackageReference Include="xunit" Version="2.4.2" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="[13.0.1, 14.0.0)" />
+    <PackageReference Include="RestSharp" Version="[107.3.0, 108.0.0)" />
```
Justintime50, line 25: Issue: These should all be pinned versions

EasyPost.Tests.FSharp/EasyPost.Tests.FSharp.fsproj
```diff
@@ -14,9 +14,10 @@
...
     <ItemGroup>
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0"/>
-        <PackageReference Include="MSTest.TestAdapter" Version="2.2.8"/>
-        <PackageReference Include="MSTest.TestFramework" Version="2.2.8"/>
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+        <PackageReference Include="coverlet.collector" Version="1.2.0" />
+        <PackageReference Include="xunit" Version="2.4.2" />
+        <PackageReference Include="xunit.runner.visualstudio" Version="2.4.2" />
         <PackageReference Include="SecurityCodeScan.VS2019" Version="[5.0.0, 6.0.0)">
             <PrivateAssets>all</PrivateAssets>
```
Justintime50, line 22: Issue: These should all be pinned versions.

EasyPost.Tests.VB/EasyPost.Tests.VB.vbproj
```diff
@@ -10,9 +10,10 @@
...
 
     <ItemGroup>
-      <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0"/>
-      <PackageReference Include="MSTest.TestAdapter" Version="2.2.8"/>
-      <PackageReference Include="MSTest.TestFramework" Version="2.2.8"/>
+      <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+      <PackageReference Include="coverlet.collector" Version="1.2.0" />
+      <PackageReference Include="xunit" Version="2.4.2" />
+      <PackageReference Include="xunit.runner.visualstudio" Version="2.4.2" />
       <PackageReference Include="SecurityCodeScan.VS2019" Version="[5.0.0, 6.0.0)">
```
Justintime50, line 17: Issue: These should all be pinned versions

Question: Why was `Microsoft.NET.Test.Sdk` downgraded?

EasyPost.Tests/EasyPostTest.cs
```diff
@@ -1,62 +1,21 @@
-using System.Dynamic;
-using System.Threading.Tasks;
-using EasyPost.Utilities;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-using RestSharp;
+
```
Justintime50, line 1: Nit: Extra line unneeded.

EasyPost.Tests/EasyPostTest.cs
```diff
@@ -1,62 +1,21 @@
...
+            Client client = new Client(FakeApikey);
             client.ConnectTimeoutMilliseconds = 5000;
             client.RequestTimeoutMilliseconds = 5000;
 
-            Assert.AreEqual(5000, client.ConnectTimeoutMilliseconds);
-            Assert.AreEqual(5000, client.RequestTimeoutMilliseconds);
-        }
-
-        [TestMethod]
-        public async Task TestClientManagerGetCurrent()
```
Justintime50, line 51: Question/Issue: Why were all of these tests removed? We should continue allowing users to override the base URL. This is important for various reasons - our internal teams override the base URL during E2E tests to hit their Homeowner instances when doing internal tests, and some users may way to hit a mock server for their own testing. If we have removed the ability for users to override this value, we need to adjust the client object to continue accepting this parameter.

EasyPost.Tests/EndShipperTest.cs
```diff
@@ -1,93 +1,87 @@
...
-    [TestClass]
-    public class EndShipperTest
+    public class EndShipperTest : UnitTest
     {
-        private TestUtils.VCR _vcr;
+        public EndShipperTest() : base("end_shipper", TestUtils.ApiKey.Production)
+        {
+        }
 
-        // This (and all Beta features) cannot use VCR because of a known conflict between using a VCR/custom HttpClient and re-constructing RestSharp.
```
Justintime50, line 13: Question: Can you explain what this was? It's been awhile and I've forgotten. What allows us to be able to do this now?

EasyPost.Tests/EnumTest.cs
```diff
@@ -1,12 +1,14 @@
...
+using Xunit;
 
 namespace EasyPost.Tests
 {
-    [TestClass]
     public class EnumTest
     {
-        [TestMethod]
+        [Fact]
+        [SuppressMessage("Assertions", "xUnit2017:Do not use Contains() to check if a value exists in a collection")]
```
Justintime50, line 11: Issue: Don't suppress errors and sweep them under the rug when there are reliable ways to fix the underlying problem. 

Something like this maybe? https://www.pmichaels.net/tag/do-not-use-enumerable-any-to-check-if-a-value-exists-in-a-collection/

EasyPost.Tests/ErrorTest.cs
```diff
@@ -1,51 +1,9 @@
-﻿using System;
```
Justintime50, line 1: Issue: Put this whole file back (multiple tests). Not testing errors is not acceptable.

I believe the error itself changed recently due to some changes made in the API. Feel free to look at some of the other libs for an example as they got their error tests updated when we switched fixtures.

EasyPost.Tests/OrderTest.cs
```diff
@@ -1,116 +1,121 @@
...
+        [CrudOperations.Read]
         public async Task TestGetRates()
         {
-            _vcr.SetUpTest("get_rates");
+            UseVCR("get_rates");
 
             Order order = await CreateBasicOrder();
 
-            await order.GetRates();
+            await order.GetRates(); // this does not return anything, does actually update in-place
```
Justintime50, line 39: Question/Suggestion: Was the comment a personal reminder or meant to be kept in code? I'd suggest that the lack of a return value here signifies it returns nothing and therefore isn't a necessary comment.

EasyPost.Tests/OrderTest.cs
```diff
@@ -1,116 +1,121 @@
...
             {
-                Assert.IsInstanceOfType(shipment, typeof(Shipment));
-                Assert.IsNotNull(shipment.postage_label);
+                Assert.IsType<Rate>(rate);
             }
         }
 
-        [TestMethod]
+        [Fact]
+        [CrudOperations.Read] // not really?
```
Justintime50, line 51: Question/Issue: What is the `not really` comment for? Comments that don't provide any context shouldn't be kept around. Can we either clarify the comment with something meaningful or remove it completely?

EasyPost.Tests/OrderTest.cs
```diff
@@ -1,116 +1,121 @@
...
+
+        [Fact]
+        [CrudOperations.Read]
+        public async Task TestRetrieve()
+        {
+            UseVCR("retrieve");
+
+            Order order = await CreateBasicOrder();
+
+
```
Justintime50, line 90: Nit: Only need one newline here

EasyPost.Tests/PickupTest.cs
```diff
@@ -1,110 +1,117 @@
...
-            Assert.IsNotNull(pickup.pickup_rates);
+            Assert.IsType<Pickup>(pickup);
+            Assert.StartsWith("pickup_", pickup.id);
+            Assert.NotNull(pickup.pickup_rates);
         }
 
-        [TestMethod]
-        public async Task TestRetrieve()
+        [Fact]
+        [CrudOperations.Read] // not really?
```
Justintime50, line 32: Question/Issue: What is the not really comment for? Comments that don't provide any context shouldn't be kept around. Can we either clarify the comment with something meaningful or remove it completely?

EasyPost.Tests/Sample.cs
```diff
@@ -0,0 +1,13 @@
+using System.Threading.Tasks;
```
Justintime50, line 1: Question: What is this file's purpose?

EasyPost.Tests/UnitTest.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        /// <summary>
+        ///     An asynchronous function executed after every test.
+        ///     Function should have a string (ID) passed in, and should return a boolean.
+        /// </summary>
+        protected Func<string, Task<bool>>? CleanupFunction { get; set; }
+
+#pragma warning disable CS8618
+        protected UnitTest(string groupName, TestUtils.ApiKey apiKey = TestUtils.ApiKey.Test) => _vcr = new TestUtils.VCR(groupName, apiKey);
+#pragma warning restore CS8618
```
Justintime50, line 26: Question/Issue: why not fix the actual warning issue instead of disabling it?

Something like https://stackoverflow.com/a/60702710/6064135?

EasyPost.Tests/UnitTest.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public void Dispose()
```
Justintime50, line 28: Issue: `Dispose` what? Function names should be `VerbNoun` for clarity.

This should probably also have a docstring attached.

EasyPost.Tests/UserTest.cs
```diff
@@ -105,70 +128,47 @@ public async Task TestUpdate()
...
 
             User user = await CreateUser();
 
-            bool success = await user.Delete();
-            Assert.IsTrue(success);
+            await user.Delete();
 
-            _userId = null; // skip deletion cleanup
-        }
+            // TODO: Assert something
```
Justintime50, line 152: Issue: Let's assert something now for the same reasons as the other delete functions.

Justintime50 (changes requested):
In an effort to keep the back and forth on this massive review easier to maintain, I'm going to review this in sections. I've started with just the tests on this pass. Please address this set of feedback completely. Once that is done, I will move on to another section of code and we'll continue until the entire PR has been reviewed and feedback addressed. This should help both you and I manage it in smaller sections.

EasyPost.Tests/EndShipperTest.cs
nwithan8: Because of the singleton client manager, we could not re-initialize the underlying RestSharp client, a step that was necessary to inject the EasyVCR HTTP client middleware. Because of the re-architecture, this is no longer the case.

EasyPost.Tests/UnitTest.cs
nwithan8: This is an inherited method from the interface, so we can't rename it. I will add a docstring.

EasyPost.Tests/Sample.cs
nwithan8: Just a scratch file to test and show off the redesign.

EasyPost.Tests/ErrorTest.cs
nwithan8: Error unit tests will be revamped when exception types and handling is refactored later in the rewrite process.

EasyPost.Tests/AddressTest.cs
```diff
@@ -1,147 +1,140 @@
...
 
-            AddressCollection addressCollection = await Address.All(new Dictionary<string, object>
+            AddressCollection addressCollection = await Client.Address.All(new Dictionary<string, object>
             {
                 {
                     "page_size", Fixture.PageSize
                 }
             });
-
+            Assert.True(addressCollection.has_more);
```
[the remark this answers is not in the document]
nwithan8: Resolved

EasyPost.Tests/BatchTest.cs
```diff
@@ -40,154 +43,159 @@ private static async Task<Batch> CreateOneCallBuyBatch()
...
 
-            await batch.GenerateScanForm();
+            batch = await batch.Buy();
+
+            if (IsRecording())
+            {
+                Thread.Sleep(30000); // Wait enough time to process
+            }
+
+            batch = await batch.GenerateScanForm();
```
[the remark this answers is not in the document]
nwithan8: Not in my experience lately.

EasyPost.Tests/CarrierAccountTest.cs
nwithan8: Resolved

EasyPost.Tests/CarrierAccountTest.cs
nwithan8: There's nothing to assert. `Delete` returns void. The unit test is inherently asserting that no error was thrown on the API call by not failing.

EasyPost.Tests.VB/EasyPost.Tests.VB.vbproj
nwithan8: It wasn't? The same version is used on master
https://github.com/EasyPost/easypost-csharp/blob/69f918c4c00a4929ac578ea835fdb8e24425561e/EasyPost.Tests/EasyPost.Tests.csproj#L22

Justintime50:
I'm also not seeing any tests that check if multi-threading does indeed work (safely). We'll need those added too.

EasyPost.Tests/OrderTest.cs
nwithan8: More of a personal reminder to myself, I guess, to remind myself about the unintuitive way we do in-place object updates, especially for a `getX` method.

EasyPost.Tests/UserTest.cs
nwithan8: Nothing to assert, since the function does not return anything.

EasyPost.Tests/UnitTest.cs
nwithan8: Because the error is presumptious. It thinks that the `Client` object is not being declared, and suggesting that it be nullable. If that were the case, it would require slapping nullability checks into every single unit test for something that I know is not nullable (`Client` is set shortly after initialization of the unit test and is guaranteed to be set by the time the unit test executes). It's a case of me being more aware of what the code is doing than the IDE is.

EasyPost.Tests/EasyPostTest.cs
nwithan8: I was unaware that was a requirement of our library. I will introduce that capability.

nwithan8:
> I'm also not seeing any tests that check if multi-threading does indeed work (safely). We'll need those added too.

Thoughts on how to prove that? Prove that the API key for a given client is what is was originally even after a new client (with a different key) is made?

EasyPost.Tests/BatchTest.cs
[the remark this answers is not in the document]
Justintime50: If for whatever reason someone needs to come bump this because our queue is backed up, they can make the one-off tweak to a higher time. We have 6 other libraries that have never needed longer than 10 seconds over various re-recordings. Having the test suite wait an entire minute (for the two 30 second waits added here) when it may not always be necessary is too long and will only frustrate users (aka: us). Can we please lower it to 10 seconds?

EasyPost.Tests/BatchTest.cs
```diff
@@ -40,154 +43,159 @@ private static async Task<Batch> CreateOneCallBuyBatch()
...
+            batch = await batch.Buy();
 
-            await batch.AddShipments(new List<Shipment>
+            if (IsRecording())
             {
-                shipment
-            });
-
-            Assert.AreEqual(1, batch.num_shipments);
+                Thread.Sleep(30000); // Wait enough time to process
```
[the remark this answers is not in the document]
Justintime50: Same comment as above ^

EasyPost.Tests/CarrierAccountTest.cs
Justintime50: Does C# not have an "assert no error" like other libs? If so, we should use that here to signify that that is our explicit test. I'd like to avoid implicit tests where possible.

EasyPost.Tests.VB/EasyPost.Tests.VB.vbproj
Justintime50: This was bumped back after my question.

EasyPost.Tests/BatchTest.cs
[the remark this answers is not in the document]
Justintime50: EDIT: After looking, it seems you reduced these to 10s. Please also mention what you did when responding to feedback. It greatly reduces the back and forth instead of requiring a reviewer to dig into each change individually.

EasyPost.Tests/BatchTest.cs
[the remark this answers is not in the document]
Justintime50: EDIT: After looking, it seems you reduced these to 10s. Please also mention what you did when responding to feedback. It greatly reduces the back and forth instead of requiring a reviewer to dig into each change individually.

EasyPost.Tests/EasyPostTest.cs
Justintime50: Thanks! I believe most of our libs allow for this and thought that's what this test was testing.

Can we please add back in the tests that check if we can override the base URL?

EasyPost.Tests/EndShipperTest.cs
Justintime50: Ah yes. Nice!

EasyPost.Tests/ErrorTest.cs
Justintime50: Although that is true, we shouldn't be ripping out tests. The tests should remain, we should tweak them to match the current API behavior, and then when we refactor the error handling soon we can also revisit these error tests. 

Please put the tests back (and tweak values as necessary) so that we can ensure behavior matches our expectations from PR to PR.

EasyPost.Tests/OrderTest.cs
Justintime50: If it was a personal reminder, can you remove this?

EasyPost.Tests/UnitTest.cs
Justintime50: Perfect explanation, can you add a simple comment detailing the same in code? That way we know why we did the thing later on when we all inevitably forget the reasoning.

EasyPost.Tests/UserTest.cs
Justintime50: Same comment as above ^, does C# not provide a "assert no error"? I prefer we be explicit over implicit.

Justintime50:
> Thoughts on how to prove that? Prove that the API key for a given client is what is was originally even after a new client (with a different key) is made?

Hmmm, fair point. Would it make sense to have an API key getter? Once you create a client you can call the getter on the client object? Could even assist users when debugging problems because they can programmatically determine which client is using which API key. Then in our tests, we can attempt to multithread a few requests with different API keys, assert the getters pull out the right keys and don't mix them up on different threads?

EasyPost.Tests/OrderTest.cs
```diff
@@ -1,116 +1,120 @@
...
             {
-                Assert.IsInstanceOfType(shipment, typeof(Shipment));
-                Assert.IsNotNull(shipment.postage_label);
+                Assert.IsType<Rate>(rate);
             }
         }
 
-        [TestMethod]
+        [Fact]
+        [CrudOperations.Read] // not really a read operation, since the only HTTP calls we're making are create and update methods
```
Justintime50, line 51: Can we describe the **why** behind this. Meaning, why is it set to `[CrudOperations.Read]` when it's not really a read?

EasyPost.Tests/PickupTest.cs
```diff
@@ -1,110 +1,117 @@
...
-            Assert.IsNotNull(pickup.pickup_rates);
+            Assert.IsType<Pickup>(pickup);
+            Assert.StartsWith("pickup_", pickup.id);
+            Assert.NotNull(pickup.pickup_rates);
         }
 
-        [TestMethod]
-        public async Task TestRetrieve()
+        [Fact]
+        [CrudOperations.Read] // not really a read operation, since the only HTTP calls we're making are create and update methods
```
Justintime50, line 32: Can we describe the **why** behind this. Meaning, why is it set to `[CrudOperations.Read]` when it's not really a read?

EasyPost.Tests/BatchTest.cs
```diff
@@ -83,111 +64,137 @@ public async Task TestAll()
...
-
+            UseVCR("create_scan_form");
 
             Batch batch = await CreateOneCallBuyBatch();
 
-            await batch.Buy();
+            if (IsRecording())
+            {
+                Thread.Sleep(10000); // Wait enough time to process
+            }
```
Justintime50, line 125: Issue: As mentioned previously, this sleep should not be here. There is no reason to wait in-between creating a batch and buying one as our API will respond immediately. This only adds an extra unnecessary 10 seconds timeout when re-recording without providing value.

Sleeps are required between buying a batch and the next operation such as creating a label or scanform. Not before.

Justintime50 (commented):
Thanks for addressing the test feedback. There are a few stragglers. I'll move on to the next section, but please continue to address those items while I do.

EasyPost.Tests/CarrierAccountTest.cs
nwithan8: There's some interesting discussion about that in [this thread](https://github.com/xunit/xunit/issues/188). I have an idea about how to explicitly check for no exception.

nwithan8: Resolved.

EasyPost/Http/Security.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// <summary>
+        ///     Get the TLS protocol version to use for the request.
+        /// </summary>
+        /// <returns>SecurityProtocolType TLS version enum.</returns>
+        internal static SecurityProtocolType GetProtocol()
+        {
+#if NET45
+            return SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12;
+#else
+            return (SecurityProtocolType)0x00000C00;
```
Justintime50, line 16: Issue: Nobody on planet Earth knows offhand what `0x00000C00` means... can we add a comment here with the human readable name of it?

EasyPost/Models/API/Address.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace EasyPost.Models.API
+{
+    public class Address : EasyPostObject
+    {
+        #region JSON Properties
+
+        [JsonProperty("carrier_facility")]
+        public string carrier_facility { get; set; }
+
```
Justintime50, line 17: Nit: extra newline isn't needed

EasyPost/Models/API/Address.cs
```diff
@@ -0,0 +1,78 @@
...
+        [JsonProperty("street2")]
+        public string street2 { get; set; }
+        [JsonProperty("verifications")]
+        public Verifications verifications { get; set; }
+        [JsonProperty("verify")]
+        public List<string> verify { get; set; }
+        [JsonProperty("verify_strict")]
+        public List<string> verify_strict { get; set; }
+        [JsonProperty("zip")]
+        public string zip { get; set; }
```
Justintime50, line 55: Thought: I'm torn here. On the one hand sorting things like this is great because it's easy to find things and maintain the list. On the other, it breaks the logical ordering of important things like addresses where it's typically street, state, zip, etc. 

I suppose it's probably not a deal **in-code** and that sort of thing is only important at the presentation layer which we aren't concerned with here. What do you think?

EasyPost/Models/API/Address.cs
```diff
@@ -0,0 +1,78 @@
...
+        [JsonProperty("email")]
+        public string email { get; set; }
+        [JsonProperty("error")]
+        public string error { get; set; }
+        [JsonProperty("federal_tax_id")]
+        public string federal_tax_id { get; set; }
+        [JsonProperty("message")]
+        public string message { get; set; }
+        [JsonProperty("mode")]
+        public new string mode { get; set; }
```
Justintime50, line 33: Question: Does `EasyPostObject` have `mode` and all of that now? If so should we remove it here?

EasyPost/Models/API/AddressCollection.cs
```diff
@@ -1,16 +1,15 @@
...
 {
-    public class AddressCollection : Resource
+    public class AddressCollection : Collection
     {
         #region JSON Properties
 
         [JsonProperty("addresses")]
         public List<Address> addresses { get; set; }
-        [JsonProperty("has_more")]
-        public bool has_more { get; set; }
```
Justintime50, line 13: Question: Removed because I assume it's now in the inherited `Collection`?

EasyPost/Models/API/Batch.cs
```diff
@@ -0,0 +1,161 @@
...
+        #region JSON Properties
+
+        [JsonProperty("error")]
+        public string error { get; set; }
+        [JsonProperty("label_url")]
+        public string label_url { get; set; }
+        [JsonProperty("message")]
+        public string message { get; set; }
+        [JsonProperty("mode")]
+        public new string mode { get; set; }
```
Justintime50, line 22: Question: is mode inherited?

EasyPost/Models/API/Form.cs
```diff
@@ -1,26 +1,20 @@
...
-        public DateTime? created_at { get; set; }
         [JsonProperty("form_type")]
         public string form_type { get; set; }
         [JsonProperty("form_url")]
         public string form_url { get; set; }
-        [JsonProperty("id")]
-        public string id { get; set; }
         [JsonProperty("mode")]
-        public string mode { get; set; }
+        public new string mode { get; set; }
```
Justintime50, line 15: Issue: I believe this is a typo? We don't want to be making a new mode here? Also, isn't mode inherited?

EasyPost/Models/API/PaymentMethod.cs
```diff
@@ -59,23 +31,30 @@ public class PaymentMethodObject : Resource
...
         // both
         [JsonProperty("last4")]
         public string last4 { get; set; }
         // credit_card
         [JsonProperty("name")]
         public string name { get; set; }
         // both
-        [JsonProperty("object")]
-        public string Object { get; set; }
+
```
Justintime50, line 42: Issue: Appears to be some duplication of extra spaces and comments here. Let's clean this up

EasyPost/Models/API/PostageLabel.cs
```diff
@@ -36,9 +33,7 @@ public class PostageLabel : Resource
         [JsonProperty("label_zpl_url")]
         public string label_zpl_url { get; set; }
         [JsonProperty("mode")]
-        public string mode { get; set; }
-        [JsonProperty("updated_at")]
-        public DateTime? updated_at { get; set; }
+        public new string mode { get; set; }
```
Justintime50, line 36: Question: Typo? We don't want a `new` string right?

If so, please do a "find all" as there appears to be a few of these that we don't want to miss.

EasyPost/Models/API/Rate.cs
```diff
@@ -29,16 +26,12 @@ public class Rate : Resource
...
-        public string id { get; set; }
         [JsonProperty("list_currency")]
         public string list_currency { get; set; }
         [JsonProperty("list_rate")]
         public string list_rate { get; set; }
         [JsonProperty("mode")]
-        public string mode { get; set; }
-        [JsonProperty("object")]
-        public string Object { get; set; }
+        public new string mode { get; set; }
```
Justintime50, line 34: Question: typo?

EasyPost/Models/API/Smartrate.cs
```diff
@@ -47,7 +15,7 @@ public class Smartrate : Resource
         [JsonProperty("carrier_account_id")]
         public string carrier_account_id { get; set; }
         [JsonProperty("created_at")]
-        public DateTime created_at { get; set; }
+        public new DateTime created_at { get; set; }
```
Justintime50, line 18: Question: typo?

EasyPost/Models/API/Smartrate.cs
```diff
@@ -58,14 +26,12 @@ public class Smartrate : Resource
...
         public int? est_delivery_days { get; set; }
-        [JsonProperty("id")]
-        public string id { get; set; }
         [JsonProperty("list_currency")]
         public string list_currency { get; set; }
         [JsonProperty("list_rate")]
         public double list_rate { get; set; }
         [JsonProperty("mode")]
-        public string mode { get; set; }
+        public new string mode { get; set; }
```
Justintime50, line 34: Question: typo?

EasyPost/Models/API/Smartrate.cs
```diff
@@ -79,7 +45,7 @@ public class Smartrate : Resource
         [JsonProperty("time_in_transit")]
         public TimeInTransit time_in_transit { get; set; }
         [JsonProperty("updated_at")]
-        public DateTime updated_at { get; set; }
+        public new DateTime updated_at { get; set; }
```
Justintime50, line 48: Question: typo?

EasyPost/Models/Shared/Collection.cs
```diff
@@ -0,0 +1,32 @@
...
+using Newtonsoft.Json;
+
+namespace EasyPost.Models.Shared
+{
+    public class Collection : EasyPostObject
+    {
+        #region JSON Properties
+
+        [JsonProperty("has_more")]
+        public bool has_more { get; set; }
```
Justintime50, line 14: Question: What happens if a collection in the future for whatever reason doesn't contain the `has_more` key, will everything blow up?

EasyPost/Models/Shared/Collection.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        [JsonProperty("filters")]
+        internal Dictionary<string, object>? filters { get; set; }
+
+        #endregion
+
+        protected void UpdateFilters(IEnumerable<EasyPostObject>? elements, string propertyName)
```
Justintime50, line 21: Question: Can you walk me through how this works? This could benefit from a docstring. Can users pass `after_id` and `before_id`? What about `page_size`?

Justintime50 (changes requested):
I've reviewed all the HTTP and Models logic. Few small things that appear to be find and replace patterns gone wrong.

2/4 of the way done. Once these are addressed I'll move on to Services and finally Utils.

EasyPost.Tests/EasyPostTest.cs
nwithan8: Resolved.

EasyPost.Tests/OrderTest.cs
nwithan8: Resolved.

EasyPost.Tests/UnitTest.cs
nwithan8: Resolved.

EasyPost.Tests/OrderTest.cs
nwithan8: Because it seemed like the best category, even though it wasn't perfect. Open to suggestions.

nwithan8:
> > Thoughts on how to prove that? Prove that the API key for a given client is what is was originally even after a new client (with a different key) is made?
> 
> Hmmm, fair point. Would it make sense to have an API key getter? Once you create a client you can call the getter on the client object? Could even assist users when debugging problems because they can programmatically determine which client is using which API key. Then in our tests, we can attempt to multithread a few requests with different API keys, assert the getters pull out the right keys and don't mix them up on different threads?

Sounds like a plan. I had originally designed the Client to be a black box (once you set the API key, you can't retrieve it or change it), but I've since re-designed it since that level of obfuscation is probably more harmful than helpful. I will write a unit test to confirm that API keys remain with the client as expected. Which API keys should we use? The test key and what else, a fake key, the production key?

EasyPost.Tests/BatchTest.cs
nwithan8: Resolved.

EasyPost/EasyPost.csproj
```diff
@@ -55,20 +55,22 @@
...
-    <None Include="..\EasyPost.pub">
-      <Link>EasyPost.pub</Link>
-    </None>
-  </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="[13.0.1, 14.0.0)" />
-    <PackageReference Include="RestSharp" Version="[107.3.0, 108.0.0)" />
-    <PackageReference Include="SecurityCodeScan.VS2019" Version="[5.0.0, 6.0.0)">
+    <PackageReference Include="Newtonsoft.Json" Version="[13.0.1, 14.0.0)"/>
+    <PackageReference Include="RestSharp" Version="[108.0.1, 109.0.0)"/>
```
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/Http/Parameters.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace EasyPost.Http
+{
+    public static class Parameters
+    {
+        internal static Dictionary<string, object> Wrap(this Dictionary<string, object> parameters, params string[] keys)
+        {
+            return keys.Reverse()
```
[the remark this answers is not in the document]
nwithan8: The key you would pass in would be the path that the parameters need to be added in.

For example, if I needed these parameters added at "level1" -> "level2" -> "level3", we need to wrap the parameters in "level3" first, then wrap that in "level2", then wrap that in "level1". So it has to traverse the path in reverse.

Justintime50:
> Sounds like a plan. I had originally designed the Client to be a black box (once you set the API key, you can't retrieve it or change it), but I've since re-designed it since that level of obfuscation is probably more harmful than helpful. I will write a unit test to confirm that API keys remain with the client as expected. Which API keys should we use? The test key and what else, a fake key, the production key?

We should just use multiple fake keys. No need to make real API calls for these particular tests and no need to expose our API keys in assertions either. `123`, `456`, `789` should do, 3 separate threads in a single test and three assertions?

EasyPost/Http/Parameters.cs
```diff
@@ -0,0 +1,36 @@
...
+        internal static Dictionary<string, object> Wrap<T>(this List<T> parameters, params string[] keys)
+        {
+            string firstKey = keys.Reverse().First();
+            Dictionary<string, object> dictionary = new Dictionary<string, object>
+            {
+                {
+                    firstKey, parameters
+                }
+            };
+            return keys.Reverse().Skip(1).Aggregate(dictionary, (current, key) => new Dictionary<string, object>
```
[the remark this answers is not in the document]
nwithan8: We skip the first key because it's already been used at this point to make the lowest-level dictionary already.

EasyPost/Http/Request.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        public readonly string? RootElement;
+
+        private readonly Dictionary<string, object> _parameters;
+        private readonly RestRequest _restRequest;
+
+        public Request(string endpoint, Method method, Dictionary<string, object>? parameters = null, string? rootElement = null, ApiVersion? apiVersion = null)
+        {
+            apiVersion ??= ApiVersion.General;
+            endpoint = $"{apiVersion.Value}/{endpoint}";
```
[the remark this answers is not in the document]
nwithan8: Correct, we can easily pass in whatever API version we want used for the underlying request. It's an issue upstream of how users will access/know that this is hitting the beta endpoint.

EasyPost/Http/Request.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            _parameters = parameters ?? new Dictionary<string, object>();
+
+            RootElement = rootElement;
+        }
+
+        /// <summary>
+        ///     Build the request parameters.
+        /// </summary>
+        internal void Build()
```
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/Http/Security.cs
nwithan8: This is some legacy code, no longer relevant, not even used: https://stackoverflow.com/questions/47269609/system-net-securityprotocoltype-tls12-definition-not-found

Deleted.

EasyPost/Models/API/Address.cs
nwithan8: Resolved.

EasyPost/Models/API/Address.cs
nwithan8: I prefer alphabetically, since it will inevitably get unsorted if we try to maintain a manual order.

EasyPost/Models/API/Address.cs
nwithan8: Good catch. Resolved.

EasyPost/Models/API/PaymentMethod.cs
nwithan8: Resolved

EasyPost/Models/Shared/Collection.cs
nwithan8: No, it will fallback to `false` by default.

EasyPost.Tests/ErrorTest.cs
nwithan8: Resolved.

EasyPost/Models/Shared/Collection.cs
nwithan8: Going to remove this for now, we can address collections in the future.

EasyPost.Tests/OrderTest.cs
Justintime50: Right, but what's the significance of the decorator? Is it for the ordering? If so, a note about "using Read here as the most logical ordering for these functions, even though they are not Read CRUD operations" or something to that effect.

EasyPost/Http/Request.cs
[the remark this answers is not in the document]
Justintime50: Ok, I'll figure this out last. Please keep this comment unresolved so I remember later.

EasyPost/Models/API/Address.cs
Justintime50: Agreed, my brain will hurt but since it's internal code I think alphabetically ordering is the best option here.

EasyPost/Models/Shared/Collection.cs
Justintime50: That's right, le sigh.

EasyPost/_base/EasyPostClient.cs
```diff
@@ -0,0 +1,150 @@
...
+            Request request = new Request(url, method, parameters, rootElement, apiVersion);
+            RestRequest restRequest = PrepareRequest(request);
+
+            // Execute the request
+            RestResponse<T> response = await _restClient.ExecuteAsync<T>(restRequest);
+
+            // Check the response's status code
+            if (!response.IsSuccessful)
+            {
+                // HTTP request threw an error (non-2xx response)
```
Justintime50, line 64: 3xx status codes can be successful, does the `IsSuccessful` account for those in the off chance we ever used them? We should really only realize a failure on `4xx` or `5xx`s.

EasyPost/_base/EasyPostClient.cs
```diff
@@ -0,0 +1,150 @@
...
+        internal async Task<bool> Request(Method method, string url, Dictionary<string, object>? parameters = null, ApiVersion? apiVersion = null)
+        {
+            // Build the request
+            Request request = new Request(url, method, parameters, null, apiVersion);
+            RestRequest restRequest = PrepareRequest(request);
+
+            // Execute the request
+            RestResponse response = await _restClient.ExecuteAsync(restRequest);
+
+            // Return whether the HTTP request produced an error (non-2xx response) or not
```
Justintime50, line 114: Same `3xx` as above ^

EasyPost/_base/WithClient.cs
```diff
@@ -0,0 +1,11 @@
+
```
Justintime50, line 1: Nit: extra line

EasyPost.Tests/OrderTest.cs
nwithan8: Gotcha. Yes, the attributes were introduced to help with auto-ordering to preserve CRUD order. I will clarify with a comment.

nwithan8: Resolved.

EasyPost.Tests/PickupTest.cs
nwithan8: Resolved.

EasyPost/Services/TrackerService.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <param name="parameters">A dictionary of tracking codes and carriers</param>
+        /// <returns>True</returns>
+        [CrudOperations.Create]
+        public async Task CreateList(Dictionary<string, object> parameters)
+        {
+            parameters = parameters.Wrap("trackers");
+            await CreateNoResponse("trackers/create_list", parameters);
+        }
+
+
```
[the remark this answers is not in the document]
nwithan8: Resolved

EasyPost/Services/UserService.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            if (id == null)
+            {
+                return await Get<User>("users");
+            }
+
+            return await Get<User>($"users/{id}");
+        }
+
+
```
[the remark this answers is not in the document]
nwithan8: Resolved

EasyPost/_base/EasyPostClient.cs
nwithan8: No, it only accounts for 2xx: https://docs.microsoft.com/en-us/uwp/api/windows.web.http.httpresponsemessage.issuccessstatuscode?view=winrt-22621#property-value

EasyPost/_base/WithClient.cs
nwithan8: Resolved.

EasyPost/_base/EasyPostClient.cs
nwithan8: Resolved.

EasyPost/_base/EasyPostClient.cs
nwithan8: Resolved.

EasyPost.Tests/AddressTest.cs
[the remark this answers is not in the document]
Justintime50: Fair enough

EasyPost.Tests/AddressTest.cs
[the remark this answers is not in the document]
Justintime50: The `HasMore` key was not changed after all in this PR.

EasyPost.Tests/Sample.cs
Justintime50: 🤷‍♂️ let's remove this. People can copy/paste the README example to test now that we know it works.

EasyPost/_base/EasyPostClient.cs
Justintime50: If so, the comment needs to be updated too

EasyPost/_base/EasyPostClient.cs
Justintime50: If so, the comment needs to be updated too.

EasyPost/Models/API/Beta/EndShipper.cs
```diff
@@ -1,9 +1,14 @@
...
+using EasyPost.Http;
+using EasyPost.Utilities.Annotations;
 using Newtonsoft.Json;
+using RestSharp;
 
-namespace EasyPost.Base
+namespace EasyPost.Models.API.Beta
 {
-    public abstract class Address : Resource
+    public class EndShipper : EasyPostObject
```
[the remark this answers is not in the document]
Justintime50: We will migrate this after.

EasyPost.Tests/UserTest.cs
```diff
@@ -1,102 +1,124 @@
...
+        public async Task TestApiKeys()
+        {
+            UseVCR("api_keys");
+
+            User user = await RetrieveMe();
+
+            // API keys will be censored, so we'll just check for the existence of the `children` element
+            List<User> children = user.children;
+            Assert.NotNull(children);
+        }
```
Justintime50, line 86: Issue: This test is bogus and should be removed...

EasyPost.Tests/UserTest.cs
```diff
@@ -1,102 +1,124 @@
...
 
-            User user = await CreateUser();
+            List<ApiKey> apiKeys = await Client.ApiKey.All();
 
-            Assert.IsInstanceOfType(user, typeof(User));
-            Assert.IsTrue(user.id.StartsWith("user_"));
-            Assert.AreEqual("Test User", user.name);
+            // API keys will be censored, so we'll just check for the existence of the list
+            Assert.NotNull(apiKeys);
         }
```
Justintime50, line 73: The assertions here should really be:

```csharp
            Assert.NotNull(apiKeys.keys);
            Assert.NotNull(apiKeys.children);
```

EasyPost.Tests/UserTest.cs
```diff
@@ -1,102 +1,124 @@
...
+            Assert.StartsWith("brd_", brand.id);
+            Assert.Equal(color, brand.color);
         }
 
-        // This endpoint returns the child user keys in plain text, do not run this test.
-        [TestMethod]
-        public async Task TestCreate()
+        [Fact]
+        [CrudOperations.Read]
+        public async Task TestAllApiKeys()
```
Justintime50, line 65: Issue: Please move this test into its own test file since it's on an ApiKey object technically.

EasyPost/Services/ApiKeyService.cs
```diff
@@ -0,0 +1,29 @@
...
+        #region CRUD Operations
+
+        /// <summary>
+        ///     Get a list of all API keys.
+        /// </summary>
+        /// <returns>A list of EasyPost.ApiKey instances.</returns>
+        [CrudOperations.Read]
+        public async Task<List<ApiKey>> All()
+        {
+            return await List<List<ApiKey>>("api_keys", null, "keys");
```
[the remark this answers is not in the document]
Justintime50: Yes, we need a collection here (follow our docs) because there is a `keys` and `children` key here.

EasyPost/Services/EndShipperService.cs
```diff
@@ -0,0 +1,74 @@
...
+        ///     * {"verifications", List&lt;string&gt;} Possible items are "delivery" and "zip4".
+        ///     * {"strict_verifications", List&lt;string&gt;} Possible items are "delivery" and "zip4".
+        ///     All invalid keys will be ignored.
+        /// </param>
+        /// <returns>EasyPost.EndShipper instance.</returns>
+        [CrudOperations.Create]
+        public async Task<EndShipper> Create(Dictionary<string, object> parameters)
+        {
+            parameters = parameters.Wrap("address");
+            return await Create<EndShipper>("end_shippers", parameters, apiVersion: ApiVersion.Beta);
```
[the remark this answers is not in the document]
Justintime50: We'll come back to migrating in another PR

EasyPost/Services/EndShipperService.cs
```diff
@@ -0,0 +1,74 @@
...
+        ///     after
+        ///     this id.
+        ///     * {"start_datetime", string} ISO 8601 datetime string. Only retrieve EndShippers created after this datetime.
+        ///     * {"end_datetime", string} ISO 8601 datetime string. Only retrieve EndShippers created before this datetime.
+        ///     * {"page_size", int} Max size of list. Default to 20.
+        ///     All invalid keys will be ignored.
+        /// </param>
+        /// <returns>A list of EasyPost.EndShipper instances.</returns>
+        [CrudOperations.Read]
+        public async Task<List<EndShipper>> All(Dictionary<string, object>? parameters = null) => await List<List<EndShipper>>("end_shippers", parameters, apiVersion: ApiVersion.Beta);
```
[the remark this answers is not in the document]
Justintime50: We'll come back to migrating in another PR

EasyPost/Services/EndShipperService.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        ///     Retrieve an EndShipper from its id.
+        /// </summary>
+        /// <param name="id">String representing an EndShipper. Starts with "es_".</param>
+        /// <returns>EasyPost.EndShipper instance.</returns>
+        [CrudOperations.Read]
+        public async Task<EndShipper> Retrieve(string id) => await Get<EndShipper>($"end_shippers/{id}", apiVersion: ApiVersion.Beta);
```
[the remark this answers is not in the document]
Justintime50: We'll come back to migrating in another PR

EasyPost/Models/API/Shipment.cs
```diff
@@ -0,0 +1,274 @@
...
+            Shipment shipment = await Request<Shipment>(Method.Post, $"shipments/{id}/buy", parameters);
+
+            insurance = shipment.insurance;
+            postage_label = shipment.postage_label;
+            tracking_code = shipment.tracking_code;
+            tracker = shipment.tracker;
+            selected_rate = shipment.selected_rate;
+            forms = shipment.forms;
+            messages = shipment.messages;
+            fees = shipment.fees;
```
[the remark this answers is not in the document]
Justintime50: Great callout! This is most likely legacy and could probably benefit from using the `Merge` function like everything else. Let's see if we can use `Merge` and re-record cassettes and see if stuff breaks or not. It'll be a much more stable approach to use Merge.

EasyPost/Models/API/WebhookList.cs
```diff
@@ -1,9 +1,10 @@
 ﻿using System.Collections.Generic;
+using EasyPost._base;
 using Newtonsoft.Json;
 
-namespace EasyPost
+namespace EasyPost.Models.API
 {
-    internal class WebhookList : Resource
+    internal class WebhookList : EasyPostObject
```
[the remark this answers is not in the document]
Justintime50: I think `List` is ok here because it's not a true `Collection` that is paginated but a loose lists of objects.

EasyPost/Services/WebhookService.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        /// <summary>
+        ///     Get a list of scan forms.
+        /// </summary>
+        /// <returns>List of EasyPost.Webhook instances.</returns>
+        [CrudOperations.Read]
+        public async Task<List<Webhook>> All(Dictionary<string, object>? parameters = null)
+        {
+            return await List<List<Webhook>>("webhooks", parameters, "webhooks");
```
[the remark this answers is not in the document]
Justintime50: Per our convo, this is the way it is because the API returns an object that has a single key `webhooks` which is then the list of webhook objects.

EasyPost/Models/API/Order.cs
```diff
@@ -0,0 +1,130 @@
...
+        [CrudOperations.Update]
+        public async Task GetRates()
+        {
+            if (id == null)
+            {
+                throw new Exception("id is null");
+            }
+
+            Order order = await Request<Order>(Method.Get, $"orders/{id}/rates");
+            rates = order.rates;
```
jchen293, line 107: We should `Merge` function here to update the current object

Justintime50 (changes requested):
There are a few hanging items (in addition to Jack's feedback) that needs to be addressed before I can sign off on this, but it's pretty close.

1. Let's pin the versions in the `nuspec` file as well if possible
2. The README example will need to be updated to match the new syntax
3. There appears to still be 2 occurrences of `public new` that need fixing
4. It appears that all of the `Merge` functionality has been removed... we should put it back to keep user expectations. We will revisit removing/refactoring that in the next major release when we tackle the data flow in the libs.
5. Various feedback on the ApiKey class and tests.

EasyPost/Models/API/Shipment.cs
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/Services/ApiKeyService.cs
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/_base/EasyPostClient.cs
nwithan8: Resolved.

EasyPost/_base/EasyPostClient.cs
nwithan8: Resolved.

EasyPost.Tests/Sample.cs
nwithan8: Resolved.

EasyPost/Models/API/Shipment.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// <summary>
+        ///     Generate a postage label for this shipment.
+        /// </summary>
+        /// <param name="fileFormat">Format to generate the label in. Valid formats: "pdf", "zpl" and "epl2".</param>
+        [CrudOperations.Update]
+        public async Task<Shipment> GenerateLabel(string fileFormat)
+        {
+            if (id == null)
+            {
+                throw new Exception("id is null");
```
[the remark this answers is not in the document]
nwithan8: This will be addressed during the later PR that revamps error handling

EasyPost/Utilities/Http.cs
```diff
@@ -0,0 +1,33 @@
...
+        ///     Return whether the given response has a successful status code.
+        /// </summary>
+        /// <param name="response">Response to check.</param>
+        /// <returns>True if the response code is in the 200-399 range, false otherwise.</returns>
+        internal static bool ReturnedNoError(this RestResponse response)
+        {
+            return StatusCodeBetween(response, 100, 399);
+        }
+
+        private static bool StatusCodeBetween(RestResponseBase response, int min, int max)
```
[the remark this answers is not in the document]
nwithan8: Resolved.

EasyPost/_base/EasyPostObject.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    return null;
+                }
+
+                string? prefix = id.Split('_').First() ?? null;
+                return prefix;
+            }
+        }
+
+        public override bool Equals(object? obj)
```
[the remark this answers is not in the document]
nwithan8: These are self-explanatory overrides of built-in functions.

EasyPost.Tests/UserTest.cs
nwithan8: Resolved.

EasyPost.Tests/UserTest.cs
nwithan8: Resolved.

EasyPost.Tests/UserTest.cs
nwithan8: Resolved.

EasyPost/Models/API/Order.cs
nwithan8: Resolved.

nwithan8:
Feedback has been addressed. Just waiting on decision about the Beta visibility.

Justintime50:
@nwithan8 were you able to attempt the namespacing concept for beta?

nwithan8:
> @nwithan8 were you able to attempt the namespacing concept for beta?

Yes, I have implemented it. I have a concern that I've commented (the best I could) in `Client.cs`. I'd like to discuss how future beta features might look to determine if my concerns are founded or not.

nwithan8:
Beta access has been reworked again. This implementation is less automagical, but does give us as maintainers more fine-grained control. Came about as I was able to disconnect the idea of "beta group of features" and "beta API prefix".

Users access the "beta group of features" via `myClient.Beta`.

Meanwhile, each and every API call in the library (regardless of what group they belong to) can, on an individual basis, target a specific API version (defaults to general availability).

This will allow us finer control over what endpoint each function hits (albeit, slightly more maintenance), and can scale to accommodate future other API versions (imagine, Address.Create always hits `v2`, but Address.Retrieve hits a new `v3`, and there's a beta feature called Address.Whatever that hits `beta`)

Justintime50 (commented):
My only thought here is instead of calling it `ApiVersion.General` we call it `ApiVersion.Current` to be a bit more clear?

nwithan8:
> My only thought here is instead of calling it `ApiVersion.General` we call it `ApiVersion.Current` to be a bit more clear?

Latest, Current, GeneralAvailability, whatever you want.

Justintime50 (approved):
I'm sure I'll regret this, but 🚀 LGTM!

Please squash the commits on merge.

nwithan8:
> I'm sure I'll regret this, but 🚀 LGTM!

That's the spirit 😄

EasyPost/Services/ShipmentService.cs
```diff
@@ -0,0 +1,129 @@
...
+        ///     * {"page_size", int} Size of page. Default to 20.
+        ///     * {"purchased", bool} If true only display purchased shipments.
+        ///     All invalid keys will be ignored.
+        /// </param>
+        /// <returns>An EasyPost.ShipmentCollection instance.</returns>
+        [CrudOperations.Read]
+        public async Task<ShipmentCollection> All(Dictionary<string, object>? parameters = null)
+        {
+            ShipmentCollection shipmentCollection = await List<ShipmentCollection>("shipments", parameters);
+            shipmentCollection.Client = Client;
```
jchen293, line 100: Question: what does this line do? I have seen this in several places for `All` function

EasyPost/Services/WebhookService.cs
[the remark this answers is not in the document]
jchen293: @nwithan8 do we need need the `WebhookList` class since it's not being used in anywhere?

EasyPost/Models/API/Options.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
+using EasyPost._base;
 using Newtonsoft.Json;
 
-namespace EasyPost
+namespace EasyPost.Models.API
 {
-    public class Options : Resource
+    public class Options : EasyPostObject
```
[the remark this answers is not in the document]
nwithan8: There are a handful of changes that were made while this was in-flight, which we will address after this PR.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+- Library better supports different API versions (i.e. `v2`, `beta`).
+- All objects now share a common base set of properties, including `id`, `created_at`, `updated_at`, and `mode`.
+- Under the hood improvements:
+  - Underlying `Request`-`Client`-`ClientConfiguration` relationship has been re-architected to allow for thread safety.
+  - Process of generating an API request has been standardized and simplified.
+  - Improved accessibility levels of internal functions, to prevent accidental use by end users.
+  - Files have been organized into a more logical structure.
+  - Methods and properties have been organized (e.g. methods ordered by CRUD, properties ordered alphabetically).
+- Dependencies updated to latest versions, including `RestSharp` v108
+- `EndShipper` moved from beta to general availability.
```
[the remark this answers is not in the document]
nwithan8: There are a handful of changes that were made while this was in-flight, which we will address after this PR.

EasyPost/Services/WebhookService.cs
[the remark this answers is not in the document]
nwithan8: Good catch, I got myself tripped up.

EasyPost/Services/AddressService.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <returns>EasyPost.Address instance.</returns>
+        [CrudOperations.Read]
+        public async Task<Address> Retrieve(string id)
+        {
+            return await Get<Address>($"addresses/{id}");
+        }
+
+        #endregion
+
+        private async Task<Address> SendCreate(string endpoint, Dictionary<string, object> parameters, string? rootElement = null)
```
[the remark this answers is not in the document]
nwithan8: Yes. It was a holdover from old code, refactored. Thanks, resolved.

EasyPost/Services/ShipmentService.cs
nwithan8: Line 99 makes the API call and deserializes the JSON data into a `ShipmentCollection` object (a wrapper that stores a list of shipments as well as information on `has_more`, etc.)

Line 100 copies the API client that was used to make this call into the collection, so the collection object itself can make API calls via instance methods.

EasyPost/_base/EasyPostObject.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <param name="source">Object instance to extract properties from to merge into this object instance.</param>
+        internal void Merge(object source)
+        {
+            foreach (PropertyInfo property in source.GetType().GetProperties())
+            {
+                property.SetValue(this, property.GetValue(source, null), null);
+            }
+        }
+
+        protected async Task Update<T>(Method method, string url, Dictionary<string, object>? parameters = null, string? rootElement = null, ApiVersion? overrideApiVersion = null) where T : class
```
jchen293, line 82: Missing docstring

README.md
```diff
@@ -30,41 +30,91 @@ namespace example
...
-            Dictionary<string, object> fromAddress = new Dictionary<string, object>() {
-                { "name", "Dr. Steve Brule" },
-                { "street1", "417 Montgomery Street" },
-                { "street2", "5th Floor" },
-                { "city", "San Francisco" },
-                { "state", "CA" },
-                { "country", "US" },
-                { "zip", "94104" },
-                { "phone", "4153334444" }
+            Client client = new Client("EASYPOST_API_KEY");
```
jchen293, line 33:
```suggestion
            Client client = new Client(Environment.GetEnvironmentVariable("EASYPOST_API_KEY"));
```
nit

README.md
```diff
@@ -77,29 +127,19 @@ namespace example
...
+A `Client` object is the entry point into the EasyPost API. It is instantiated with your API key:
 
-If you are operating with multiple EasyPost API keys, or wish to delegate the construction of the client requests, configure the `ClientManager` with a delegate at application initialization.
-
-```cs
+```csharp
 using EasyPost;
 
-ClientManager.SetCurrent(() => new Client(new ClientConfiguration("yourApiKeyHere")));
+Client myClient = new Client("EASYPOST_API_KEY");
```
jchen293, line 135:
```suggestion
Client myClient = new Client(Environment.GetEnvironmentVariable("EASYPOST_API_KEY"));
```

jchen293 (approved):
lgtm, nice job!

EasyPost/_base/EasyPostObject.cs
nwithan8: This has been fixed in a new PR

PR #312, opened 2022-08-17
Add validate webhook secret function

jchen293:
# Description

Adds validate Webhook function that returns your webhook or raises an error if there is a webhook_secret mismatch

# Testing

Adds a test to ensure you get the webhookBody back on a validated webhook
Adds two tests to ensure that errors are thrown when there is an invalid webhook

# Pull Request Type


EasyPost/Utilities/Cryptography.cs
```diff
@@ -0,0 +1,152 @@
...
+            if (normalizationForm != null)
+            {
+                secret = secret.Normalize(normalizationForm.Value);
+            }
+
+            byte[] keyBytes = Encoding.UTF8.GetBytes(secret);
+
+            using HMACSHA256 hmac = new HMACSHA256(keyBytes);
+            byte[] hash = hmac.ComputeHash(data);
+            return hash.AsHexString();
```
Justintime50, line 89: Nit: return statement could use a line break before it.

EasyPost/Utilities/Cryptography.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+
```
Justintime50, line 92: Nit: Extra line here unneeded.

EasyPost/Webhook.cs
```diff
@@ -98,5 +100,29 @@ public static async Task<Webhook> Retrieve(string id)
...
+            if (headers.ContainsKey("X-Hmac-Signature"))
+            {
+                providedSignature = headers["X-Hmac-Signature"]?.ToString();
+            }
+            else
+            {
+                throw new Exception("Webhook received does not contain an HMAC signature.");
+            }
+
+            string computedHexDigest = data.CalculateHMACSHA256HexDigest(secret, NormalizationForm.FormKD); // normalize with NFKD profile
```
Justintime50, line 116: Issue: Comments that say what a thing does when the thing has the same info as the comment don't provide value and shouldn't be included as they only add noise to the code.

```suggestion
            string computedHexDigest = data.CalculateHMACSHA256HexDigest(secret, NormalizationForm.FormKD);
```

EasyPost/Utilities/Cryptography.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <summary>
+        ///     Check whether two signatures match. This is safe against timing attacks.
```
Justintime50, line 94: Question: How are these safe against timing attacks? You are still just comparing strings which isn't typically safe.

We should probably implement something like [SlowEquals](https://gist.github.com/adrianstevens/b053ec17cfc5ca868e71) to assist with this.

EasyPost/Utilities/Cryptography.cs
```diff
@@ -0,0 +1,152 @@
...
+                {
+                    err = true;
+                }
+            }
+
+            return !err;
+        }
+
+        /// <summary>
+        ///     Check whether two signatures match. This is safe against timing attacks.
```
Justintime50, line 120: Question: How are these safe against timing attacks? You are still just comparing strings which isn't typically safe.

We should probably implement something like [SlowEquals](https://gist.github.com/adrianstevens/b053ec17cfc5ca868e71) to assist with this.

Justintime50 (dismissed):

EasyPost.Tests/WebhookTest.cs
```diff
@@ -97,6 +98,70 @@ public async Task TestRetrieve()
...
+            // ReSharper disable once StringLiteralTypo
+            const string webhookSecret = "sécret";
+            Dictionary<string, object?> headers = new Dictionary<string, object?>
+            {
+                {
+                    "X-Hmac-Signature", "hmac-sha256-hex=e93977c8ccb20363d51a62b3fe1fc402b7829be1152da9e88cf9e8d07115a46b"
+                }
+            };
+
+            Event @event = Webhook.ValidateWebhook(data, headers, webhookSecret);
```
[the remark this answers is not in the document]
nwithan8: `event` is a reserved word in C#. A [common tactic](https://stackoverflow.com/questions/254669/what-does-placing-a-in-front-of-a-c-sharp-variable-name-do) to get around this is to append an @ at the front of the variable.

EasyPost/EasyPost.csproj
```diff
@@ -58,6 +57,7 @@
     <None Include="..\EasyPost.pub">
       <Link>EasyPost.pub</Link>
     </None>
+    <None Remove="FodyWeavers.xsd" />
```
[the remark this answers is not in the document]
nwithan8: Old data, removed

EasyPost/Utilities/Cryptography.cs
nwithan8: It's safe against timing attacks because it does not return early when a mismatch is detected. It waits to go through the entire array before returning a boolean, making it impossible to tell where in the array the mismatch occurred.

Justintime50: Can you provide some documentation on the matter? How do we know this?

EasyPost.Tests/Fixture.cs
```diff
@@ -706,5 +709,23 @@ public static string Event
...
+        public static byte[] EventBody
+        {
+            get
+            {
+                const string relativePath = "eventBody.json";
+                string fullPath = Path.Combine(TestUtils.GetSourceFileDirectory(), relativePath);
+
+                try
+                {
+                    string jsonString = File.ReadAllText(fullPath).Replace("\n", "").Replace("\r", ""); // remove potential newline at end of JSON file, it works, don't touch it :)
```
Justintime50, line 721: Issue: Just read the first line of the file, it's much less brittle to do that than this double replace. Go and Java read the first line, you can use those as an example.

EasyPost/Utilities/Cryptography.cs
nwithan8: https://codahale.com/a-lesson-in-timing-attacks/

EasyPost/Utilities/Cryptography.cs
Justintime50: Same question as above.

EasyPost/Utilities/Cryptography.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// </summary>
+        /// <param name="signature1">First signature.</param>
+        /// <param name="signature2">Second signature.</param>
+        /// <returns>Whether the two signatures match.</returns>
+        public static bool SignaturesMatch(string signature1, string signature2)
+        {
+            signature1 = signature1.ToLower();
+            signature2 = signature2.ToLower();
```
Justintime50, line 128: Issue: I don't foresee a need for us to be lowercasing here. That lowers the integrity of the check. Let's remove this completely and see if the tests pass if they do, if so we don't need this.

EasyPost/Utilities/Cryptography.cs
Justintime50: I was looking at some weird cached results for this page and the initial commit. It appears my concerns have since been addressed.

EasyPost/Utilities/Cryptography.cs
Justintime50: I was looking at some weird cached results for this page and the initial commit. It appears my concerns have since been addressed.

EasyPost.Tests/Fixture.cs
nwithan8: Not an issue, but addressed.

nwithan8 (approved):

EasyPost.Tests/Fixture.cs
```diff
@@ -1,5 +1,8 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
```
Justintime50, line 3: Please undo all the auto formatting...

Justintime50 (dismissed):

EasyPost.Tests/Fixture.cs
Justintime50: An issue is something that must be fixed before the reviewer will approve. Please re-review the conventional comments internal doc.

EasyPost.Tests/Fixture.cs
```diff
@@ -706,5 +709,23 @@ public static string Event
                 };
             }
         }
+        public static byte[] EventBody
```
Justintime50, line 712: Nit: should have a line break before the function

Justintime50 (approved):

Justintime50:
@nwithan8 this should have had @jchen293 take another look since you both contributed code here

nwithan8:
> @nwithan8 this should have had @jchen293 take another look since you both contributed code here

Couldn't tag him as a reviewer since he was the PR author.

Justintime50:
You can just @ him 😉

jchen293:
Code LGTM as I also worked on this, the only question I have is can there be just one `SignaturesMatch` function since one of them is not being used elsewhere?

PR #323, opened 2022-09-13
Improvements to error handling

nwithan8:
# Description

- Introduce a handful of explicit error types
- Base EasyPostError exception class
- Base ApiError (extends EasyPostError), several sub-classes for API/HTTP related exceptions
- Several other exception classes (extend EasyPostError) - for JSON, filtering, validation, etc.
- API exceptions are generated based on the status code of the failed API call.
  - End users can use `Easypost.Exceptions.Constants.GetEasyPostExceptionType(myStatusCodeInteger)` to see what type of exception will be thrown for a given status code
- Exception strings/templates are constants
  - Consistent wording across all exceptions
  - End users can retrieve the string/template for each exception, in case they want to parse the text for whatever reason
- Improvements to custom switch-case and enum mechanics
  - Switch-case can handle first match or all match
  - All enums in the library now use the custom enum class for consistency and maintainability (no change needed syntactically for the end-user)

# Testing

# Pull Request Type


EasyPost.Tests/ErrorTest.cs
```diff
@@ -29,10 +36,107 @@ public async Task TestBadParameters()
...
+
+                // Generate a dummy RestResponse with the given status code to parse
+                HttpStatusCode statusCode = (HttpStatusCode)Enum.Parse(typeof(HttpStatusCode), statusCodeInt.ToString());
+                RestResponse response = new RestResponse { StatusCode = statusCode };
+
+                ApiError generatedError = ApiError.FromResponse(response);
+
+                // should be of base type ApiError
+                Assert.Equal(typeof(ApiError), generatedError.GetType().BaseType);
+                // should be specifically of the type we expect
```
Justintime50, line 80: Nit: this sentence reads weird

EasyPost.Tests/ErrorTest.cs
```diff
@@ -29,10 +36,107 @@ public async Task TestBadParameters()
...
         {
+            UseVCR("empty_api_key");
+
             // No longer possible to have an empty API key
+        }
+
+        [Fact]
+        public Task TestKnownApiExceptionGeneration()
+        {
+            UseVCR("known_api_exception_generation");
```
Justintime50, line 49: Question: What's with the `UseVCR` function here? This doesn't even use VCR... Kinda misleading.

EasyPost.Tests/ErrorTest.cs
```diff
@@ -29,10 +36,107 @@ public async Task TestBadParameters()
...
+                { 404, typeof(NotFoundError) },
+                { 405, typeof(MethodNotAllowedError) },
+                { 408, typeof(TimeoutError) },
+                { 422, typeof(InvalidRequestError) },
+                { 429, typeof(RateLimitError) },
+                { 500, typeof(InternalServerError) },
+                { 503, typeof(ServiceUnavailableError) },
+                { 504, typeof(GatewayTimeoutError) }
+            };
+            foreach (KeyValuePair<int, Type> exceptionDetails in exceptionsMap)
```
Justintime50, line 67: Thought: I'd like to be testing each type to ensure they contain the properties/keys we'd expect to pretty please.

EasyPost/Exceptions/API/_Base.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        public readonly string? Code;
+        public readonly List<Error>? Errors;
+        public readonly int? StatusCode;
+
+        /// <summary>
+        ///     Get a formatted error string with expanded details about the error.
+        /// </summary>
+        /// <returns>A formatted error string.</returns>
+        public string PrettyPrintString
```
Justintime50, line 22: Question: Can we just call it `PrettyPrint` so it's simpler and matches other already existing langs?

We also need a unit test for this to ensure the output is correct.

EasyPost/Exceptions/API/_Base.cs
```diff
@@ -0,0 +1,108 @@
...
+        // great minds think alike: https://github.com/stripe/stripe-dotnet/blob/6b9513d3b938d265c7607db919ad2c536ab578c3/src/Stripe.net/Infrastructure/Public/StripeClient.cs#L171
+
+        /// <summary>
+        ///     Parse a RestResponse response object and return an instance of the appropriate exception class.
+        ///     Do not pass a non-error response to this method.
+        /// </summary>
+        /// <param name="response">RestResponse response to parse</param>
+        /// <raises>EasyPostError if a non-400/500 error code is extracted from the response.</raises>
+        /// <returns>An instance of an HttpError-inherited exception.</returns>
+        internal static ApiError FromResponse(RestResponse response)
```
Justintime50, line 62: Thought: If we are so concerned about not using this for success, should we instead rename it to `FromErrorResponse` or something similar to more clearly call out what this is for?

EasyPost/Exceptions/Constants.cs
```diff
@@ -0,0 +1,76 @@
...
+            return exceptionType;
+        }
+
+        public static Type? GetEasyPostExceptionType(HttpStatusCode statusCode)
+        {
+            return GetEasyPostExceptionType((int)statusCode);
+        }
+
+        // public so end-users can access if need to (i.e. regex?)
+        public static class ErrorMessages
```
Justintime50, line 59: Yesss, I love this.

EasyPost/Exceptions/Constants.cs
```diff
@@ -0,0 +1,76 @@
...
+            public const string InvalidApiKeyType = "Invalid API key type.";
+            public const string InvalidParameter = "Invalid parameter: {0}.";
+            public const string InvalidWebhookSignature = "Invalid webhook signature.";
+            public const string JsonDeserializationError = "Error deserializing JSON into object of type {0}.";
+            public const string JsonNoDataToDeserialize = "No data to deserialize.";
+            public const string JsonSerializationError = "Error serializing {0} object into JSON.";
+            public const string MissingProperty = "{0} object is missing attribute: {0}.";
+            public const string MissingRequiredParameter = "Missing required parameter: {0}.";
+            public const string NoObjectFound = "No {0} found.";
+            public const string NoPaymentMethods = "No payment methods are set up. Please add a payment method via the web dashboard.";
```
Justintime50, line 70: Issue: We should probably not say "dashboard" as many users don't have access to the EasyPost dashboard since they are child or referral users and are being whitelisted. Let's make this more generic like "Please add a payment method and try again."

Justintime50 (changes requested):

EasyPost.Tests/ErrorTest.cs
nwithan8: Fair enough, just wanted to remain consistent in case this unit test did make API calls in the future.

EasyPost/Exceptions/API/_Base.cs
nwithan8: Resolved

EasyPost/Exceptions/Constants.cs
nwithan8: Resolved.

EasyPost/Exceptions/API/_Base.cs
nwithan8: Sounds good. I didn't want it to sound like this function will actually print the string (i.e. to console); it only returns the string.

nwithan8: Resolved

EasyPost.Tests/ErrorTest.cs
nwithan8: Resolved

EasyPost/Exceptions/API/_Base.cs
```diff
@@ -0,0 +1,108 @@
+﻿using System;
```
jchen293, line 1: Question: why does this file start with a underscore

EasyPost/Exceptions/General/_ValidationError.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace EasyPost.Exceptions.General
```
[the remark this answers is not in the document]
nwithan8: Just so it's at the top of the file browser since it's a base type that shouldn't be directly accessed.

EasyPost/Exceptions/API/_Base.cs
nwithan8: Just so it's at the top of the file browser since it's a base type that shouldn't be directly accessed.

EasyPost.Tests/ErrorTest.cs
```diff
@@ -118,25 +121,42 @@ public void TestUnknownApiException5xxGeneration()
...
+        {
+            const int statusCode = 401;
+
+            // Generate a dummy RestResponse with the given status code to parse
+            HttpStatusCode httpStatusCode = (HttpStatusCode)Enum.Parse(typeof(HttpStatusCode), statusCode.ToString());
+            RestResponse response = new RestResponse { StatusCode = httpStatusCode };
+
+            ApiError generatedError = ApiError.FromErrorResponse(response);
+
+            // unfortunately, we can't easily load error-related JSON data into the response for parsing, so the pretty print is going to fallback to default values.
```
Justintime50, line 154: Can we not mock the return value of the called error? I don't love not being able to test something.

EasyPost/Exceptions/API/_Base.cs
```diff
@@ -88,7 +88,7 @@ internal static ApiError FromResponse(RestResponse response)
                 // could not extract error details from the API response (or API did not return data, i.e. 500) -> HttpError type rather than ApiError
                 errorMessage = response.ErrorMessage // fallback to standard HTTP error message
                                ?? response.StatusDescription // fallback to standard HTTP status description
-                               ?? "HTTP call did not return a standard error message"; // fallback to unknown error
+                               ?? "API did not return any error data"; // fallback to unknown error
```
Justintime50, line 91: Thought: An argument could be made this should also live in that constants file?

EasyPost.Tests/ErrorTest.cs
nwithan8: In a way, the `TestBadParameters` test above this validates that error parsing works properly, since that makes a real, bad API call and evaluates the exception thrown by it.

EasyPost/Exceptions/API/_Base.cs
nwithan8: Smart thinking, resolved.

Justintime50 (approved):

PR #326, opened 2022-09-14
feat: Include optional EndShipper ID when buying a shipment

nwithan8:
# Description

- Users can optionally provide an EndShipper ID when buying a shipment

# Testing

- Unit test for feature won't work without serious dependency injection and mocking
- Can manually confirm that the parameter was passed properly in the API call

# Pull Request Type

### DIFF
--- a/EasyPost/Shipment.cs
+++ b/EasyPost/Shipment.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading.Tasks;
 using EasyPost.Utilities;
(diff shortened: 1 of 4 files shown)


EasyPost/Shipment.cs
```diff
@@ -118,6 +119,11 @@ public async Task Buy(string rateId, string? insuranceValue = null, bool withCar
                 body["insurance"] = insuranceValue;
             }
 
+            if (endShipperId != null)
+            {
+                body["end_shipper_id"] = endShipperId;
+            }
```
Justintime50, line 125: I'd prefer if we simply exclude test coverage of this if statement here than adding another skipped test since we won't be testing this. See PHP for an example. It's not ideal, but it's the next best thing to call out "we meant not to test this"

Justintime50: Any news on skipping the test coverage here? Since we are already in these weeds?

nwithan8: I have excluded the `Buy` functions from code coverage. Unfortunately, there's not a way to exclude just a parameter from code coverage, I have to mark the whole function as excluded

Justintime50: hmm... we don't want to exclude the whole function because this is arguably the most important function in the whole lib, we need to know it's getting tested and ignoring it outright for ~2 lines makes us worse off. Let's go ahead and take the hit for the uncovered lines, keep the coverage for it, and we'll circle back.

nwithan8: Coverage of that function is at 50% of potential branches
image

jchen293 (approved):

PR #327, opened 2022-09-16
Port missing features

nwithan8:
# Description

- All [features introduced into `master` ](https://github.com/EasyPost/easypost-csharp/pulls?q=is%3Apr+is%3Aclosed+created%3A%3E2022-08-04+is%3Amerged+base%3Amaster)since the rewrite process began have been ported into the rewrite.
- Improvements to client handoff
  - All EasyPostObject objects and sub-objects will be handed the client used to create them (i.e. when creating a Shipment, not only will the Shipment object have the client used to create it, assigned to it, but the sub-objects of that Shipment (ToAddress, FromAddress, Parcel, etc.) will also have the client assigned to them)
- Client equality is now determined by the client configuration (if two clients have the same API key and API base, they are equivalent)
- Beta models/services have been removed from `beta` folders/namespaces
  - Access to these beta features is dictated by their presence in the beta client, and beta endpoint usage is dictated by the API version override parameter in functions.
  - The namespaces/folders, while adding some organization, are ultimately unnecessary, and cause more refactoring needs when moving a feature from beta to GA

# Testing

- All unit tests pass
- New unit tests have been added and recorded as needed
- EndShipper unit tests have been re-recorded for GA (no beta unit tests exist for EndShipper)

# Pull Request Type


Justintime50 (approved):
I double checked the list and this __seems__ to be everything. Please double check yourself to ensure we haven't missed something.

nwithan8:
> I double checked the list and this **seems** to be everything. Please double check yourself to ensure we haven't missed something.

I have manually checked every PR made to `master` in the past two months, and it does seem like we have gotten everything.

PR #329, opened 2022-09-19
feat: Add support for Partner/White Label

nwithan8:
# Description

- Add support for Partner/White Label:
  - Create a new referral user
  - Update a referral user's email address
  - List all referral users
  - Add a credit card to a referral user's account
- Add unit tests for Partner/White Label

# Testing

- New unit tests added and recorded
- Unit test for adding credit card disabled due to VCR issues
  - Can confirm test works E2E

# Pull Request Type

### DIFF
--- a/EasyPost.Tests/ReferralTest.cs
+++ b/EasyPost.Tests/ReferralTest.cs
@@ -78,7 +78,7 @@ namespace EasyPost.Tests
         [TestMethod]
         public async Task TestReferralAddCreditCard()
         {
-            // This test will not record the Stripe API calls, nor the EasyPost add credit card call, since both use different API clients.
+            // This test is skipped because it will not record the Stripe API calls, nor the EasyPost add credit card call, since both use different API clients.
             _vcr.SetUpTest("referral_add_credit_card");
 
             Dictionary<string, object> parameters = Fixture.CreditCardDetails;
--- a/EasyPost.Tests/ReferralTest.cs
+++ b/EasyPost.Tests/ReferralTest.cs
@@ -1,0 +1,99 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace EasyPost.Tests
+{
+    [TestClass]
+    public class ReferralTest
+    {
+        private TestUtils.VCR _vcr;
+
+        private static string ReferralUserKey
+        {
+            get { return TestUtils.GetApiKey(TestUtils.ApiKey.Referral); }
+        }
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            _vcr = new TestUtils.VCR("referral", TestUtils.ApiKey.Partner);
+        }
+
+        private static async Task<ReferralCustomer> CreateReferral()
+        {
+            return await ReferralCustomer.Create(Fixture.ReferralUser);
+        }
+
+        [TestMethod]
+        public async Task TestCreate()
+        {
+            _vcr.SetUpTest("create");
+
+            ReferralCustomer referral = await CreateReferral();
+
+            Assert.IsNotNull(referral);
+            Assert.IsInstanceOfType(referral, typeof(ReferralCustomer));
+            Assert.IsNotNull(referral.id);
+            Assert.IsTrue(referral.id.StartsWith("user_"));
+            Assert.AreEqual("Test Referral", referral.name);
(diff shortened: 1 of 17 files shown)


EasyPost.Tests/ReferralTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                Assert.IsInstanceOfType(referral, typeof(ReferralCustomer));
+            }
+        }
+
+        [Ignore]
+        [TestMethod]
+        public async Task TestReferralAddCreditCard()
+        {
+            // This test will not record the Stripe API calls, nor the EasyPost add credit card call, since both use different API clients.
```
Justintime50, line 81: Nit
```suggestion
            // This test is skipped because it will not record the Stripe API calls, nor the EasyPost add credit card call, since both use different API clients.
```

EasyPost.Tests/TestUtils.cs
```diff
@@ -24,7 +24,8 @@ public class TestUtils
             "keys",
             "phone_number",
             "phone",
-            "test_credentials"
+            "test_credentials",
+            "public_key"
```
Justintime50, line 28: Please leave the `public_key` in the cassettes. The fact that it's public means it's not sensitive and intended for consumption. This could eventually backfire on us if we start scrubbing it.

EasyPost/Billing.cs
```diff
@@ -27,9 +27,9 @@ public static async Task<bool> DeletePaymentMethod(PaymentMethod.Priority priori
         /// <param name="amount">Amount to fund.</param>
         /// <param name="priority">Which type of payment method to use to fund the wallet. Defaults to primary.</param>
         /// <returns>True if successful, false otherwise.</returns>
-        public static async Task<bool> FundWallet(string amount, PaymentMethod.Priority priority = PaymentMethod.Priority.Primary)
+        public static async Task<bool> FundWallet(string amount, PaymentMethod.Priority? priority)
```
Justintime50, line 30: I'd prefer we expose the default to the user and not hide it in the function. Can't it be a nullable default? If not then I say (if we can), we just leave the default here.

EasyPost/ReferralCustomer.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// <returns>An EasyPost.PaymentMethodObject instance.</returns>
+        /// <exception cref="Exception">When the request fails.</exception>
+        public static async Task<PaymentMethodObject> AddCreditCardToUser(string referralApiKey, string number, int expirationMonth, int expirationYear, string cvc, PaymentMethod.Priority? priority = null)
+        {
+            string easypostStripeApiKey = await RetrieveEasypostStripeApiKey();
+            string stripeToken;
+
+            try
+            {
+                stripeToken = await CreateStripeToken(number, expirationMonth, expirationYear, cvc, easypostStripeApiKey);
```
Justintime50, line 92: This variable should have a fallback if we cannot get it from Stripe (in other libs, we set it to an empty string then check for existence), this is important since we don't control this endpoint and they could (although unlikely) change the expectations of this response.

EasyPost/Billing.cs
nwithan8: You can't use a non-static variable as a default value in a method declaration (one downside to using a custom Enum class rather than the built-in one)

EasyPost/ReferralCustomer.cs
nwithan8: That check is done at the end of the `CreateStripeToken` function.

EasyPost/Billing.cs
nwithan8: Actually, I was mistaken about the reason.
You can't have default values of anything other than `null` and strings in C#:
image

nwithan8: The docstring for this method indicates that "primary" is the default value if one is not provided.

EasyPost.Tests/TestUtils.cs
nwithan8: Resolved

EasyPost.Tests/ReferralTest.cs
nwithan8: Resolved

EasyPost/Billing.cs
Justintime50: That's super annoying, ok.

Justintime50 (approved):

PR #338, opened 2022-09-22
Use dotnet-format instead of broken dotnet format

nwithan8:
# Description

- Swap linting utility for one that isn't broken 🙄 

# Testing

- N/A

# Pull Request Type

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ coverage:
 
 ## format - Formats the project
 format:
-	dotnet format
+	dotnet dotnet-format --no-restore
 
 ## install-cert - Install the PFX certificate to your system (Windows only)
 # @parameters:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -27,7 +27,7 @@ coverage:
 
 ## format - Formats the project
 format:
-	dotnet format
+	dotnet dotnet-format --no-restore
```
Justintime50, line 30: Question: Is it not just `dotnet-format` instead of `dotnet dotnet-format`? Sad if it is... redundant

Makefile
```diff
@@ -36,13 +36,20 @@ format:
...
 	scripts\install_cert.bat ${cert} ${pass}
 
-## install-scanner - Install SecurityCodeScan to your system
-install-scanner:
-	dotnet tool install --local security-scan --version 5.6.3
+## install-tools - Install required dotnet tools
+install-tools:
+	dotnet new tool-manifest || true
+	dotnet tool install --local security-scan --version 5.6.3 || true
+	dotnet tool install --local dotnet-format || true
```
Justintime50, line 43: Prefer `exit 0` for fallbacks as it more clearly defines our expected outcome.

Makefile
nwithan8: `dotnet` brings up the dotnet CLI utility. From there, we tell it to use the `dotnet-format` tool.
You can also summon `dotnet-format` directly if it was installed globally. Our tools are installed locally, so no.

Makefile
nwithan8: Resolved

Justintime50 (approved):

PR #360, opened 2022-10-21
Fix parameter wrapping

nwithan8:
# Description

- Fix bug where parameters were not being wrapped properly, causing issues for embedded elements on creation (i.e. creating a customs item while creating a customs info)

# Testing

- New unit test to confirm customs item created successfully when creating customs info
- New cassettes for unit test
- Re-record all cassettes that need to be wrapped (which is pretty much all of them since almost all actions involve creating a resource)

# Pull Request Type


EasyPost.Tests/ServicesTests/CustomsInfoServiceTest.cs
```diff
@@ -31,6 +32,30 @@ public async Task TestCreate()
...
+
+            CustomsInfo customsInfo = await Client.CustomsInfo.Create(data);
+
+            Assert.IsType<CustomsInfo>(customsInfo);
+            Assert.StartsWith("cstinfo_", customsInfo.Id);
+            Assert.NotEmpty(customsInfo.CustomsItems);
+            foreach (CustomsItem item in customsInfo.CustomsItems)
+            {
+                Assert.StartsWith("cstitem_", item.Id);
+            }
```
Justintime50, line 56: Can we instead just append this assertion to the already existing test? The fixture already has a single item included in the customs info data.

EasyPost/Services/CustomsInfoService.cs
```diff
@@ -34,6 +35,7 @@ internal CustomsInfoService(EasyPostClient client) : base(client)
         [CrudOperations.Create]
         public async Task<CustomsInfo> Create(Dictionary<string, object> parameters)
         {
+            parameters = parameters.Wrap("customs_info");
```
Justintime50, line 38: We seem to be missing this on quite a few items... do we need to be doing this in more places than just here? For instance, `customs_item` isn't wrapped along with others.

EasyPost.Tests/ServicesTests/CustomsInfoServiceTest.cs
nwithan8: Oh, it does? I didn't see that, sounds good.

EasyPost.Tests/ServicesTests/CustomsInfoServiceTest.cs
```diff
@@ -29,6 +29,11 @@ public async Task TestCreate()
             Assert.IsType<CustomsInfo>(customsInfo);
             Assert.StartsWith("cstinfo_", customsInfo.Id);
             Assert.Equal("NOEEI 30.37(a)", customsInfo.EelPfc);
+            Assert.NotEmpty(customsInfo.CustomsItems);
```
Justintime50, line 32: Nit: This assertion doesn't do anything for us since we'll attempt to iterate the collection on the very next line. Inherently if we can't iterate an object and assert something against it then it was empty (AKA: asserting against individual records will get us this assertion on line 32 for free without needing to add an extra line here)

Prefer to keep code as simple and uncluttered as possible

EasyPost/Services/CustomsInfoService.cs
nwithan8: List of all places where parameters are being wrapped (I imagine these are mostly all on `create` calls):
image


I don't see one for customs_info, but the data is clearly being made?

Per the internal conversation, it seems like the absence of the key only causes problems if you have embedded data (i.e. without `customs_info` originally, it was making customs info just fine, but wasn't making the embedded customs items). Perhaps an odd quirk of our API, but one we should definitely guard against.

We'll need to go through and make sure that everything that should be wrapped, is wrapped.

Justintime50: Interesting. I do know that some objects can be wrapped or not and it'll work but some cannot. I'd suggest we make sure each request follows the expectation (see Postman data) just to be safe. Since we're in the weeds here for this, might as well fix any other wrapping items so we capture them all without missing something here.

nwithan8:
Walked through the library, comparing everything to the API docs to fix everywhere where data should be wrapped.

Called out a few places in the code where the data does not need to be wrapped, but is incorrect on the API docs (see TODOs)

Webhook is seeing some degraded performance. The `create` call is showing success, but the webhook is not appearing on the UI dashboard, and when attempting to then `delete` the webhook, we get a "Resource not found" error. As a result, the CI will fail until the issue is rectified

Justintime50 (changes requested):
There is more changed here than asked. (eg: address verification)

We'll regroup on this Monday when hopefully webhooks are working again.

EasyPost/Models/API/Batch.cs
```diff
@@ -47,6 +47,7 @@ internal Batch()
         [CrudOperations.Update]
         public async Task<Batch> AddShipments(Dictionary<string, object> parameters)
         {
+            // parameters = parameters.Wrap("batch");  // TODO: Update docs to remove wrapped "batch" key
```
Justintime50, line 50: `batch` should be wrapped and included here

EasyPost/Models/API/Batch.cs
```diff
@@ -115,6 +115,7 @@ public async Task<Batch> GenerateScanForm(string fileFormat = "pdf")
         [CrudOperations.Update]
         public async Task<Batch> RemoveShipments(Dictionary<string, object> parameters)
         {
+            // parameters = parameters.Wrap("batch");  // TODO: Update docs to remove wrapped "batch" key
```
Justintime50, line 118: `batch` should be wrapped and included here

EasyPost/Services/AddressService.cs
```diff
@@ -34,10 +35,22 @@ internal AddressService(EasyPostClient client) : base(client)
         ///     * {"strict_verifications", List&lt;string&gt;} Possible items are "delivery" and "zip4".
         ///     All invalid keys will be ignored.
         /// </param>
+        /// <param name="verify">Whether to verify this address during the creation process. Default: false</param>
+        /// <param name="verifyStrict">Whether to strictly verify this address during the creation process. Default: false</param>
         /// <returns>EasyPost.Address instance.</returns>
         [CrudOperations.Create]
-        public async Task<Address> Create(Dictionary<string, object> parameters)
+        public async Task<Address> Create(Dictionary<string, object> parameters, bool verify = false, bool verifyStrict = false)
```
Justintime50, line 42: Please rollback these changes. As with other items, we were going to focus on updated params at a later date across all libs. Let's not do that now

EasyPost/Services/ScanFormService.cs
```diff
@@ -25,7 +24,7 @@ internal ScanFormService(EasyPostClient client) : base(client)
         public async Task<ScanForm> Create(List<Shipment> shipments)
         {
             Dictionary<string, object> parameters = new Dictionary<string, object> { { "shipments", shipments } };
-            parameters = parameters.Wrap("scan_form");
+            // parameters = parameters.Wrap("scan_form"); // TODO: Remove if not necessary
```
Justintime50, line 27: This should not be needed (probably one of the only ones) and the comment can be removed

EasyPost/Services/WebhookService.cs
```diff
@@ -30,6 +30,7 @@ internal WebhookService(EasyPostClient client) : base(client)
         [CrudOperations.Create]
         public async Task<Webhook> Create(Dictionary<string, object> parameters)
         {
+            // parameters = parameters.Wrap("webhook");// TODO: Update docs to remove wrapped "webhook" key
```
Justintime50, line 33: Although this may work without, please include it. When we are passing params of an associated object, they should always be wrapped.

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
```diff
@@ -40,9 +40,8 @@ public async Task TestCreateVerify()
             UseVCR("create_verify");
 
             Dictionary<string, object> addressData = Fixtures.IncorrectAddress;
-            addressData.Add("verify", true);
 
-            Address address = await Client.Address.Create(addressData);
+            Address address = await Client.Address.Create(addressData, true);
```
Justintime50, line 44: Please rollback these changes. As with other items, we were going to focus on updated params at a later date across all libs. Let's not do that now

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
```diff
@@ -57,9 +56,8 @@ public async Task TestCreateVerifyArray()
             UseVCR("create_verify_array");
 
             Dictionary<string, object> addressData = Fixtures.IncorrectAddress;
-            addressData.Add("verify", new List<bool> { true });
 
-            Address address = await Client.Address.Create(addressData);
+            Address address = await Client.Address.Create(addressData, true);
```
Justintime50, line 60: Please rollback these changes. As with other items, we were going to focus on updated params at a later date across all libs. Let's not do that now

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
```diff
@@ -74,9 +72,8 @@ public async Task TestCreateVerifyStrict()
             UseVCR("create_verify_strict");
 
             Dictionary<string, object> addressData = Fixtures.CaAddress1;
-            addressData.Add("verify_strict", true);
 
-            Address address = await Client.Address.Create(addressData);
+            Address address = await Client.Address.Create(addressData, verifyStrict: true);
```
Justintime50, line 76: Please rollback these changes. As with other items, we were going to focus on updated params at a later date across all libs. Let's not do that now

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
```diff
@@ -91,9 +88,8 @@ public async Task TestCreateVerifyStrictArray()
             UseVCR("create_verify_strict_array");
 
             Dictionary<string, object> addressData = Fixtures.CaAddress1;
-            addressData.Add("verify_strict", new List<bool> { true });
 
-            Address address = await Client.Address.Create(addressData);
+            Address address = await Client.Address.Create(addressData, verifyStrict: true);
```
Justintime50, line 92: Please rollback these changes. As with other items, we were going to focus on updated params at a later date across all libs. Let's not do that now

Justintime50 (changes requested):
I've reviewed the rest of this diff. Please adjust accordingly and update the PR title and description to match the new changes.

EasyPost/Models/API/Batch.cs
nwithan8: It does not work when wrapped.
image

EasyPost/Models/API/Batch.cs
nwithan8: Same as https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003569828

EasyPost/Services/AddressService.cs
nwithan8: These are needed because "verify" and "verify_strict" are not wrapped with the rest of the parameters in the "address" key.

EasyPost/Services/ScanFormService.cs
nwithan8: Resolved

EasyPost/Services/WebhookService.cs
nwithan8: Resolved

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: See https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003570705

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: See https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003570705

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: See https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003570705

EasyPost/Models/API/Batch.cs
Justintime50: Hmm... I was able to successfully do this in Postman before posting this.

We may need to do some additional testing to find the actual answer.

EasyPost/Services/AddressService.cs
Justintime50: The problem is we have a usage pattern established for years now where these are passed in the dictionary. We should handle something like this with a major release and call it out and instead fix this under the hood since it's what users have expected for years and this major release already went out (AKA: this is a bug we didn't catch in time and shouldn't be a feature change)

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
Justintime50: Replied in https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003578479

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
Justintime50: Replied in https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003578479

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
Justintime50: Replied in https://github.com/EasyPost/easypost-csharp/pull/360#discussion_r1003578479

EasyPost/Services/AddressService.cs
nwithan8: Alright, resolved, but it seems we agree this should be properly addressed in the future.

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: Resolved

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: Resolved

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: Resolved

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
nwithan8: Resolved

Justintime50 (approved):

jchen293 (approved):

PR #363, opened 2022-10-28
Code cleanup

nwithan8:
# Description

- Move Security.cs file to proper directory
- Make sure all models consistently extend base EasyPostObject class
- `myBatch.GenerateLabel()` no longer has a default file format value

# Testing

- No new unit tests required
- All unit tests pass

# Pull Request Type


EasyPost/Models/API/Shipment.cs
```diff
@@ -154,7 +154,7 @@ public async Task Buy(Rate rate, string? insuranceValue = null, bool withCarbonO
         /// </summary>
         /// <param name="fileFormat">Format to generate the label in. Valid formats: "pdf", "zpl" and "epl2".</param>
         [CrudOperations.Update]
-        public async Task<Shipment> GenerateLabel(string fileFormat)
+        public async Task<Shipment> GenerateLabel(string fileFormat = "pdf")
```
Justintime50, line 157: Question: Why have a default instead of requiring users to specify?

Justintime50 (commented):
Do we need to re-record these cassettes? No data changed, the build should have passed without re-recording them.

nwithan8:
> Do we need to re-record these cassettes? No data changed, the build should have passed without re-recording them.

Collateral damage as I was experimenting with changing the HTTP method.

EasyPost/Models/API/Shipment.cs
nwithan8: We can do that too. We have a default on batch generate label, so just want to remain consistent either way.

Justintime50: I don't love that we have one on batch generate label... I'd prefer we didn't include them but we can't just rip out the batch one.

Do we do this in any of our other libs?

Justintime50:
Let's reset the cassettes so they aren't tracked in this PR if possible.

EasyPost/Models/API/Shipment.cs
nwithan8: Java and Python have no default, so we can go with that.

EasyPost/Models/API/Batch.cs
```diff
@@ -89,7 +89,7 @@ public async Task<Batch> Buy()
         /// </summary>
         /// <param name="fileFormat">Format to generate the label in. Valid formats: "pdf", "zpl" and "epl2".</param>
         [CrudOperations.Update]
-        public async Task<Batch> GenerateLabel(string fileFormat = "pdf")
```
Justintime50, line 92: Issue: This is a breaking change, we can't rip out a default here. 

What I meant was although sad and not consistent, let's not add it to the Shipment function. We could note this in code with a TODO to remove the default in an upcoming major release instead?

Justintime50 (changes requested):

EasyPost/Models/API/Batch.cs
nwithan8: Sounds good

Justintime50 (approved):

PR #366, opened 2022-11-02
chore: add fuget link

Justintime50:
# Description

Adds Fuget link for library docs. Because these get handled automatically for us and are versioned into the past, we shouldn't need to build these ourselves locally.

# Testing

# Pull Request Type

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ using the wrong API key when interacting with a resource in a multi-client envir
 
 ## Documentation
 
-API Documentation can be found at: <https://easypost.com/docs/api>.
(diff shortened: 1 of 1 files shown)


jchen293 (approved):

nwithan8:
I would prefer we build these ourselves, just to remain consistent with the other libs (and in case Fuget becomes unavailable in the future)

Justintime50:
> I would prefer we build these ourselves, just to remain consistent with the other libs (and in case Fuget becomes unavailable in the future)

If Fuget ever disappears we can cross that bridge then. Why add overhead when another system already handles this on our behalf? It's the same reason we won't generate them for Go either.

PR #370, opened 2022-11-09
Project, test suite, NuGet package updated to add .NET 7 support

nwithan8:
# Description

- Add net7.0 support to source and unit tests
- Use net7.0 when testing F# and VB compatibility
- Add net7.0 support to nuspec

# Testing

- Tested locally using .NET 7 RC, works as expected
- GitHub CI will test .NET 7 as well
  -~~In draft mode until GitHub Actions runners add .NET 7 support: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#net-core-sdk, https://github.com/actions/runner-images/issues/6539~~ Resolved by forcibly installing .NET versions via [`actions/setup-dotnet`](https://github.com/actions/setup-dotnet) rather than relying on pre-installed .NET on GitHub runner
  - This will help us lock down any .NET version issues.

# Pull Request Type

Please select the option(s) that are relevant to this PR.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

PR #372, opened 2022-11-15
Fix coverage warnings

nwithan8:
# Description

- Fixes coverage warnings raised by the coverage tool (noted here: https://github.com/EasyPost/easypost-csharp/pull/370/files)

# Testing

- N/A

# Pull Request Type

### DIFF
--- a/EasyPost/Calculation/Rates.cs
+++ b/EasyPost/Calculation/Rates.cs
@@ -33,40 +33,50 @@ namespace EasyPost.Calculation
 
             foreach (Rate rate in rates)
             {
-                if ((includeCarriers.Count > 0 || excludeCarriers.Count > 0) && rate.Carrier == null)
+                if (includeCarriers.Count > 0 || excludeCarriers.Count > 0)
                 {
-                    // If we are filtering by carrier and the rate doesn't have a carrier, skip it
-                    continue;
+                    // we have a carrier filter
+
+                    if (rate.Carrier == null)
+                    {
+                        // If we are filtering by carrier and the rate doesn't have a carrier, skip it
+                        continue;
+                    }
+
+                    if (includeCarriers.Count > 0 && !includeCarriers.Contains(rate.Carrier!.ToLower()))
+                    {
+                        // If we have a list of carriers to include and the rate's carrier isn't in the list, skip it
+                        continue;
+                    }
+
+                    if (excludeCarriers.Contains(rate.Carrier!.ToLower()))
+                    {
+                        // If the rate's carrier is in the list of carriers to exclude, skip it
+                        continue;
+                    }
                 }
 
-                if ((includeServices.Count > 0 || excludeServices.Count > 0) && rate.Service == null)
+                if (includeServices.Count > 0 || excludeServices.Count > 0)
                 {
-                    // If we are filtering by service and the rate doesn't have a service, skip it
-                    continue;
-                }
+                    // we have a service filter
 
-                if (includeCarriers.Count > 0 && !includeCarriers.Contains(rate.Carrier!.ToLower()))
-                {
-                    // If we have a list of carriers to include and the rate's carrier isn't in the list, skip it
-                    continue;
-                }
+                    if (rate.Service == null)
+                    {
+                        // If we are filtering by service and the rate doesn't have a service, skip it
+                        continue;
+                    }
 
-                if (excludeCarriers.Contains(rate.Carrier!.ToLower()))
-                {
-                    // If the rate's carrier is in the list of carriers to exclude, skip it
-                    continue;
-                }
+                    if (includeServices.Count > 0 && !includeServices.Contains(rate.Service!.ToLower()))
+                    {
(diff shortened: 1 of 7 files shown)


EasyPost/Calculation/Rates.cs
```diff
@@ -31,37 +31,87 @@ public static Rate GetLowestObjectRate(IEnumerable<Rate> rates, List<string>? in
...
                 {
+                    // if lowest rate is null, set it to this rate
                     lowestRate = rate;
                     continue;
                 }
 
                 if (rate.Price == null || lowestRate.Price == null)
                 {
+                    // if the rate or lowest rate doesn't have a price, throw an exception
                     throw new FilteringError(Constants.ErrorMessages.NullObjectComparison);
```
jchen293, line 92: I don't think it makes sense to use `NullObjectComparsion` for the error message when the price is null since we are not comparing objects here, `NoObjectFound` should be used instead

EasyPost/Calculation/Rates.cs
```diff
@@ -31,37 +31,87 @@ public static Rate GetLowestObjectRate(IEnumerable<Rate> rates, List<string>? in
 
             Rate? lowestRate = null;
 
-            foreach (Rate rate in from rate in rates where includeCarriers.Count <= 0 || includeCarriers.Contains(rate.Carrier.ToLower()) where excludeCarriers.Count <= 0 || !excludeCarriers.Contains(rate.Carrier.ToLower()) where includeServices.Count <= 0 || includeServices.Contains(rate.Service.ToLower()) where excludeServices.Count <= 0 || !excludeServices.Contains(rate.Service.ToLower()) select rate)
```
jchen293, line 34: Thank you!

jchen293 (changes requested):

EasyPost/Calculation/Rates.cs
nwithan8: It's because the price is null, and we can't compare a null price to something else, i.e. NullObjectComparison.

Changing this would introduce a breaking change if users are expecting a specific error message. We'll keep this as is for now.

jchen293: Sure, we should have caught this in the initial thread-safe rewrite

jchen293 (approved):

EasyPost/Calculation/Rates.cs
```diff
@@ -31,37 +31,87 @@ public static Rate GetLowestObjectRate(IEnumerable<Rate> rates, List<string>? in
...
+            foreach (Rate rate in rates)
             {
-                // if lowest rate is null, set it to this rate
+                if (includeCarriers.Count > 0 || excludeCarriers.Count > 0)
+                {
+                    // we have a carrier filter
+
+                    if (rate.Carrier == null)
```
Justintime50, line 40: A rate will always have a carrier.

EasyPost/Calculation/Rates.cs
```diff
@@ -31,37 +31,87 @@ public static Rate GetLowestObjectRate(IEnumerable<Rate> rates, List<string>? in
...
+                        // If the rate's carrier is in the list of carriers to exclude, skip it
+                        continue;
+                    }
+                }
+
+                if (includeServices.Count > 0 || excludeServices.Count > 0)
+                {
+                    // we have a service filter
+
+                    if (rate.Service == null)
```
Justintime50, line 63: A rate will always have a service.

EasyPost/Calculation/Rates.cs
```diff
@@ -76,29 +126,52 @@ public static Smartrate GetLowestShipmentSmartrate(IEnumerable<Smartrate> smartr
         {
             Smartrate? lowestSmartrate = null;
 
-            foreach (Smartrate? smartrate in from smartrate in smartrates let smartrateAccuracy = smartrate.TimeInTransit.GetBySmartrateAccuracy(deliveryAccuracy) where smartrateAccuracy != null where !(smartrateAccuracy > deliveryDays) select smartrate)
+            foreach (Smartrate? smartrate in smartrates)
             {
-                // if lowest smartrate is null, set it to this smartrate
+                if (smartrate.TimeInTransit == null)
```
Justintime50, line 131: SmartRate will always have a Time in Transit property, that's what makes it a smartrate.

Justintime50 (commented):
There are a few checks in here that are misleading and unnecessary. Rates will always have a service and carrier and SmartRates will always have TnT

EasyPost/Calculation/Rates.cs
nwithan8: But, what if it doesn't?

EasyPost/Calculation/Rates.cs
nwithan8: But, what if it doesn't?

EasyPost/Calculation/Rates.cs
nwithan8: But, what if it doesn't?

PR #374, opened 2022-11-17
Undo code signing changes

nwithan8:
# Description

Confusion over the renewed code signing certificate caused a premature change to the code signing/NuGet packing process. Most of those changes have been reverted in this PR, and the process will largely stay the same as it was.

# Testing

- N/A

# Pull Request Type

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,13 @@ coverage-check:
 format:
 	dotnet dotnet-format --no-restore
 
+## install-cert - Install the PFX certificate to your system (Windows only)
+# @parameters:
+# cert= - The certificate to use for signing the built assets.
+# pass= - The password for the certificate.
+install-cert:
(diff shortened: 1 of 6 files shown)


jchen293 (approved):

Justintime50:
Did we use `git revert` here? If not, the PR shouldn't be named `revert` as it's misleading. I'd highly recommend we do use `git revert` if it wasn't used as it will guarantee all the things get put back in the right place and it will preserve git history better.

nwithan8:
> Did we use `git revert` here? If not, the PR shouldn't be named `revert` as it's misleading. I'd highly recommend we do use `git revert` if it wasn't used as it will guarantee all the things get put back in the right place and it will preserve git history better.

No, there are a few things (clean up) that were kept from the original change.

Justintime50 (approved):

PR #376, opened 2022-11-22
Use constants for custom workflow carrier account types

nwithan8:
# Description

- Internally, uses constants from checking custom workflow carrier account type names
- These constants are publicly available for end-users to use if they want
- BREAKING CHANGE: Path to `Constants` file has changed, if users were utilizing it for exception parsing

# Testing

- N/A

# Pull Request Type


EasyPost/Constants.cs
```diff
@@ -90,5 +90,17 @@ public static class ErrorMessages
             internal const string ApiErrorDetailsParsingError = "RESPONSE.PARSE_ERROR"; // not for public consumption
             public const string CouldNotPassClient = "Could not pass client to {0}.";
         }
+
+        public static class CarrierAccountTypes
```
Justintime50, line 94: Question: Do we want to separate normal constants and error strings to separate files?

EasyPost/Constants.cs
```diff
@@ -90,5 +90,17 @@ public static class ErrorMessages
...
+        {
+            public const string FedExAccountType = "FedexAccount";
+            public const string UpsAccountType = "UpsAccount";
+
+            internal static List<string> CustomWorkflowCarrierAccountTypes => new List<string>
+            {
+                FedExAccountType,
+                UpsAccountType
```
jchen293, line 102: Nit: No need to define these variables, I would just pass them directly

jchen293 (approved):

EasyPost/Constants.cs
nwithan8: Having two constants files seems redundant and counter-intuitive to me.

EasyPost/Constants.cs
nwithan8: This way, users can use the constants if they'd like

nwithan8: When building their data for the API call. That way they don't have to remember if it's "FedExAccount" or "FedexAccount" (it's the latter)
image

jchen293: Ah, ok! Makes sense to me now

EasyPost/Constants.cs
```diff
@@ -90,5 +90,17 @@ public static class ErrorMessages
...
+            };
+        }
```
Justintime50, line 104: This is so much code where it's not necessary. KISS.

Let's have it be as simple as this psuedo code:

`CARRIER_TYPES_WITH_CUSTOM_WORKFLOWS = ['FedexAccount', 'UpsAccount']`

We need to be writing the minimal amount of code possible to get tests to pass/behavior met. The more code we have, the more liability we have. Let's design for what we need today (always) and refactor for what we need tomorrow when tomorrow comes because sometimes tomorrow doesn't even come. If ever there comes a day where there is another carrier with a custom workflow (unlikely), we'd just append a new string to the simple list.

TLDR: Let's just use a flat constant instead of a whole class for this.

Justintime50 (changes requested):

EasyPost/Constants.cs
nwithan8: I agree. My one concern is, if someone were to send `type="FedExAccount"` (not FedEx vs Fedex), it wouldn't be captured by the list, and it would attempt to hit the normal endpoint with the data (granted, the data is bad anyway; assuming the API is case-sensitive for account type strings)

Justintime50: Valid, though I thiiink the API would fail because we validate the type there so it may be a non-issue

Justintime50 (approved):

PR #381, opened 2022-12-13
Improve, clean up exception logic, naming and organization

nwithan8:
# Description

- [REMOVED] Removed `UnknownApiError` exception type.
- [CHANGED] Renamed `UnexpectedHttpError` exception type to `UnknownHttpError` to better reflect its purpose.
- [CHANGED] `ExternalApiError` no longer inherits from `ApiError` (`ApiError` reserved from EasyPost API errors only).
- [IMPROVED] All `EasyPostError` exceptions and subclasses now have a `PrettyPrint` getter that returns a human-readable string representation of the error.
  - Previously, only `ApiError` exceptions had this. Now, all exceptions thrown by the library should have this.
- [IMPROVED] Logic for calculating exception type to throw based on API error
  - EasyPost API failures can trigger a variety of specific exceptions, all inheriting from `ApiError`.
  - Non-EasyPost API/HTTP failures will trigger an `ExternalApiError` exception.

# Testing

- Unit tests updated as needed

# Pull Request Type


Justintime50 (changes requested):
Many of these are breaking changes. We discussed we wouldn't be doing this for now.

PR #322, opened 2022-09-08
Improve style guide

nwithan8:
# Description

- Change style guide to compress dictionary/list initializers
- Apply new styling rules

# Testing

- No impact on functionality, tests

# Pull Request Type

Please select the option(s) that are relevant to this PR.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

PR #195, opened 2022-02-01
Add linting to CI

nwithan8:
Let's try to keep our code style consistent. This PR:
- Adds a linting step (using `dotnet-format`) to our GitHub CI so all PR pushes will be run through the check
- Adjusts some pre-existing formatting issues
- Tweaks our `.editorconfig` file to remain consistent with what `dotnet-format` expects.*

*For `.editorconfig` ruleset and descriptions, see [here](https://www.jetbrains.com/help/resharper/EditorConfig_Index.html).

PR #383, opened 2022-12-22
Enable consistent static analysis

nwithan8:
# Description

Recent versions of C# IDEs like Visual Studio and Rider come with [static analysis](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-6) out of the box, but the settings for these can vary from user to user.

To standardize this, we are utilizing the `.editorconfig` file in this codebase to implement specific static analysis rules.

This PR:
- Forcibly enables static analysis during build process, using all the latest rules
  - This does not seem to work as expected, so just in case (see below)
- Adds generic rule to `.editorconfig` that will throw warnings for all static analysis violations ([rules](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/)).
  - These are warnings, not errors, and therefore won't prevent a build, but will likely present a lot of suggestions to the end-user about how to improve their code.
  - A number of code quality analysis ("CAxxxx")  and code-style analysis ("IDExxxx") rules have been disabled or suppressed via the `.editorconfig` file to match our preferred .NET style.
- Adds [StyleCop](https://github.com/DotNetAnalyzers/StyleCopAnalyzers) as a NuGet dependency (only as a development dependency; this will not be included in our final NuGet package nor required to install our NuGet package in user's own applications).
  - StyleCop works in tandem with the Roslyn static analysis tool, while also serving as a backup in case a developer does not have Roslyn enabled in their IDE.
  - StyleCop comes with [its own set of rules](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/DOCUMENTATION.md), some of which have been purposefully disabled/overridden via the `.editorconfig` file to match our preferred .NET style.
- Enables .NET static analysis on CI with [an Action](https://github.com/dotnet/code-analysis) using a built-in Microsoft static analysis tool.
  - This tool seems to be better than CodeQL (see below) at finding stylistic issues, but the Action itself is a bit flaky. This action will never cause a build to fail; everything logged by it should be a warning, not an error.
- Makes several stylistic and cosmetic changes to the source code, abiding by styling rules from Roslyn, StyleCop and CodeQL that we have decided we'd like to adhere to.
  - Rules we do not want to enforce are suppressed project-wide via the `.editorconfig` file.
  - Some rules are circumvented with one-off suppressions via comments in the source code (similar to inline `#type ignore` mypy comments in Python)
- Improves many docstrings (StyleCop doesn't like missing or incomplete docstrings).

This task is ultimately another part in [our efforts](https://github.com/EasyPost/easypost-csharp/pull/192) [to establish](https://github.com/EasyPost/easypost-csharp/pull/195) [a .NET style guide](https://github.com/EasyPost/easypost-csharp/pull/322) for EasyPost using the `.editorconfig` file. The style guide is ever-changing and evolving, but especially now that we have some more experience with .NET, we can better made decisions about styling.

# Testing

- All unit tests pass
  - Only [one small security change](https://github.com/EasyPost/easypost-csharp/pull/383/commits/f7ec1b2d0f30b672ae44cb028929e3606e70da7c) was made that could potentially (though unlikely) impact end-users. All other changes are purely stylistic and cosmetic, and have no impact on the functionality of the library, nor do they change any existing methods or properties in terms of spelling, capitalization, etc.

# Pull Request Type


jchen293 (approved):

Justintime50 (approved):
Hmm, we should probably fix these things as a part of this effort, otherwise we just introduced something that throws 2000 warnings without resolving any heh

nwithan8:
All changes being made are styling changes and spelling errors. No user-facing functionality will be changed.

Justintime50 (changes requested):
This PR has absolutely ballooned. We should chat before moving on it.

EasyPost/Http/ClientConfiguration.cs
```diff
@@ -14,10 +14,10 @@ public class ClientConfiguration
         ///     This cannot be changed after the client has been initialized.
         /// </summary>
         // This cannot be changed, because the internal RestClient is initialized with this value immediately.
-        public readonly string ApiBase;
+        internal readonly string ApiBase;
```
Justintime50, line 17: This is a breaking change, please keep these public.

EasyPost/Utilities/Cryptography.cs
```diff
@@ -118,35 +121,41 @@ private static byte[] AsByteArray(this string str, Encoding? encoding = null)
...
+        private static string AsHexString(this IReadOnlyList<byte> bytes)
         {
             // Fastest safe way to convert a byte array to hex string,
             // per https://stackoverflow.com/a/624379/13343799
-
             uint[] lookup32 = Lookup32;
-            char[] result = new char[bytes.Length * 2];
-            for (int i = 0; i < bytes.Length; i++)
+            char[] result = new char[bytes.Count * 2];
+            for (int i = 0; i < bytes.Count; i++)
```
Justintime50, line 131: Does `Length` and `Count` actually do the same thing here?

EasyPost/Utilities/RuntimeInfo.cs
```diff
@@ -31,49 +31,43 @@ internal static string ApplicationVersion
...
-                        default:
-                            return "Unknown";
-                    }
+                        PlatformID.Win32S => "Windows",
+                        PlatformID.Win32Windows => "Windows",
+                        PlatformID.Win32NT => "Windows",
+                        PlatformID.WinCE => "Windows",
+                        PlatformID.Unix => "Linux",
+                        PlatformID.MacOSX => // in newer versions, Mac OS X is PlatformID.Unix unfortunately
+                            "Darwin",
```
Justintime50, line 67:
```suggestion
                        PlatformID.MacOSX => "Darwin", // in newer versions, Mac OS X is PlatformID.Unix unfortunately
```

EasyPost/Utilities/RuntimeInfo.cs
```diff
@@ -31,49 +31,43 @@ internal static string ApplicationVersion
...
+                        PlatformID.Xbox => "Unknown",
```
Justintime50, line 68: Please add `// try it, I dare you` comment here

EasyPost/_base/EasyPostClient.cs
```diff
@@ -5,27 +5,30 @@
...
 using RestSharp;
 
+#pragma warning disable SA1300
 namespace EasyPost._base
+#pragma warning restore SA1300
 {
     public abstract class EasyPostClient
     {
-        public readonly ClientConfiguration Configuration;
+        internal readonly ClientConfiguration Configuration;
```
Justintime50, line 22: I want users to be able to pull out things like their api key or timeouts so they can troubleshoot if necessary

Justintime50 (dismissed):
1. Let's consolidate the CI files back into one for now, remove extra comments, and simplify where possible.
2. Please fix the various instances of the following similar errors: `Add the following constructor to ExternalApiError: public ExternalApiError()`

EasyPost/Utilities/Cryptography.cs
nwithan8: Yes. In fact, if I try to write `.Length`, my IDE automatically changes it to `.Count`:

Count is preferred, but could be mutable on a normal array, whereas Length gives a consistent result. Since this is an immutable list, Count == Length.
https://softwareengineering.stackexchange.com/a/294619

EasyPost.Tests/ServicesTests/ShipmentServiceTest.cs
```diff
@@ -145,23 +145,23 @@ public async Task TestAll()
...
-            Shipment shipment = await CreateBasicShipment();
+            Shipment shipment = await Client.Shipment.Create(Fixtures.BasicShipment);
 
-            // test lowest smartrate with valid filters
-            List<Smartrate> smartrates = await shipment.GetSmartrates();
-            Smartrate lowestSmartrate = ShipmentService.GetLowestSmartrate(smartrates, 2, SmartrateAccuracy.Percentile90);
-            Assert.Equal("Priority", lowestSmartrate.Service);
-            Assert.Equal(8.15, lowestSmartrate.Rate);
-            Assert.Equal("USPS", lowestSmartrate.Carrier);
+            // test lowest Smart Rate with valid filters
```
Justintime50, line 154: Heh these should be `SmartRate` per my suggestion (no space)

Justintime50 (approved):

jchen293 (approved):

Issue #190, opened 2022-01-27
New RestSharp library


dolphinsd:
If use latest restsharp, version 107, see https://restsharp.dev/v107/#restsharp-v107
and this library, it causes errors!
Method not found: 'RestSharp.IRestRequest RestSharp.RestRequest.AddHeader(System.String, System.String)'.

Justintime50:
Thanks for reporting this! We also have an internal task for this issue and are actually looking into it now. We'll update you here when we have more info.

Justintime50:
After digging into this a bit, RestSharp `v107` is not compatible with this library. We have plans in the coming couple weeks to upgrade to v107 in this library and make the necessary compatibility changes. We'll also be pinning RestSharp to 107 so that these breaking changes cannot happen again in the future when a new major version comes out.

My suggestion for now is to pin RestSharp to `v106` in your application until we've made the tweaks necessary in this library at which point you could upgrade.

Issue #147, opened 2020-05-08
RestSharp compatibility


davitgr:
We are using the EasyPost .Net Core libraries. The library is incompatible with newer versions of RestSharp. Will you be adding support? The latest version is 106.11.2, but the EasyPost libraries are only compatible with 106.4.2.

Issue #123, opened 2018-10-10
RestSharp issue


dharamhbtik:
I have updated to latest version of EasyPost but while calling create method I am getting below error
Method not found: 'Void RestSharp.RestRequest..ctor(System.String, RestSharp.Method)'.

EasyPost Version : 2.5.1
RestSharp Version : 106.5.2

Any help on this?

jmalatia:
@dktechindia  What .net version are you targeting?

dktechindia:
its 4.7.1

jmalatia:
@dktechindia I can verify that you are correct with RestSharp version 106.5.2 that error is encountered.

@att14  The problem does not occur with RestSharp 106.4.2... I tested targeting `net472`

Full Error I received:
```
Method not found: 'Void RestSharp.RestRequest..ctor(System.String, RestSharp.Method)'.

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.MissingMethodException: Method not found: 'Void RestSharp.RestRequest..ctor(System.String, RestSharp.Method)'.

Source Error: 
Line 85:             };
Line 86: 
Line 87:             shipment.Buy(shipment.LowestRate(
Line 88:                 includeServices: new List<string>() { "Priority" },
Line 89:                 includeCarriers: new List<string>() { "USPS" }

Stack Trace: 
[MissingMethodException: Method not found: 'Void RestSharp.RestRequest..ctor(System.String, RestSharp.Method)'.]
   EasyPost.Request..ctor(String IResource, Method method) +0
   EasyPost.Shipment.sendCreate(Dictionary`2 parameters) +35
   EasyPost.Shipment.Create() +27
   EasyPost.Shipment.GetRates() +20
   EasyPost.Shipment.LowestRate(IEnumerable`1 includeCarriers, IEnumerable`1 includeServices, IEnumerable`1 excludeCarriers, IEnumerable`1 excludeServices) +82
   Ep472.Controllers.HomeController.Easypost() in C:\Dev\Ep472\Ep472\Controllers\HomeController.cs:87
   lambda_method(Closure , ControllerBase , Object[] ) +62
   System.Web.Mvc.ActionMethodDispatcher.Execute(ControllerBase controller, Object[] parameters) +14
   System.Web.Mvc.ReflectedActionDescriptor.Execute(ControllerContext controllerContext, IDictionary`2 parameters) +169
[...]

PR #384, opened 2022-12-27
Performance improvements to HTTP connections

nwithan8:
# Description

Inspired by some old issues on this repo (one in particular that I can't seem to find re: our old singleton client re-initializing and exhausting resources) and some personal research on the topic, this is a proposed improvement to our client library's HTTP capabilities, primarily by **dropping RestSharp as a dependency**.

RestSharp is arguably an outdated solution to an underdeveloped built-in HTTP client in .NET that, over the years, has seen improvements that make RestSharp almost redundant (the same way that the built-in JSON serialization capabilities will eventually make Newtonsoft irrelevant).

This library especially has evolved to the point that it no longer needs RestSharp. In fact, since implementing a custom JSON serializer several months ago, it wasn't really even using RestSharp for its JSON serialization capabilities, which is really the whole point of RestSharp in the first place. Because of this, it was relatively easy to simply remove RestSharp (which, as of v107, is effectively a wrapper around HttpClient with some additional functionality that we weren't even using) and replace it with a direct HttpClient instead.

This PR:
- Replaces the RestSharp client under-the-hood with an HttpClient
- Replaces all RestSharp abstractions with their HttpClient equivalent
- BREAKING CHANGE: Removes RequestTimeoutMilliseconds (the only valid timeout is ConnectTimeoutMilliseconds)
- Introduces a new Utilities.Http.Method class used to account for PATCH, which is missing in HttpClient in .NET Standard 2.0
- Updates adjacent JSON serialization and request construction classes to account for the RestSharp drop
- Consolidates the request construction logic, combining two needlessly-separate functions into one
- Makes the Client (the user-facing object, not the internal HttpClient) disposable, allowing users to properly dispose of a Client instance for better memory management in their applications.

There are other improvements slated for potential adoption (e.g. [HttpClientFactory](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests)) in later PRs.

This change will remove an extraneous dependency, one whose breaking changes [have caused](https://github.com/EasyPost/easypost-csharp/issues/190) [us issues](https://github.com/EasyPost/easypost-csharp/issues/147) [in the past](https://github.com/EasyPost/easypost-csharp/issues/123), and reduce our library to only one external dependency (Newtonsoft.Json, which currently may still be irreplaceable).

# Testing

- All unit tests pass
  - Slight modifications to asserted values, primarily for exception testing due to a slight change to how exception pretty print messages are created.
  - No cassettes needed to be re-recorded, suggesting that the underlying HTTP requests and responses were constructed and recorded the same as when RestSharp was being used.

# Pull Request Type

### DIFF
--- a/EasyPost.Tests.FSharp/packages.lock.json
+++ b/EasyPost.Tests.FSharp/packages.lock.json
@@ -94,11 +94,6 @@
         "resolved": "5.11.0",
         "contentHash": "eaiXkUjC4NPcquGWzAGMXjuxvLwc6XGKMptSyOGQeT0X70BUZObuybJFZLA0OfTdueLd3US23NBPTBb6iF3V1Q=="
       },
-      "RestSharp": {
-        "type": "Transitive",
-        "resolved": "108.0.1",
-        "contentHash": "XKwuwWq7A4dxZ8l9QkS2+ePLJ9ImmLaJSeFO6H2kpII4Us4n1NRs4w3c//8H2BYpL1XqBE8nmsd9aNCVBwXmOA=="
-      },
       "System.Reflection.Metadata": {
         "type": "Transitive",
         "resolved": "1.6.0",
@@ -152,8 +147,7 @@
       "easypost": {
         "type": "Project",
         "dependencies": {
-          "Newtonsoft.Json": "[13.0.1, 14.0.0)",
-          "RestSharp": "[108.0.1, 109.0.0)"
+          "Newtonsoft.Json": "[13.0.1, 14.0.0)"
         }
       }
     }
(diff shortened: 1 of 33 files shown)


Justintime50 (commented):
This seems like a worthy change, always a big fan of less dependencies. We can't move on this however until the next major release but let's keep it around until then.

We will also definitely want to re-record all cassettes at that time. Although it is "constructing" the requests the same, it's not actually making any without re-recording heh. We'd want to re-record them all to make sure all our weird edge cases get run through the different HTTP client flow.

nwithan8:
Closing in favor of #433

PR #395, opened 2023-01-24
[chore] Add docs pages

nwithan8:
# Description

- Use docfx to generate docs pages from docstrings
- Add "make docs" step to Makefile

Final product:
image

# Testing

- N/A

# Pull Request Type


Makefile
```diff
@@ -45,6 +49,7 @@ install-tools:
 	dotnet new tool-manifest || exit 0
 	dotnet tool install --local security-scan --version 5.6.3 || exit 0
 	dotnet tool install --local dotnet-format || exit 0
+	dotnet tool update -g docfx || exit 0
```
Justintime50, line 52: We should be installing locally, not globally. Can we change that?

Justintime50 (commented):
Please also include the README link like I did with all the others this morning. Copy/paste

Justintime50:
> I'll update the README in a later PR, need to confirm that the final URL path will look how I think it will look.

They all have the same path but change the lib name. Can we not use the same process as the other libs and simply serve the `docs` directory?

nwithan8:
> > I'll update the README in a later PR, need to confirm that the final URL path will look how I think it will look.
> 
> They all have the same path but change the lib name. Can we not use the same process as the other libs and simply serve the `docs` directory?

Yes, it will. I just want to confirm the link before I update the README, since I think it will have to go into the "api" subdirectory to reach the index page. But the main domain will be correct.

Makefile
```diff
@@ -49,7 +49,7 @@ install-tools:
 	dotnet new tool-manifest || exit 0
 	dotnet tool install --local security-scan --version 5.6.3 || exit 0
 	dotnet tool install --local dotnet-format || exit 0
-	dotnet tool update -g docfx || exit 0
+	dotnet tool update --local docfx || exit 0
```
Justintime50, line 52: Why update when the others are install?

nwithan8: 🤷🏻  Was just following the quick-start guide from DocFX. I can make it install since this realistically is a one-time step. Just means if you run it in the future, it won't update your existing installation.

Justintime50: Install is ideal since the Makefile target is install. Updating would be a side-effect and we never want those.

Justintime50 (approved):

nwithan8:
This PR contains the config file and the GitHub Action steps to generate the API docs when a push to master occurs.

The generated HTML files and the middle-man YAML files are not included in this PR. Those are generated during the CI process.

As a result, there are no plain HTML files in this repo to look at. They are only available on the GitHub page.

nwithan8:
@Justintime50 @bnbry We'll need to switch the Pages option for this repository to "Build from branch" rather than the new Jekyll option.

There is a GitHub Action added here that will auto-generate and deploy the docs on a push to master (or when manually triggered)

PR #398, opened 2023-01-25
[feat] Add stateless rating under beta

nwithan8:
# Description

- Added new beta `RateService`, accessible via `myClient.Beta.Rate`.
- Added `RetrieveStatelessRate` function under beta `RateService` to pull stateless rates when shipment data is provided
- Adds `GetLowestStatelessRate` function under `Calculations.Rate` to filter the lowest stateless rate
- Adds new `GetLowest` instance functions, callable on `List<Rate>`, `List<Smartrate>` and `List<StatelessRate>`, to filter the lowest rate
- Add new `EphemeralEasyPostObject` base class, used by objects that don't have IDs/created_at/updated_at properties
- Deprecate legacy rate filtering methods in `EasyPost.Calculations.Rate`, `ShipmentService` and `RateService` namespaces, redirect to new methods in `EasyPost.Utilities.Rates` namespace

# Testing

- New unit tests, cassettes

# Pull Request Type


EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        [Testing.Function]
+        public async Task TestGetLowestRateInstanceFunction()
+        {
+            UseVCR("get_lowest_rate_instance_function");
+
+            Dictionary<string, object> data = Fixtures.BasicShipment;
+
+            List<StatelessRate> rates = await Client.Beta.Rate.RetrieveStatelessRates(data);
+
+            StatelessRate lowestStatelessRate = Calculation.Rates.GetLowestStatelessRate(rates);
```
Justintime50, line 48: hmm, I'd expect user-facing functions to be in `Utils`, not `Calculation`.

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        [Testing.Function]
+        public async Task TestGetLowestRateStaticFunction()
+        {
+            UseVCR("get_lowest_rate_static_function");
+
+            Dictionary<string, object> data = Fixtures.BasicShipment;
+
+            List<StatelessRate> rates = await Client.Beta.Rate.RetrieveStatelessRates(data);
+
+            StatelessRate lowestStatelessRate = Client.Beta.Rate.GetLowestStatelessRate(rates);
```
Justintime50, line 63: Let's have this only available as a util, like in Python since the client isn't needed for this standalone helper function

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        [Testing.Function]
+        public async Task TestGetLowestRateExtensionFunction()
+        {
+            UseVCR("get_lowest_rate_extension_function");
+
+            Dictionary<string, object> data = Fixtures.BasicShipment;
+
+            List<StatelessRate> rates = await Client.Beta.Rate.RetrieveStatelessRates(data);
+
+            StatelessRate lowestStatelessRate = rates.GetLowest();
```
Justintime50, line 78: How does this even work?...

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
```diff
@@ -44,7 +44,7 @@ public async Task TestAll()
 
             List<ScanForm> scanForms = scanFormCollection.ScanForms;
 
-            Assert.True(scanFormCollection.HasMore);
+            // Assert.True(scanFormCollection.HasMore);  // Not guaranteed to have more
```
Justintime50, line 47: mmm it should always have more?

Justintime50 (changes requested):
The fact that so many cassettes needed re-recording when it hasn't yet been 6 months tells me we made some extra unintended changes here that shouldn't be included.

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
nwithan8: Should I remove the non-API filtering functions from the normal Rate service (filtering rates and smart rates) as well?

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
nwithan8: Extension method magic in C#, basically an method that's available as an instance on a particular type (peep the `this` keyword on the first parameter):
https://github.com/EasyPost/easypost-csharp/blob/71e732e48c3e6ea257c956628dcf33b3bef9907f/EasyPost/Calculation/Rates.cs#L127

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
nwithan8: The cassette I recorded didn't return more than 20 scan forms/didn't have a second page, so `HasMore` was false.

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
Justintime50: No, removing things is a breaking change

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
Justintime50: Well, this assertion is wrong, we really want to be asserting the existence of the key. If C# doesn't even populate the key if it's false, then let's just remove this instead of commenting it out.

nwithan8: It will populate the key one way or another (either with the value from JSON or `false` by default if the JSON key does not exist). The `HasMore` boolean will always be **something**

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
nwithan8: Calculation is public, but realistically was not meant for end-user consumption. Users are expected to access rate filtering via the Rate service (even though it doesn't make an API call).

If we want to preserve "services" for solely API call functions and move all non-API call functions into a set of publicly-accessible utilities, that's a design choice we should discuss.

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
nwithan8: https://github.com/EasyPost/easypost-csharp/pull/398#discussion_r1088365090

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
Justintime50: We did, it's just C# is now behind. Java and PHP along with soon-to-be Node have their non-API functions in the Utils namespace.

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
Justintime50: Ok, then yeah let's change the bad assertion to test that the key is there. This is important because in some libs we were previously not returning this key so we want to protect against the regression to ensure people can paginate. The content of said key isn't important.

nwithan8:
> The fact that so many cassettes needed re-recording when it hasn't yet been 6 months tells me we made some extra unintended changes here that shouldn't be included.

I believe I've diagnosed the issue.

As part of the new `EphemeralEasyPostObject` base class, `mode` and `object` existing in this base class, while `created_at`, `updated_at` and `id` exist in the subclass `EasyPostObject`. This split has inadvertently altered the order in which properties of these objects are serialized into JSON. Namely, "mode" used to exist between "id" and "updated_at", but now is positioned after "updated_at" in the final JSON body. The order of keys obviously doesn't matter in a JSON body, but the strict body matching enforcement on our VCR means that the new order does not match the old order, and is therefore failing.

Comparing the old and new cassettes manually, I can confirm that no new data was included (other than expected changed values, like IDs and datetimes), nor was any old data excluded. Just a simple re-order of JSON keys getting in the way.

Justintime50:
@nwithan8 that makes sense, thanks for looking into it!

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
nwithan8: Yeah, type-strict languages are always going to have the property, versus dynamic languages were we don't know if the property exists.

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
nwithan8: Should we preserve the "old" way and fix this all at once in the future, or start splitting it now?

Justintime50: Oh, my suggestion is that since this is a fresh function, we set it up the right way, how we want things to be, and in the next major release we can swap things we don't love to match.

nwithan8: Alright, the legacy methods have been marked as deprecated and redirect to the functions now located in "Utilities"

EasyPost/Models/API/Beta/StatelessRate.cs
```diff
@@ -0,0 +1,50 @@
...
+        [JsonProperty("delivery_days")]
+        public int? DeliveryDays { get; set; }
+        [JsonProperty("est_delivery_days")]
+        public int? EstDeliveryDays { get; set; }
+        [JsonProperty("list_currency")]
+        public string? ListCurrency { get; set; }
+        [JsonProperty("list_rate")]
+        public string? ListRate { get; set; }
+        [JsonProperty("rate")]
+        public string? Price { get; set; } // "Rate" is the enclosing class name
```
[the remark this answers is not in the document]
nwithan8: `Rate` is the name of the class. In .NET, you can't have a property of a class that is the same name as the enclosing class.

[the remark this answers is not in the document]
nwithan8: I realize now this is on StatelessRate, but to remain consistent with how it works in the normal `Rate` class (and since this will eventually become the default `Rate` class)

EasyPost.Tests/ServicesTests/Beta/RateServiceTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        #region Test CRUD Operations
+
+        [Fact]
+        [CrudOperations.Read]
+        [Testing.Function]
+        public async Task TestRetrieve()
+        {
+            UseVCR("retrieve");
+
+            Dictionary<string, object> data = Fixtures.BasicShipment;
```
Justintime50, line 29: `data` is too ambiguous, please describe what the data is in the var. `params` or `shipmentData` are good options. This is true for all the `data` vars in this file.

```suggestion
            Dictionary<string, object> shipmentData = Fixtures.BasicShipment;
```

jchen293 (approved):

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
```diff
@@ -44,7 +44,6 @@ public async Task TestAll()
 
             List<ScanForm> scanForms = scanFormCollection.ScanForms;
 
-            Assert.True(scanFormCollection.HasMore);
```
Justintime50, line 47: I believe we still want this assertion even though it will always be present due to the type-strict nature of the lib, it will protect against a regression of someone removing it, or it getting overridden per object.

EasyPost/Calculation/Rates.cs
```diff
@@ -17,148 +15,29 @@ public static class Rates
...
+        ///     Get the lowest rate from a list of ephemeral rates.
+        /// </summary>
+        /// <param name="rates">List of rates to parse.</param>
+        /// <param name="includeCarriers">Carriers to include in the filter.</param>
+        /// <param name="includeServices">Services to include in the filter.</param>
+        /// <param name="excludeCarriers">Carriers to exclude in the filter.</param>
+        /// <param name="excludeServices">Services to exclude in the filter.</param>
+        /// <returns>Lowest rate matching the filter.</returns>
+        [Obsolete("This method is deprecated. Please use EasyPost.Utilities.Rates.GetLowestStatelessRate() instead. This method will be removed in a future version.", false)]
+        public static EasyPost.Models.API.Beta.StatelessRate GetLowestStatelessRate(IEnumerable<EasyPost.Models.API.Beta.StatelessRate> rates, List<string>? includeCarriers = null, List<string>? includeServices = null, List<string>? excludeCarriers = null, List<string>? excludeServices = null) => Utilities.Rates.GetLowestStatelessRate(rates, includeCarriers, includeServices, excludeCarriers, excludeServices);
```
Justintime50, line 31: Don't deprecate a function that never existed please, let's just remove it outright. No need to add dead code that is about to get yanked out.

Justintime50 (changes requested):
Looks like the instance-based `GetLowestStatelessRate` function still exists and it needs to be removed. Once complete, the PR summary and CHANGELOG will need to be updated to reflect this.

EasyPost.Tests/ServicesTests/ScanFormServiceTest.cs
nwithan8: Not sure why this assertion failed (was false) earlier but is true now...

nwithan8: image

nwithan8: Per our internal discussions:

- This assertion will be removed, because we cannot guarantee the data returned by the API (when recording a cassette) will always return `true` for "has_more"
- The assertion is more so after confirming that "has_more" was properly deserialized, testing presence rather than value. This will be offloaded onto a standalone JSON de/serialization unit test.

nwithan8: JSON de/serialization unit tests already exist to confirm that JSON data gets deserialized into models properly: https://github.com/EasyPost/easypost-csharp/blob/27037d5c2d0c2bfeeba457e832ca859bf2e4281d/EasyPost.Tests/UtilitiesTests/JsonSerializationTest.cs#L11

EasyPost.Tests/CalculationTests/RatesTests.cs
```diff
@@ -81,31 +81,31 @@ public void TestLowestRateFilters()
...
             Assert.Equal("CarrierB", lowestRate.Carrier);
 
             // It may go with a higher price if filters are involved
-            lowestRate = Calculation.Rates.GetLowestObjectRate(rates, new List<string> { "CarrierC" });
+            lowestRate = Utilities.Rates.GetLowestRate(rates, new List<string> { "CarrierC" });
             Assert.Equal("100.00", lowestRate.Price);
             Assert.Equal("CarrierC", lowestRate.Carrier);
 
-            lowestRate = Calculation.Rates.GetLowestObjectRate(rates, includeServices: new List<string> { "ServiceC" });
+            lowestRate = Utilities.Rates.GetLowestRate(rates, includeServices: new List<string> { "ServiceC" });
```
Justintime50, line 108: Oof, we were using the internal interface and not the user-facing one heh. Glad this is fixed.

CHANGELOG.md
```diff
@@ -4,90 +4,107 @@
...
 
 ## v4.2.0 (2023-01-11)
 
 - Added new beta billing functionality for referral customer users, accessible via `myClient.Beta.Referral` service
-  - `AddPaymentMethod` to add an existing Stripe bank account or credit card to your EasyPost account
-  - `RefundByAmount` refunds you wallet balance by a specified amount
-  - `RefundByPaymentLog` refunds you wallet balance by a specified payment log
+    - `AddPaymentMethod` to add an existing Stripe bank account or credit card to your EasyPost account
+    - `RefundByAmount` refunds you wallet balance by a specified amount
+    - `RefundByPaymentLog` refunds you wallet balance by a specified payment log
```
Justintime50, line 23: Please put-back all the changes to CHANGELOG, should only be the new entries that got affected

Justintime50:
I think this needs a rebase now that you've landed the assertion fixes? Or was this one not included in that other PR?

EasyPost.Tests/CalculationTests/RatesTests.cs
nwithan8: To be fair, there was no user-facing one (or at least, "Calculation" was the user-facing one).

Justintime50 (approved):
There's a lot here to sift through, it appears all is in order though

PR #405, opened 2023-02-14
Re-enable strong-naming

nwithan8:
# Description

After investigation, it appears we erroneously disabled strong-naming of our assemblies back in November while re-issuing our code signing certificate. This has inadvertently created issues for downstream users on .NET Framework, which relies on strong-named assemblies (.NET Core and modern .NET users should not be impacted).

This PR restores the strong-naming process, re-introduces our code signing public key into the repository and clarifies the strong-naming process to prevent such an incident in the future.

- Re-enable strong-name signing during release build process (disabled during debug build)
- Add public key to repository
- Clarify build scripts, split strong name signing and authenticity signing into two steps

This PR also manually adds .NET 5 to the CI, as that was [deprecated and removed from GitHub runners in February](https://github.com/actions/runner-images/issues/6840).

# Testing

- Public key generated (will share scripts internally) and used without error during test build in Release mode
- Further testing, verification of result needed on a Windows machine (confirmed)

# Pull Request Type


scripts/sign_nuget.bat
```diff
@@ -9,9 +9,9 @@
 SET certFile=%1
 SET certPass=%2
 
-:: Sign all NuGet packages found
+:: Sign all NuGet packages found with our certificate to guarantee authenticity
 @ECHO:
-@ECHO Signing NuGet package with %certFile%...
+@ECHO Authenticating NuGet package with %certFile%...
```
Justintime50, line 14: I don't know that the changes of `Signing` to `Authenticating` is appropriate. Think of a login system, we authenticate the user against a well-known set of creds that belong to them to make sure the person logging in is who they say they are (but this is an after-the-fact operation) meaning we have their creds already to authenticate against. In this case, we are signing our DLLs so that a system in the future can then authenticate that this came from EasyPost.

Thoughts? This is true of all the changes of `signing` to `authenticating` throughout this PR

nwithan8: Fair enough, I'm not a huge fan of the change either. Wanted to clarify the difference between strong-name signing and signing for authenticity like a hash, any synonym suggestions?

Justintime50 (approved):

nwithan8:
After testing, it would appear that, for one reason or another, using our code-signing certificate both to sign our DLLs and NuGet package files for authenticity and to sign our assemblies for strong-naming is incompatible.

Since our research turned up that strong-name signing is not a security measure (so much so to the point that [Microsoft recommends checking in the private key associated with a strong-name certificate into a public repo](https://github.com/dotnet/runtime/blob/main/docs/project/strong-name-signing.md)), it is wise to take this opportunity to formally split the single certificate into two.

This PR has introduced a public key (`[EasyPostNETStrongNameSigning.pub`), which is the public key extracted from a new strong-name-signing-specific key-pair. This public key will exist in the repository, and is used to "delay-sign" the assemblies (the assemblies are partially signed with the public key, with the expectation of finishing signing with the private key prior to release). The build process scripts have been updated to use the associated private key to complete the strong-name signing during the release process.

I have personally verified via Microsoft's Strong-Name utility that a) old DLLs (i.e. .NET Standard 2.0 from v4.2.0) were not strongly-typed and b) DLLs produced by this new process are strongly-typed.

This strong-name process is now separate from the pre-existing signing process, which is still intact. After compilation and strong-signing, the final DLLs will be signed with EasyPost's authenticity certificate (not shared in this repository for obvious reasons). The DLLs are then packages into a NuGet package file, which is also signed with the authenticity certificate.

jchen293 (approved):

Justintime50 (approved):

PR #408, opened 2023-02-21
Port release process from Windows to Unix

nwithan8:
# Description

After investigation, it appears the only step in our build-release process for our .NET library that was locked to Windows was signing our DLLs for authenticity, which relied on the Windows-only SignTool. It appears that [`osslsigncode`](https://github.com/mtrojnar/osslsigncode) operates as a perfect drop-in replacement for the tool on Unix systems.

All other steps in the process could run directly on Unix as they did on Windows.

As a result, our build-release process can now run entirely on Unix systems, including MacOS, which will make our releasing process far easier and remove our reliance on a Windows machine.

This PR ports all Windows Batch scripts used for our release process to bash scripts, and alters our Makefile steps to use the bash scripts rather than the Batch scripts.

This PR also cleans up some existing inconsistencies and outdated comments.

Our Windows Batch scripts will continue to exist, but are no longer configured to be used by our Make steps. Batch script linting (will eventually be phased out) and Windows setup are the only two remaining Batch script steps configured in our Makefile; all other steps are designed to be run on Unix.

# Testing

I have personally verified that each step of the process executes as expected:
- Old files are cleaned up
- Code is compiled
- Assemblies are strong-signed
- Assemblies are signed for authenticity
- Assemblies are packaged
- Package is signed for authenticity

# Pull Request Type


Makefile
```diff
@@ -23,11 +23,11 @@ clean:
 
 ## coverage - Generate coverage reports for the project
 coverage:
-	./generate_test_reports.sh
+	bash scripts/unix/generate_test_reports.sh
```
Justintime50, line 26: We should just be able to call these directly?

```suggestion
	./scripts/unix/generate_test_reports.sh
```

scripts/unix/build_project.sh
```diff
@@ -0,0 +1,15 @@
...
+# - dotnet is installed on the machine and is accessible everywhere (added to PATH)
+
+# Parse command line arguments
+BUILD_MODE=$1
+
+# Restore dependencies and build solution
+echo "Restoring and building project..."
+dotnet msbuild -property:Configuration="$BUILD_MODE" -target:Rebuild -restore || exit 1
+
+exit 0
```
Justintime50, line 15: No need to set an exit code here like in Windows, bash scripts exit with 0 if nothing else is thrown.

Justintime50 (commented):
Seems fine, I'd like to test drive it with a real release before we merge it in. We able to draft this till the next release when we can use it then review and merge?

Makefile
nwithan8: I added them 
1) to make sure they run with the correct shell (yes, I know the shebang handles this, but see 2)
2) to make sure they run in case the scripts have not been giving the execute permission (`chmod +x`).

nwithan8:
> Seems fine, I'd like to test drive it with a real release before we merge it in. We able to draft this till the next release when we can use it then review and merge?

Sure, it doesn't touch any source code, and the only foreseeable merge conflict would be changes to the Makefile. That said, I have run the process from start to finish (made a mock v4.4.0 .nupkg file), so it's been test-driven already.

nwithan8:
@Justintime50 heads up, the latest release (v4.4.0) here on GitHub (but not the NuGet release) was (accidentally) made using this new Unix process.

Justintime50:
As a reminder, I'd like to physically walk through this on our next release before we merge it in and call it "official". I'll draft this till we do that at which point, please remind me and we'll give it a whirl where I can watch it do its thing.

jchen293 (approved):

Issue #402, opened 2023-02-01
Add bool verify and bool verify_strict properties to Address object


chadwixk:
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

The docs show these 2 `bool` parameters that are not currently part of the [Address object](https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Models/API/Address.cs). Also, I've seen some other issues that show these aren't bool, but are values like `zip4` and `delivery`, so it is unclear what types and valid values they are to be.

<img width="548" alt="image" src="https://user-images.githubusercontent.com/5115308/215928284-9e95c899-3fa6-4f1f-a2c1-5611b2ed6000.png">

nwithan8:
The `Address` object you linked to is the Address model created by the "create" request. This is different than the address creation parameters, which is where  `verify` and `verify_strict` are used.

chadwixk:
We're actually using v3, so I linked to the wrong Address object.  Looks like [older version Address.cs v2](https://github.com/EasyPost/easypost-csharp/blob/v2/EasyPost/Address.cs) had it.

We create our [Address objects inline with shipment creation](https://www.easypost.com/docs/api/csharp#create-and-verify-addresses) to reduce api calls and increase performance for the user. 

And after we create the full `Shipment` object, we then make our first api call `client.CreateShipment(shipment)`.

The native api allows you to pass in these parameters in line with the address:
![image](https://user-images.githubusercontent.com/5115308/216175228-d2e55f04-1e53-4b54-8e2d-4d53dd489f8c.png)

Hence just asking for this same capability via this library as the native API supports it and it is more efficient (less api calls) and more performant for the user. Wouldn't EasyPost want to offer methods that lessen API calls and improve user experience?

Additionally, we try to stay away from using generic dictionaries and use the class objects to lessen errors, be able to find property references, etc. 

It just seems that this request only helps improve the library by lessening API calls (it would literally double the API calls to implement this way), improving end user experience (1 call is faster than 2) and enabling developers to reduce errors in code by using classes instead of string dictionaries.

nwithan8:
As part of v3, we removed the ability to construct local objects. The only way to build an `Address` object would be via an API call (such as `myClient.Address.Create`), which will deserialize the response data into an `Address` object. This ensures that data is created on the server first, echoing back a read-only copy to you locally as an object.

[...]

Issue #380, opened 2022-12-10
Add back ability to instantiate model classes


jmalatia:
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

C# is typed language!  Why would this library move backwards and now require the use of Dictionary<string, object>?

For those of us who have been using EasyPost for years (we've been using it for 8+ years)... it was a chore to get the EasyPost-csharp Official to a point it was "usable" for .net.core.  Many of us had our own forks and/or used the Official but had helpers in our applications.  We have been using [v2.5.1.2](https://www.nuget.org/packages/EasyPost-Official/2.5.1.2) for years because it just works.  

Lately we have been seeing this library getting some much needed attention!  We were excited!  Async has been added and finally C# naming guidelines are being used... so we thought it would be a good time to update.  To our surprise we can no longer create and initialize the key models (address, shipment, parcel, etc.) but we have to use Dictionary<string, object>!

Why would a C# programmer want to do:
```
Address address = new Dictionary<string, object>()
{
    { "name", "Dr. Steve Brule" },
    { "street1", "179 N Harbor Dr" },
    { "city", "Redondo Beach" },
    { "state", "CA" },
    { "zip", "90277" },
    { "country", "US" },
    { "phone", "8573875756" },
    { "email", "dr_steve_brule@gmail.com" }
}
```
When you can do this:
```
Address address = new Address()
{
    Name =  "Dr. Steve Brule",
    Street1 =  "179 N Harbor Dr",
    City =  "Redondo Beach",
    State =  "CA",
    Zip =  "90277",
    Country =  "US",
    Phone =  "8573875756",
    Email =  "dr_steve_brule@gmail.com",
}
```

The later is type checked, can easily be passed around as a typed object and one does not have to worry about mis-spelling any of the key names `{ "nammee", "" },`.  Who wants unchecked code these days when your using a typed language.

Additionally, the latter is the C# way to do it and this is a C# sharp library.

I willing to bet that very few programmers implemented EasyPost C# using the Dictionary<string, object> pattern in their calls, outside of when they were first testing things out (because that was the first sample code on the readme page).   [See nopCommerce compare commit updating from v2 to v4 which shows use of typed objects](https://github.com/nopSolutions/nopCommerce/commit/c6901df4217280d05ca4d534737b289155937ba4).

[...]

PR #409, opened 2023-02-23
[feat] Basics of new parameter objects

nwithan8:
# Description

This PR is the first part of introducing parameter objects, allowing users to use .NET objects rather than dictionaries to construct the parameters for API calls and functions.

This PR is a lot of copy-paste from the unofficial EasyPost Extensions library, with some extraneous features scaled back or removed. The basics of parameter creation are included in here, including:

- The custom `RequestParameter` attribute, which marks all parameter properties and does a lot of the heavy-lifting here
- The base `Parameters` class for all parameter objects, which contains the methods that, in coordination with the `RequestParameter` attributes, validates and processes a parameter object into a dictionary to be used downstream.
- Basic parameter objects for most API calls, grouped by API version and related service
- Basic unit test to confirm that parameter object -> dictionary serialization is working as expected.

There are a number of things not yet addressed in this PR, including:
- The required/optional levels for all parameters (currently, all parameters are marked as optional)
- The use of the parameters. All functions still currently expect dictionaries; in the future, we will provide override methods that allow users to directly pass in the new parameters objects.

There are also minor tweaks included in this PR to utilities, mainly the re-organization of "internal-only" utilities into a new `Internal` namespace. These utilities were marked with the "internal" access modifier previously, meaning there were already inaccessible to end-users. These changes merely re-organize the file structure for readability purposes, and should not introduce any breaking changes to end users. Utilities that are meant to be accessed (namely, the cryptographic services and rate filtering) are still accessible and have not been moved.

See reviewer comments for more details.

Related to #380 and #402 

# Testing

- All existing unit tests work (none of their functionality was adjusted)
- New unit test for basic parameter -> dictionary conversion
- New unit test for sub-parameter -> dictionary conversion

# Pull Request Type


.editorconfig
```diff
@@ -568,3 +568,5 @@ dotnet_diagnostic.SA1402.severity = none
 dotnet_diagnostic.SA1401.severity = none
 ### Private fields should not start with underscore
 dotnet_diagnostic.SA1309.severity = none
+### Use "gets" for properties docstrings
+dotnet_diagnostic.SA1623.severity = none
```
nwithan8, line 572: ### Author's Note

With all the properties for the parameter objects being added, this was going to get really annoying if it wasn't disabled.

EasyPost/Exceptions/General/MissingParameterError.cs
```diff
@@ -12,5 +13,14 @@ internal MissingParameterError(string parameterName)
             : base(string.Format(CultureInfo.InvariantCulture, Constants.ErrorMessages.MissingRequiredParameter, parameterName))
         {
         }
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="MissingParameterError" /> class.
+        /// </summary>
+        /// <param name="property">The <see cref="PropertyInfo"/> of the missing property.</param>
+        internal MissingParameterError(PropertyInfo property)
```
nwithan8, line 21: ### Author's Note

Using the property to determine the name, rather than a static string, is actually a cleaner, less-error-prone way of calculating the error message. This was ported from the Extensions library.

EasyPost/Parameters/V2/Base.cs
```diff
@@ -0,0 +1,93 @@
+using EasyPost._base;
+using EasyPost.Utilities.Annotations;
+
+namespace EasyPost.Parameters.V2
+{
+    /// <summary>
+    ///     Base class for all parameters used in requests to <see cref="ApiVersion.V2"/> endpoints.
+    /// </summary>
+#pragma warning disable SA1649
```
nwithan8, line 9: ### Author's Note

This is a complaint that the class name does not match the file name. This is deliberate.

EasyPost/Utilities/Internal/Extensions/Dictionaries.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public static class Dictionaries
+    {
+        /// <summary>
+        ///     Converts a <see cref="Dictionary{TKey,TValue}"/> of string, object? (nullable) key-value pairs to a dictionary of string, object key-value pairs
+        ///     by omitting key-value pairs with null values.
+        /// </summary>
+        /// <param name="dictionary">A <see cref="Dictionary{TKey,TValue}"/> to convert.</param>
+        /// <returns>A <see cref="Dictionary{TKey,TValue}"/> of string, object pairs.</returns>
+        public static Dictionary<string, object> ConvertToStringNonNullableObjectDictionary(Dictionary<string, object?> dictionary)
```
nwithan8, line 13: ### Author's Note

This was ported from the Extensions library. It's needed because our functions expect `Dictionary<string, object>`-type (non-nullable) dictionaries, but the parameters serialization creates `Dictionary<string, object?>`-type (nullable) dictionaries. This function converts a nullable dictionary to a non-nullable dictionary by simply removing all key-pairs with null values.

EasyPost/Utilities/Internal/Extensions/Enumerables.cs
```diff
@@ -0,0 +1,22 @@
...
+    public static class Enumerables
+    {
+        /// <summary>Foreach with an index.</summary>
+        /// <param name="ie">IEnumerable element.</param>
+        /// <param name="action">
+        ///     Action to do for each loop.
+        ///     Action receives a (int, T) tuple containing the index and the current element in the loop as a parameter
+        /// </param>
+        /// <typeparam name="T">Type of input.</typeparam>
+        public static void Each<T>(this IEnumerable<T> ie, Action<int, T> action)
```
nwithan8, line 15: ### Author's Note

This extension method is used to ease the process of conducting an action on each element in a list. Specifically, this is used by the `toDictionary` method override for making a tracker list.

EasyPost/_base/ApiVersion.cs
```diff
@@ -7,6 +7,7 @@ namespace EasyPost._base
     public class ApiVersion : ValueEnum
     {
         public static readonly ApiVersion Beta = new(2, "beta");
+        public static readonly ApiVersion V2 = new(1, "v2");
```
nwithan8, line 10: ### Author's Note

This was added initially purely for documentation purposes, so V2 parameters can reference the V2 API version in their docstrings.

However, this also offers an alternative/future-prep for eventual API versioning. Users can select Current or V2, they both (currently) do the same thing.

EasyPost/Parameters/V2/CarrierAccounts.cs
```diff
@@ -0,0 +1,209 @@
...
+            [RequestParameter(Necessity.Optional, "carrier_account", "type")]
+            public string? Type { get; set; }
+
+            #endregion
+        }
+
+        /// <summary>
+        ///     Parameters for FedEx <see cref="EasyPost.Services.CarrierAccountService.Create"/> creation API calls.
+        /// </summary>
+        public sealed class CreateFedEx : CreateRequestParameters
```
nwithan8, line 38: ### Author's Note

We can get rid of these FedEx- and UPS-specific parameter sets if we don't want to worry about that for now.

EasyPost/Parameters/V2/Base.cs
Justintime50: I'd argue that the linter is correct and that class names should always match file names. Why can't we call this `BaseRequestParameters` or something similar so you keep your `Base` as a part of it?

EasyPost/Parameters/V2/Base.cs
```diff
@@ -0,0 +1,93 @@
+using EasyPost._base;
```
Justintime50, line 1: Why are there ultimately two of these same files? One in Beta and one in V2? Since an API version will always be tied to the lib, we should probably flatten this structure and remove the `V2` directory completely. Whatever the lib version is will always use the current API version.

EasyPost/Parameters/V2/Addresses.cs
```diff
@@ -0,0 +1,79 @@
+using EasyPost.Utilities.Annotations;
```
Justintime50, line 1: I thought we were going to break this up into smaller PRs? It's difficult to review individual properties of so many resources in a single go. I'd suggest keeping `Address` here to prove the concept with the guts/internals of the logic, then breaking the rest of the resources into two separate PRs that build on top of this one once merged.

Justintime50 (changes requested):
I'd suggest we add an example of how to use this in the README in this PR. I'd also suggest we maybe label it as beta for now and in a month or two after getting more feedback/usage we can remove the label if things are going well.

EasyPost.Tests/ParametersTests/ParametersTest.cs
```diff
@@ -0,0 +1,101 @@
...
+            const string street = "388 Townsend St";
+
+            var parameters = new Parameters.V2.Addresses.Create
+            {
+                Street1 = street,
+                Street2 = "Apt 20",
+                City = "San Francisco",
+                State = "CA",
+                Zip = "94107",
+                Country = "US",
```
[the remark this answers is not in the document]
nwithan8: The dummy data here doesn't really matter, an API call is not being sent.

EasyPost.Tests/ParametersTests/ParametersTest.cs
```diff
@@ -0,0 +1,101 @@
...
+            // Check that the dictionary contains 2 elements, including an "shipment" key
+            Assert.Equal(2, dictionary.Count); // "shipment" + default "carbon_offset" top-level key
+            Assert.True(dictionary.ContainsKey("shipment"));
+
+            // Check that the value of the "shipment" key is a dictionary with 2 elements
+            var shipmentData = dictionary["shipment"] as Dictionary<string, object>;
+            Assert.NotNull(shipmentData);
+            Assert.Equal(3, shipmentData.Count); // the 1 element we set + default "is_return" boolean and "insurance" 0 value
+
+            // TODO: This test fails because address data will get embedded as "to_address" -> "address" -> "street1" instead of "to_address" -> "street1"
```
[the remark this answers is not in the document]
nwithan8: Yes, this is something we need to discuss as a team

EasyPost.Tests/ParametersTests/ParametersTest.cs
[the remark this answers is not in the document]
Justintime50: True, though the point of fixtures is you don't need to define your own test data

EasyPost/Parameters/Beta/Base.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        internal UpdateRequestParameters()
+        {
+        }
+    }
+
+    /// <summary>
+    ///     Base class for all parameters used in "list" requests to <see cref="ApiVersion.Beta"/> endpoints.
+    /// </summary>
+    public abstract class AllRequestParameters : RequestParameters
```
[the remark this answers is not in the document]
nwithan8: If that's the case, "all" is the only parameter base class that has properties. We can get rid of the "all", "create" and "update" base classes and simply re-implement the whole thing on a per-case basis. Was just trying to cut down on redundancy

EasyPost/Parameters/V2/Base.cs
nwithan8: I was unaware that was the official approach we had decided on for the eventual API versioning. Will do

EasyPost/Parameters/V2/Addresses.cs
nwithan8: Fair enough, was a lot of copy-pasta. Will remove all classes except Address Create and Shipment Create, used in the current unit tests.

EasyPost.Tests/ParametersTests/ParametersTest.cs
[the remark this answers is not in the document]
nwithan8: Extracting the fixture data from a JSON dictionary into these primitives is more work that it's worth for this specific unit test.

EasyPost.Tests/ParametersTests/ParametersTest.cs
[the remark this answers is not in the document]
nwithan8: This has been rectified.

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            Assert.True(addressData.ContainsKey("residential"));
+
+            // Check that the "street1" key was set with the correct value
+            Assert.NotNull(addressData["street1"]);
+            Assert.Equal(street, addressData["street1"]);
+        }
+
+        [Fact]
+        [Testing.Parameters]
+        public void TestParametersToDictionaryWithSubDictionary()
```
Justintime50, line 63: I'd really like us to be annotating our tests with docstrings to define what the expected outcome of the test is. These tests are pretty long, it would be good if there was a simple single human readable line saying "this test does this thing and we expect it to do xyz"

AKA: The name of the test doesn't necessarily match up with the contents of this test. I get two different stories by looking at each.

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            // No errors here confirm that the dictionary was serialized to the expected schema
+            var dictionary = parameters.ToDictionary();
+            var shipmentData = dictionary["shipment"] as Dictionary<string, object>;
+            var addressData = shipmentData["to_address"] as Dictionary<string, object>;
+
+            // The value of "street1" should be the value of "streetA" via the Address object
+            Assert.Equal(streetA, addressData["street1"]);
+
+            // Users can also pass in an Address.Create parameter object as the "ToAddress" parameter
```
Justintime50, line 93: Plural right?

I don't know that I love this interface of plural? You'd only ever be creating one of anything. What happened to the `Params` name? Weren't we going to do something like `AddressParams` or `AddressCreateParams`?

```suggestion
            // Users can also pass in an Addresses.Create parameter object as the "ToAddress" parameter
```

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            Assert.Equal(streetB, addressData["street1"]);
+        }
+
+        [Fact]
+        [Testing.Logic]
+        public void TestTopLevelVersusNestedParameters()
+        {
+            // This test proves that we can reuse the Address.Create parameter object,
+            // with its serialization logic adapting to whether it is
+            // a top-level parameter object or a nested parameter object within another parameter object.
```
Justintime50, line 115: This is exactly the kind of stuff I'd hope each test have in a docstring, can we move it to a docstring instead of in the function?

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            // Path to "street1" should be dictionary["shipment"]["to_address"]["street1"]
+            var shipmentData = dictionary["shipment"] as Dictionary<string, object>;
+            var toAddressData = shipmentData["to_address"] as Dictionary<string, object>;
+            Assert.Equal(street, toAddressData["street1"]);
+
+            // Notice how the paths to "street1" are different depending on whether the Address.Create parameter object
+            // The schema for an address creation API call contains all address data wrapped in an "address" key (excluding irrelevant "verify" and "verify_strict" keys)
+            // The schema for a shipment creation API call does not contain this "address" key, with all address data instead wrapped inside a "to_address" key
+            // Behind the scenes, the Address.Create parameter object is the same, but the serialization path for each property (parameter) adapts to the context in which it is used
```
Justintime50, line 147: This should also be in the docstring of the test, not buried at the bottom of a function. I'd expect this to be a paragraph after the intro line to the test in the docstring.

EasyPost/BetaFeatures/Parameters/Parameters.cs
```diff
@@ -0,0 +1,186 @@
...
+            {
+                // Don't need to go down
+                case 0:
+                    return dictionary;
+
+                // Last key left
+                case 1:
+                    string soloKey = keys[0];
+                    if (dictionary.ContainsKey(soloKey))
+                        // Key-value pair already exists in dictionary (likely because of override parameters)
```
Justintime50, line 156: Yeah I definitely don't agree with this `Single-line comment should be preceded by blank line` github actions checker rule. Can we please turn these off? They are incredibly noisy and we haven't come back to fix them. They are doing more harm during code reviews than good since we are ignoring them all anyways.

README.md
```diff
@@ -139,6 +139,46 @@ Any generated local resource will have stored internally the same client used to
...
+
+// You can add additional parameters as needed outside of the constructor
+addressCreateParameters.Company = "My Company";
+
+// Then convert the object to a dictionary
+// This step will validate the data and throw an exception if there are any errors (i.e. missing required parameters)
+var addressCreateDictionary = addressCreateParameters.ToDictionary();
+
+// Pass the dictionary into the address creation method as normal
+var address = await myClient.Address.Create(addressCreateDictionary);
```
Justintime50, line 168: Did we intend to only allow them to pass in the dictionary? I thought we were going to look into allowing them to pass either that or the object? Or did I misunderstand?

README.md
```diff
@@ -139,6 +139,46 @@ Any generated local resource will have stored internally the same client used to
...
 
+## Parameters (BETA):
+
+Most functions in this library accept a `Dictionary<string, object>` as their sole parameter, which is ultimately used as the body of the HTTP request against EasyPost's API. If you instead would like to use .NET objects to construct API call parameters, you can use the various `Parameters` classes (currently in beta).
+
+For example, to create an address:
+
+```csharp
+// Use an object constructor to create the address creation parameters
+var addressCreateParameters = new EasyPost.BetaFeatures.Parameters.Address.Create {
```
Justintime50, line 150: Question: This interface doesn't seem to match the one in the tests?

check_coverage.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-THRESHOLD=90
+THRESHOLD=84
```
Justintime50, line 3: 6% is too much of a drop, what other tests can we add to keep a drop in coverage closer to only a couple percentage points?

The point of enforcing coverage is so we can't arbitrarily cover new features and that we ensure that anything new gets full coverage. There will be times when a percent or two drop is acceptable, we can't test every branch - 6% to me means we have large chunks of logic that are now going untested.

Justintime50 (changes requested):
Conceptually this seems fine. Seems we need to tighten down on the interface (or at least I need my understanding clarified) and we need some better test coverage.

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
nwithan8: The namespace to the object (the full name of the class) is `EasyPost.BetaFeatures.Parameters.Addresses.Create`, so I figured it was pretty self-explanatory that these are parameters.

As for the plural, it's more for categorization. There would be `Addresses.All`, `Addresses.Retrieve`, `Addresses.Create`, etc.

README.md
nwithan8: Yes, we will, in a later PR. I wanted to get the dictionary serialization working and signed off before we started overloading methods.

check_coverage.sh
nwithan8: Not sure why it failed the first time and required a drop. Line coverage is actually at 92 percent now, bumping

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
Justintime50: Ah, with them missing the full namespace it was confusing. This is probably fine.

EasyPost.Tests/BetaFeaturesTests/ParametersTests/ParametersTest.cs
```diff
@@ -102,34 +102,43 @@ public void TestParametersToDictionaryWithSubDictionary()
...
+        /// or a nested parameter object within another parameter object.
+        ///
+        /// Notice how the paths to "street1" are different depending on whether the Addresses.Create parameter object.
+        ///
+        /// The schema for an address creation API call contains all address data wrapped in an "address" key (excluding irrelevant "verify" and "verify_strict" keys).
+        ///
+        /// The schema for a shipment creation API call does not contain this "address" key, with all address data instead wrapped inside a "to_address" key.
+        ///
+        /// Behind the scenes, the Addresses.Create parameter object is the same, but the serialization path for each property (parameter) adapts to the context in which it is used
+        /// </summary>
```
Justintime50, line 121: Looove this. A single place where I can get all the details I need to know about this test.

Justintime50 (approved):
Seems fine, let's keep refining as we go.

EasyPost/BetaFeatures/Parameters/Shipments/Create.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    /// <summary>
+    ///     Parameters for <see cref="EasyPost.Services.ShipmentService.Create"/> API calls.
+    /// </summary>
+    public class Create : Parameters, IShipmentParameter
+    {
+        #region Request Parameters
+
+        [TopLevelRequestParameter(Necessity.Optional, "carbon_offset")]
+        public bool AddCarbonOffset { get; set; }
```
jchen293, line 14:
```suggestion
        public bool? AddCarbonOffset { get; set; }
```

Missing `?` on some of these optional properties. This applies to other places.

README.md
```diff
@@ -139,6 +139,46 @@ Any generated local resource will have stored internally the same client used to
...
+    Zip = "94105",
+    Country = "US",
+    Phone = "415-123-4567"
+};
+
+// You can add additional parameters as needed outside of the constructor
+addressCreateParameters.Company = "My Company";
+
+// Then convert the object to a dictionary
+// This step will validate the data and throw an exception if there are any errors (i.e. missing required parameters)
```
jchen293, line 164: Correct me if I am wrong. Currently, all the parameters are optional, so I don't think an exception will be thrown in `ToDictionary` method.

jchen293 (approved):
lgtm, left two comments

EasyPost/BetaFeatures/Parameters/Shipments/Create.cs
nwithan8: Bool doesn't need to be explicitly nullable. Will default to "false" (or whatever default value we set it to) and always be included.

Or we could make it nullable, in which case it won't be included in the JSON body unless explicitly set.

Thoughts @Justintime50 ?

README.md
nwithan8: Yes, all parameters are optional right now, to keep this PR simple. In the next PRs, we'll actually start making some required.

EasyPost/BetaFeatures/Parameters/Shipments/Create.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        [TopLevelRequestParameter(Necessity.Optional, "shipment", "carrier_accounts")]
+        public List<EasyPost.Models.API.CarrierAccount>? CarrierAccounts { get; set; }
+
+        [TopLevelRequestParameter(Necessity.Optional, "shipment", "customs_info")]
+        public EasyPost.Models.API.CustomsInfo? CustomsInfo { get; set; }
+
+        [TopLevelRequestParameter(Necessity.Optional, "shipment", "insurance")]
+        public double Insurance { get; set; }
```
jchen293, line 23:
```suggestion
        public double? Insurance { get; set; }
```

This is optional, let's make it nullable

jchen293 (approved):

EasyPost/BetaFeatures/Parameters/Shipments/Create.cs
Justintime50: Why is this called `AddCarbonOffset` instead of just `CarbonOffset`?

We determined we'd always send this when we first implemented it (meaning it's always true or false), let's not change directions now.

Justintime50 (commented):
And `VerifyStrict` changes please

Justintime50 (approved):

PR #410, opened 2023-03-01
[chore]: Remove excessive CI warnings

nwithan8:
# Description

- Update style guide to disable several excessive warnings, lower severity
- Adopt a suggestion about sealing attributes
- Finally bump F# to 7.0 for CI so it will stop auto-updating all the time.

# Testing

- N/A

# Pull Request Type


Justintime50:
This doesn't seem to actually address the issue, unless it doesn't work till it's committed to master?

Screenshot 2023-03-01 at 5 44 23 PM

jchen293 (approved):

nwithan8:
All warnings that are going to be addressed now have been addressed.

The warning about considering ConfigureAwait is (for once) actually a good suggestion, and will be explored in a separate PR.
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007

nwithan8:
This reopens the confusing discussion around static analysis in our codebase. The code coverage step in CI, during the build-run process, is also triggering warnings (which GitHub is extracting from terminal logs and presenting in a GUI) about violations of our code style (rules set in the `.editorconfig` file). By suppressing/editing the rules in `.editorconfig`, these warnings go away.

At the same time, Roslyn is being run on the code: (configured to run every time the code compiles, I can't confirm if that's working or not; explicitly running via a CI step, that is running). The explicit CI step for Roslyn static analysis (static analysis being more "you shouldn't nest classes" versus code style "you should use two spaces here instead of one"-things that don't actually affect compilation/running) does not seem to respect the rules in `.editorconfig`. As far as I can tell, it should, or at least was going to be made to do so a few years ago. Regardless, it doesn't, and so all the notices of Roslyn analysis violations (which there are a lot because we have it configured for the strictest level) are collected by GitHub.

The Roslyn violations are available [here](url) in GitHub's "Code Scanning" section. All notices from this run as considered "notes", the lowest level of severity. There are no legitimate errors and no security-related concerns. I would suggest that we simply ignore them. Occasionally some Roslyn notices are injected into this PR review page by the GitHub bot; I would suspect those are things with higher than "note" severity, which should be considered something that needs to be addressed.

nwithan8:
Documenting this here and will document it internally:

When it comes to .NET, there are two separate analyses that we conduct on our code: code styling and static analysis 

Both analyses follow their own customizable rulesets, defined in an `.editorconfig` file.

Code styling (rules with `IDE` prefixes) pertains to the visual style of the source code. An example would be "should there be a blank line between a comment and a line of code?" These are cosmetic decisions, which impact the readability of the code from a developer perspective. However, these visual decisions have little to no impact on the compiled output. The compiler does not care about line breaks, comment length, whether a docstrings has a period, etc, and these have no impact on the end-user's ability to execute the code.

Static analysis (rules with `CA` prefixes) pertain to the functionality or operability of the code. An example would be "you should not iterate over the same list twice." These are decisions that impact the performance of the final compiled output, including optimizations and security vulnerabilities.

In our CI, we conduct these analyses separately.

We have an explicit step for linting, which takes into account the `IDE` rules in the `.editorconfig` file and parses the source code for violations.

We have an explicit step for static analysis, which (due to an issue with the Roslyn static analyzer and the GitHub Action) does not take into account the `CA` rules in the `.editorconfig` file (instead, it uses the default strict rules), and performs static analysis on the source code, looking for security vulnerabilities and performance issues. (We will soon introduce a Makefile step to hopefully patch this issue and migrate to using this step instead of the GitHub Action). Results of the static analysis are uploaded to GitHub, which store the results in a GUI format; a GitHub bot comments directly on a PR with high-priority results from the static analysis.

Static analyzers like Roslyn do not care about code styling rules (`IDE` rules), since these become irrelevant once the source code is compiled.

nwithan8:
GitHub Actions appears to be stalled at the moment.

Justintime50 (approved):

Justintime50:
@nwithan8 you'll need to push a trivial commit to get this working again.

Justintime50 (approved):

PR #412, opened 2023-03-07
[feat]: Parameter objects, part 2

nwithan8:
# Description

Now that we've got the framework of the parameters out of the way, it's time to start adding in all the parameter sets and overloaded methods that accept them.

This PR adds the parameters sets used in the `Create`, `All` and `Update` functions in our library (which is a significant portion of the total available API-calling functions).

This PR also adds overloaded `Create` functions that accept a Create parameters object, rather than a dictionary.

This PR also adds unit tests to confirm these overloaded functions are working as expected, and defends a design decision, due to an unfortunate prior design decision, that mildly shackles how users can (and can't) use these new Parameters objects.

# Testing

- All unit tests pass (nothing should interfere with the existing ones, and the new ones pass)
- All Parameters sets have been marked to be excluded from code coverage. Otherwise, the simple getters/setters for each property dramatically lower the coverage.
  - This was initially attempted using a `.runsettings` file, but the coverage collector used by the CLI for whatever reason seemed to ignore the ruleset in the file. Instead, the classes were marked with the `[ExcludeFromCodeCoverage]` attribute, which works to exclude the whole classes from coverage reports.

# Pull Request Type


EasyPost/BetaFeatures/Parameters/Refunds/Create.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace EasyPost.BetaFeatures.Parameters.Refunds
+{
+    public class Create : BaseParameters, IRefundParameter
+    {
+        #region Request Parameters
+
+        [TopLevelRequestParameter(Necessity.Required, "refund", "carrier")]
+        public string? Carrier { get; set; }
+
+        [TopLevelRequestParameter(Necessity.Required, "refund", "tracking_codes")] // yes, the param name is plural when it's really just one code
```
Justintime50, line 12: Issue: This comment is invalid, you can pass a list of tracking codes (plural) to create refunds in bulk. This is why we recently fixed the refund create tests in our libs to reflect this realization.

Easy fix, please remove the comment here.

EasyPost/BetaFeatures/Parameters/Shipments/Create.cs
```diff
@@ -11,39 +11,78 @@ public class Create : BaseParameters, IShipmentParameter
         #region Request Parameters
 
         [TopLevelRequestParameter(Necessity.Optional, "carbon_offset")]
-
-        // non-nullable, will always be included (default: false)
-        public bool CarbonOffset { get; set; }
+        // "carbon_offset" is not included when shipment creation parameters are used in a non-shipment creation request.
```
Justintime50, line 14: Issue: Please fix this GitHub check in this file

EasyPost/Models/API/CustomsInfo.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System.Collections.Generic;
 using EasyPost._base;
+using EasyPost.BetaFeatures.Parameters;
 using Newtonsoft.Json;
 
 namespace EasyPost.Models.API
 {
-    public class CustomsInfo : EasyPostObject
+    public class CustomsInfo : EasyPostObject, ICustomsInfoParameter
```
Justintime50, line 8: Please fix/ignore these GitHub checks

Justintime50 (changes requested):

EasyPost/BetaFeatures/Parameters/EndShippers/Create.cs
```diff
@@ -2,14 +2,13 @@
...
 namespace EasyPost.BetaFeatures.Parameters.EndShippers
 {
+    /// <summary>
+    ///     Parameters for <see cref="EasyPost.Services.EndShipperService.Create(Create)"/> API calls.
+    /// </summary>
     public class Create : BaseParameters, IEndShipperParameter
     {
         #region Request Parameters
 
-        [TopLevelRequestParameter(Necessity.Optional, "address", "carrier_facility")]
```
Justintime50: I'm not seeing these same changes made to updating an endshipper

EasyPost/BetaFeatures/Parameters/Trackers/All.cs
```diff
@@ -1,11 +1,13 @@
+using System.Diagnostics.CodeAnalysis;
 using EasyPost.Utilities.Internal.Annotations;
 
 namespace EasyPost.BetaFeatures.Parameters.Trackers
 {
     /// <summary>
     ///     Parameters for <see cref="EasyPost.Services.TrackerService.All"/> API calls.
     /// </summary>
-    public class All : BaseParameters
+    [ExcludeFromCodeCoverage]
```
Justintime50, line 9: Question: I'd prefer if we could exclude these all via config instead of in-code via this `ExcludeFromCodeCoverage`, is that possible?

nwithan8: See comment in PR description. I spent over an hour trying to get it to work, but no dice...

Justintime50 (approved):
Seems fine, as long as this is labeled as beta for now.

PR #414, opened 2023-03-13
[feat] Parameters objects, part 3

nwithan8:
# Description

- Adds parameter sets and associated overloads for all API-calling methods in the library
- Fixes bug in parameter dictionary generation that wasn't accounting for non-public properties
  - Adds unit test to verify this works, prevent regression
- Fixes bug in parameter dictionary generation that wasn't serializing lists properly

# Testing

# Pull Request Type


nwithan8:
TODO: Now that all the parameter sets have been created, we need to determine which parameters are required versus optional for each API call.

nwithan8:
> TODO: Now that all the parameter sets have been created, we need to determine which parameters are required versus optional for each API call.

We have decided to, for the time being, keep all parameters as optional. This could allow users to send malformed/incomplete request bodies in their API requests, but this is also possible under the current dictionary system, and therefore does not introduce a regression.

EasyPost.Tests/UtilitiesTests/AttributesTest.cs
```diff
@@ -18,29 +85,61 @@ public class AttributesTest
...
             Assert.True(methods.Any());
         }
 
+        [Fact]
+        [Testing.Logic]
+        public void TestCanGetNonPublicAndStaticMethodsWithAttributes()
+        {
+            // via type
+            IEnumerable<MethodInfo> methods = BaseCustomAttribute.GetMethodsWithAttribute<ExampleAttribute>(typeof(ExampleClass));
+            // we should have 8 methods with the attribute
```
Justintime50, line 103: Nit: The assertion says the same as this comment. This comment isn't helpful (in it's current form) and should either be removed or clarified as to **why** we expect 8

EasyPost.Tests/UtilitiesTests/AttributesTest.cs
```diff
@@ -18,29 +85,61 @@ public class AttributesTest
...
+            Assert.True(methods.Count() == 8);
+
+            // via instance
+            ExampleClass exampleClass = new();
+            methods = BaseCustomAttribute.GetMethodsWithAttribute<ExampleAttribute>(exampleClass);
+            // we should have 8 methods with the attribute
```
Justintime50, line 109: Nit: The assertion says the same as this comment. This comment isn't helpful (in it's current form) and should either be removed or clarified as to **why** we expect 8

EasyPost.Tests/UtilitiesTests/AttributesTest.cs
```diff
@@ -18,29 +85,61 @@ public class AttributesTest
...
             Assert.True(properties.Any());
         }
 
+        [Fact]
+        [Testing.Logic]
+        public void TestCanGetNonPublicAndStaticPropertiesWithAttributes()
+        {
+            // via type
+            IEnumerable<PropertyInfo> properties = BaseCustomAttribute.GetPropertiesWithAttribute<ExampleAttribute>(typeof(ExampleClass));
+            // we should have 8 properties with the attribute
```
Justintime50, line 133: Nit: The assertion says the same as this comment. This comment isn't helpful (in it's current form) and should either be removed or clarified as to **why** we expect 8

EasyPost.Tests/UtilitiesTests/AttributesTest.cs
```diff
@@ -18,29 +85,61 @@ public class AttributesTest
...
+            Assert.True(properties.Count() == 8);
+
+            // via instance
+            ExampleClass exampleClass = new();
+            properties = BaseCustomAttribute.GetPropertiesWithAttribute<ExampleAttribute>(exampleClass);
+            // we should have 8 properties with the attribute
```
Justintime50, line 139: Nit: The assertion says the same as this comment. This comment isn't helpful (in it's current form) and should either be removed or clarified as to **why** we expect 8

EasyPost.Tests/UtilitiesTests/AttributesTest.cs
```diff
@@ -18,29 +85,61 @@ public class AttributesTest
...
-            Shipment shipment = new();
-            properties = BaseCustomAttribute.GetPropertiesWithAttribute<JsonPropertyAttribute>(shipment);
+            ExampleClass exampleClass = new();
+            properties = BaseCustomAttribute.GetPropertiesWithAttribute<ExampleAttribute>(exampleClass);
             Assert.True(properties.Any());
         }
 
+        [Fact]
+        [Testing.Logic]
+        public void TestCanGetNonPublicAndStaticPropertiesWithAttributes()
```
Justintime50, line 129: suggestion: Please annotate tests. This looks identical on the surface to the previous test. Why/how is it different

EasyPost/Models/API/Batch.cs
```diff
@@ -123,20 +195,42 @@ public async Task<Batch> GenerateScanForm(string fileFormat = "pdf") // TODO: Re
...
+            }
+
+            await Update<Batch>(Method.Post, $"batches/{Id}/remove_shipments", parameters.ToDictionary());
+            return this;
+        }
+
         /// <summary>
         ///     Remove shipments to this batch.
         /// </summary>
         /// <param name="shipmentsToAdd">List of Shipment objects to be removed.</param>
```
Justintime50, line 228: Issue: fix this github concern

EasyPost/Models/API/Batch.cs
```diff
@@ -123,20 +195,42 @@ public async Task<Batch> GenerateScanForm(string fileFormat = "pdf") // TODO: Re
...
         /// <returns>The updated Batch.</returns>
         [CrudOperations.Update]
-        public async Task<Batch> RemoveShipments(List<Shipment> shipmentsToAdd)
+        public async Task<Batch> RemoveShipments(List<Shipment> shipmentsToRemove)
```
Justintime50, line 231: Issue: fix this github concern

EasyPost/Services/TrackerService.cs
```diff
@@ -48,9 +54,11 @@ public async Task<Tracker> Create(BetaFeatures.Parameters.Trackers.Create parame
         /// <param name="parameters">A dictionary of tracking codes and carriers.</param>
         /// <returns>True if successful, False otherwise.</returns>
         [CrudOperations.Create]
+        [Obsolete("This method is deprecated. Please use TrackerService.Create() instead. This method will be removed in a future version.", false)]
         public async Task CreateList(Dictionary<string, object> parameters)
         {
             parameters = parameters.Wrap("trackers");
+            // This endpoint does not return a response so we return the request was successful
```
Justintime50, line 61: Question: This comment doesn't make sense? We return void here

Makefile
```diff
@@ -31,7 +31,7 @@ coverage:
 
 ## coverage-check - Check if the coverage is above the minimum threshold
 coverage-check:
-	./check_coverage.sh 90
+	./check_coverage.sh 86
```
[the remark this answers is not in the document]
nwithan8: At present, there's no _easy_ way to generate the Parameters objects from our fixture data. We can make unit tests for each overload, but the parameters for each function will need to be made manually and won't automatically stay in sync if our fixture data JSON file ever changes.

Let me know how we want to proceed.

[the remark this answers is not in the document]
nwithan8: UPDATE: Working on copying all the existing service and model tests into a new set of unit tests that will effectively do the same thing, but this time with the new Parameters objects. This will confirm that the overloaded methods work and that the Parameters objects are serializing to JSON correctly (creating the same exact payload as the existing dictionary methods).

I have completed all the service tests, working on models right now. With these new unit tests (and the fixtures needed to make them work), almost all lines for the Parameters objects are being touched, which in the end should result in potentially a net-gain in coverage (with the services done, coverage climbed back from 86 to 88.5 already)

nwithan8:
This PR is going to be split into separate, smaller PRs, starting with #416

PR #415, opened 2023-03-15
[chore] Fix bad anti-pattern

nwithan8:
# Description

This PR removes an inconsistent anti-pattern of storing a copy of the Client used to retrieve an XCollection inside the collection itself.

This was implemented earlier to cater to the initial concept of instance-based functions in the library. While that pattern still exists (for now), this was applied inconsistently. It's not a bug, but we should remove it to avoid unintended consequences.

Upcoming work for getting the next page of a paginated collection will most likely reintroduce this, but holistically.

This is unlikely to introduce a breaking change for end-users (unless they were stepping outside of expected workflow patterns)

# Testing

- No changes to unit tests or functionality

# Pull Request Type


Justintime50 (approved):
Should be fine since you can't call an action against a collection so the client shouldn't be needed for these

jchen293 (approved):

PR #416, opened 2023-03-15
[feat] Parameter sets, part 3, take 2

nwithan8:
# Description

- Add remaining parameter sets for classes A-H (Addresses - Events)
- Add overload methods using Parameter objects for classes A-H
- Add unit tests, fixtures needed, cassettes for overloaded methods (services)
- Fix bug in parameter dictionary generation that wasn't accounting for non-public properties
  - Adds unit test to verify this works, prevent regression
- Fix bug in parameter dictionary generation that wasn't serializing lists properly
- Minor cleanup, bug fixes of existing parameter set classes as needed

# Testing

- New unit tests for overloaded functions for parameter sets A-H (Addresses - Events)
  - New cassettes

# Pull Request Type


nwithan8:
89.99 doesn't round up to 90. I feel like I'm in college again begging my professor for a grade bump.
image

EasyPost/Utilities/Internal/Extensions/Dictionaries.cs
```diff
@@ -1,5 +1,7 @@
+using System;
```
Justintime50, line 1: Thought: The amount of utility code this repo needs is astronomically too large for what the library actually does/is. I'd like us over time to be trimming this down, not adding more to it.

How can we do more with less? What patterns can we identify and refactor (without affecting user expectations and behavior under the hood) to remove dead or duplicated code? Something to think and improve on.

Justintime50 (approved):
Nice work on the tests, thanks for doing that.

EasyPost/Utilities/Internal/Extensions/Dictionaries.cs
nwithan8: I can start in the next PR by ripping out the three overloads for this method that I added that aren't being used :)

jchen293 (approved):

PR #420, opened 2023-03-17
[feat] POC: Functionality to get the next page of a paginated collection

nwithan8:
# Description

This is a proof-of-concept for a new `GetNextPage` function for each of our services (e.g. `Address`) (as needed).

Users pass in the current paginated collection object (e.g. `AddressCollection`), and it will craft the parameters and execute the request to retrieve the next page of results for a paginated collection (or throw an exception if there are no more pages to retrieve).

The underlying base functions for handling this utilize interfaces to abstract away the type restrictions. This means that the base functions will handle any type of paginated collection thrown at them, with the upstream function being more specific and locked down to a specific type (ex. you can't pass a `ShipmentCollection` object into the `Address.GetNextPage()` function).

The base function to calculate parameters for the API call is abstract, and must be implemented in each `XCollection` class to build specific parameters for that class.

For collection types that do not have a pagination capabilities (i.e. EndShipper), the `GetNextPage` function will not be implemented in the service class. The parameter set creation function needs to be implemented due to inheritance, but can simply throw a NotImplementedException, since it will never be called. 

This PR:
- Introduces the base functionality to craft parameters and execute API calls to get the next page of a paginated collection
- Adds a `GetNextPage` function to the base service unit test + cassette

# Testing

- Added unit tests, cassettes for new feature, all pass

# Pull Request Type

### DIFF
--- a/EasyPost.Tests/ServicesTests/AddressServiceTest.cs
+++ b/EasyPost.Tests/ServicesTests/AddressServiceTest.cs
@@ -124,6 +124,21 @@ namespace EasyPost.Tests.ServicesTests
         [Fact]
         [CrudOperations.Read]
         [Testing.Function]
+        public async Task TestGetNextPage()
+        {
+            UseVCR("get_next_page");
+
+            AddressCollection addressCollection = await Client.Address.All(new Dictionary<string, object> { { "page_size", Fixtures.PageSize } });
+
+            AddressCollection nextPageAddressCollection = await Client.Address.GetNextPage(addressCollection);
+
+            // If the first ID in the next page is the same as the first ID in the current page, then we didn't get the next page
+            Assert.NotEqual(addressCollection.Addresses[0].Id, nextPageAddressCollection.Addresses[0].Id);
+        }
+
+        [Fact]
+        [CrudOperations.Read]
+        [Testing.Function]
         public async Task TestRetrieve()
         {
             UseVCR("retrieve");
--- a/EasyPost/Exceptions/General/EndOfPaginationError.cs
+++ b/EasyPost/Exceptions/General/EndOfPaginationError.cs
@@ -1,0 +1,13 @@
+namespace EasyPost.Exceptions.General
+{
+    public class EndOfPaginationError : EasyPostError
+    {
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="EndOfPaginationError" /> class.
+        /// </summary>
+        internal EndOfPaginationError()
+            : base("There are no more items to iterate through.")
+        {
+        }
+    }
+}
--- a/EasyPost/Exceptions/General/EndOfPaginationError.cs
+++ b/EasyPost/Exceptions/General/EndOfPaginationError.cs
@@ -6,7 +6,7 @@ namespace EasyPost.Exceptions.General
         ///     Initializes a new instance of the <see cref="EndOfPaginationError" /> class.
         /// </summary>
         internal EndOfPaginationError()
-            : base("There are no more items to iterate through.")
+            : base("There are no more pages to retrieve.")
         {
         }
     }
--- a/EasyPost/Models/API/EndShipper.cs
+++ b/EasyPost/Models/API/EndShipper.cs
@@ -2,7 +2,6 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using EasyPost._base;
 using EasyPost.BetaFeatures.Parameters;
-using EasyPost.Exceptions.General;
 using EasyPost.Models.Shared;
 using EasyPost.Utilities.Internal.Attributes;
 using EasyPost.Utilities.Internal.Extensions;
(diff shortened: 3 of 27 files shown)


EasyPost/Models/API/EndShipper.cs
```diff
@@ -90,5 +91,34 @@ public class EndShipperCollection : Collection
...
+            {
+                throw new EndOfPaginationError();
+            }
+
+            if (HasMore == null || !(bool)HasMore)
+            {
+                throw new EndOfPaginationError();
+            }
+
+            // TODO: How would we get the next page of EndShippers?
```
Justintime50, line 119: Answer: we wouldn't. Without the ability to set `before_id` on EndShipper, people cannot currently get the next page. Since records are returned most recent to oldest, there is currently no parameter usable to get what we'd need so for EndShipper, it would just throw an error till this is patched at the API level.

Justintime50 (commented):
Question: I don't love that we'd need to implement this across all the different services. I forgot that this will need the client and can't just be a loose utility function

nwithan8:
> Question: I don't love that we'd need to implement this across all the different services. I forgot that this will need the client and can't just be a loose utility function

What is the question?

Justintime50:
> > Question: I don't love that we'd need to implement this across all the different services. I forgot that this will need the client and can't just be a loose utility function
> 
> What is the question?

What alternatives do we have? Are there any?

nwithan8:
‏

> > > Question: I don't love that we'd need to implement this across all the different services. I forgot that this will need the client and can't just be a loose utility function
> > 
> > 
> > What is the question?
> 
> What alternatives do we have? Are there any?

Not much that I can see. We'd either need to implement this across all the services, or across all the different collections (not advised since we're trying to move away from that instance functions design).

Each "get more" function requires a different collection type, different type inside the collection, a different endpoint and a different set of API parameters, so it would be quite tedious if we made a one-off function and had the end-user provide all that information rather than the library handling it behind the scenes.

I'll look into how to abstract it away even more, but as you pointed out, we will need the API client to execute this function since it makes an API call.

EasyPost/Models/API/EndShipper.cs
Justintime50: Because this isn't implemented at the API level, let's not implement anything here other than throwing an error maybe, something along the lines of "can't get next page of endshippers"

EasyPost/Models/Shared/Collection.cs
```diff
@@ -1,15 +1,65 @@
...
+        /// <summary>
+        ///     Get the next page of a paginated collection.
+        /// </summary>
+        /// <param name="apiCallFunction">The function to execute to retrieve a page (most likely an All function).</param>
+        /// <param name="currentEntries">The results on the current page. Used to determine the API call parameters to retrieve the next page.</param>
+        /// <typeparam name="T">The type of <see cref="Collection"/> to get the next page of.</typeparam>
+        /// <typeparam name="T2">The type of <see cref="EasyPost._base.EasyPostObject"/> the entries are.</typeparam>
+        /// <returns>The next page of a paginated collection.</returns>
+        /// <exception cref="EndOfPaginationError">Thrown if there is no next page to retrieve.</exception>
+        internal async Task<T> GetNextPage<T, T2>(Func<Dictionary<string, object>, Task<T>> apiCallFunction, List<T2>? currentEntries) where T : Collection where T2 : EasyPost._base.EasyPostObject
```
Justintime50, line 29: issue: I'm not seeing a way to pass an optional page size of the next page which is part of the acceptance criteria

EasyPost/Exceptions/General/EndOfPaginationError.cs
```diff
@@ -0,0 +1,13 @@
+namespace EasyPost.Exceptions.General
+{
+    public class EndOfPaginationError : EasyPostError
+    {
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="EndOfPaginationError" /> class.
+        /// </summary>
+        internal EndOfPaginationError()
+            : base("There are no more items to iterate through.")
```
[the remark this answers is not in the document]
nwithan8: Resolved

EasyPost.Tests/ServicesTests/AddressServiceTest.cs
```diff
@@ -121,6 +121,21 @@ public async Task TestAll()
...
+        public async Task TestGetNextPage()
+        {
+            UseVCR("get_next_page");
+
+            AddressCollection addressCollection = await Client.Address.All(new Dictionary<string, object> { { "page_size", Fixtures.PageSize } });
+
+            AddressCollection nextPageAddressCollection = await Client.Address.GetNextPage(addressCollection);
+
+            // If the first ID in the next page is the same as the first ID in the current page, then we didn't get the next page
+            Assert.NotEqual(addressCollection.Addresses[0].Id, nextPageAddressCollection.Addresses[0].Id);
```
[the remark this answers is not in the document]
nwithan8: Resolved

nwithan8:
Per our internal conversations, it sounds like we are proceeding with this POC, correct?

jchen293 (approved):

Justintime50 (approved):
Looks good! Let's get it added to all the rest next

PR #422, opened 2023-03-23
[feat] Implement GetNextPage function

nwithan8:
# Description

This PR completes the work from #420 , implementing the GetNextPage function for all appropriate services.

- Tweak to `Collection` base class (renamed to `PaginatedCollection`, added type parameter for clarity, type-strictness, reduction of type parameters for each method)
- Implement parameter set generation for each `PaginatedCollection` inheritor (all "collections" except for `ApiKeyCollection`, which is not a paginated collection)
- Implement `GetNextPage` function for all expected services (skip EndShipper and Batch services due to API issues)

# Testing

- Add unit tests, cassettes for `GetNextPage` function for each service

# Pull Request Type


EasyPost/Models/API/Event.cs
```diff
@@ -33,7 +34,7 @@ internal Event()
         }
     }
 
-    public class EventCollection : Collection
+    public class EventCollection : PaginatedCollection<Event>
```
Justintime50, line 37: question: I don't love we're changing the return type here for fear of what it may to do user integrations. What is the user-impact of making this change?

nwithan8: We're not changing the return type, necessarily, but the underlying base class for all the `XCollection` classes. I don't imagine any user would be doing something along the lines of "I want to get a `Collection` class, but not specifically a `AddressCollection` class"

Justintime50 (approved):

jchen293 (approved):

PR #424, opened 2023-03-28
Account for special parameters for All and GetNextPage for shipments and trackers

nwithan8:
# Description

The List endpoints for [Shipments](https://www.easypost.com/docs/api#retrieve-a-list-of-shipments) and [Trackers](https://www.easypost.com/docs/api#retrieve-a-list-of-trackers) take special parameters in addition to the typical `before_id`, `after_id`, etc. parameters.

This PR:
- Captures those unique parameters (both from legacy dictionary and new parameter object flows) and stores them in the resultant `ShipmentCollection` and `TrackerCollection` objects, respectively (internal; cannot be viewed by end-user)
- Reuses these parameters on GetNextPage parameter construction, to ensure any parameters included on the first request for an XCollection will be reused when getting the subsequent pages

This was already in effect for `ReportCollection`, which needed to store the `type` to reuse for the GetNextPage function.

# Testing

- Adds unit tests to ensure that parameters are being captured from dictionary/parameter objects and stored in the XCollection objects
- Adds unit tests to ensure that parameters are being extracted from XCollection objects and reused in GetNextPage parameter construction

# Pull Request Type

### DIFF
--- a/EasyPost/BetaFeatures/Parameters/Shipments/All.cs
+++ b/EasyPost/BetaFeatures/Parameters/Shipments/All.cs
@@ -42,6 +42,18 @@ namespace EasyPost.BetaFeatures.Parameters.Shipments
         [TopLevelRequestParameter(Necessity.Optional, "start_datetime")]
         public string? StartDatetime { get; set; }
 
+        /// <summary>
+        ///     If true, only return purchased shipments. Defaults to true.
+        /// </summary>
+        [TopLevelRequestParameter(Necessity.Optional, "purchased")]
+        public bool? Purchased { get; set; } = true;
+
+        /// <summary>
+        ///     If true, also include shipments created by child users.
+        /// </summary>
+        [TopLevelRequestParameter(Necessity.Optional, "include_children")]
+        public bool? IncludeChildren { get; set; }
+
         #endregion
     }
 }
--- a/EasyPost.Tests/BetaFeaturesTests/ServicesTests/ReportServiceTest.cs
+++ b/EasyPost.Tests/BetaFeaturesTests/ServicesTests/ReportServiceTest.cs
@@ -121,6 +121,26 @@ namespace EasyPost.Tests.BetaFeaturesTests.ServicesTests
             }
         }
 
+        /// <summary>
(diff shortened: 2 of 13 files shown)


Justintime50 (approved):

EasyPost/BetaFeatures/Parameters/Shipments/All.cs
```diff
@@ -42,6 +42,18 @@ public sealed class All : BaseParameters
...
+        ///     If true, only return purchased shipments. Defaults to true.
+        /// </summary>
+        [TopLevelRequestParameter(Necessity.Optional, "purchased")]
+        public bool? Purchased { get; set; } = true;
+
+        /// <summary>
+        ///     If true, also include shipments created by child users.
+        /// </summary>
+        [TopLevelRequestParameter(Necessity.Optional, "include_children")]
+        public bool? IncludeChildren { get; set; }
```
jchen293, line 55:
```suggestion
        public bool? IncludeChildren { get; set; } = false;
```

Question: should we also include default value for `IncludeChildren`

jchen293 (approved):

EasyPost/BetaFeatures/Parameters/Shipments/All.cs
nwithan8: Thoughts @Justintime50 ? We could always send `false` by default. Or, if we don't indicate a value here, it's wired to not send anything, which means it would fall back to the server (which, according to our docs, defaults to `false` for this).

I believe we spoke earlier about wanting all boolean parameters, other than carbon offset, to not be included unless explicitly set by the end-user.

jchen293: I asked this because there is a default value for `purchased` in the above lines.

Justintime50: Let's try not to set these and let the API handle the defaults. If ever they changed, we wouldn't need to make alterations to the libs when they aren't set.

Justintime50 (approved):

PR #425, opened 2023-03-29
POC: Store parameters used for API call in resultant objects

nwithan8:
# Description

As we discovered in https://github.com/EasyPost/easypost-csharp/pull/424, certain `All()` API calls accept custom parameters, and keeping track of those parameters for future use (e.g. in getting the next page of a collection) proves to be a bit verbose.

For example, right now we store the `IncludedChildren` and `Purchased` parameters used in the initial call for getting all Shipments inside the resultant `ShipmentCollection` object as individual properties. These are then extracted and reused when crafting the parameters to get the next page of a Shipment collection.

This works for now, but it's not scalable. What if there were a dozen custom parameters we have to track?

Instead, this POC demonstrates storing the entire set of parameters used when making an API call inside the resultant object, with some limitations:
- Currently, this is only done for classes inheriting a specific `AllBaseParameter` class (we don't need to store the parameters used for, e.g. a Create Address call inside the resultant Address). As such the class is named since it would only apply to parameter sets for `All()` functions, which have a reusability purpose now (thanks to the `GetNextPage` functionality)
- Only the parameter set for getting all Shipments inherits this class, so it is the only one that can utilize the process below

All `PaginatedCollection` objects, including `ShipmentCollection` have a new `Filters` property (internal, inaccessible currently to end-users) that, during the JSON deserialization process, will be populated with the parameter set used to make the API call (this also gets a bit confusing because we have to account for dictionaries and parameter objects, until the legacy dictionary is dropped)

The function to craft the parameters for getting the next page of a `ShipmentCollection` has been updated to use the cached `Filters` parameters set as a base. This ensures that any parameters used to retrieve the first page of a `ShipmentCollection` will be reused to retrieve all subsequent pages.

This scales better when there are more parameters to track, utilizes abstracted functions behind the scenes that ensure consistency, and may prove helpful to end-users (if the parameter cache is made public, users could audit what filters were used to retrieve a given collection). This could also be expanded to all `EasyPostObject`s, storing the parameters used to make any given API call for later reference.

# Testing

- Existing `GetNextPage` and parameter handoff unit tests for Shipment have been modified to work with the new cache technique
  - Re-recorded cassettes as needed

# Pull Request Type


nwithan8:
Closing in favor of #449

PR #426, opened 2023-03-31
[chore] Tools to help Mac release process

nwithan8:
# Description

- new script, makefile step to install tools needed for releasing on Mac

# Testing

- Works

# Pull Request Type

### DIFF
--- a/scripts/unix/install_osslsigncode.sh
+++ b/scripts/unix/install_osslsigncode.sh
@@ -31,7 +31,7 @@ OSSLSIGNCODE_EXE=$(find . -name "osslsigncode")
 chmod +x "$OSSLSIGNCODE_EXE"
 
 # Move the executable to the PATH_STORAGE folder
-mv bin/osslsigncode "$PATH_STORAGE"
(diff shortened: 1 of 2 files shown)


scripts/unix/install_osslsigncode.sh
```diff
@@ -0,0 +1,38 @@
...
+cd "$TEMP_FOLDER" || exit
+
+# Download the latest macOS release
+# Courtesy: https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8
+REPO="mtrojnar/osslsigncode"
+curl -s https://api.github.com/repos/$REPO/releases/latest \
+| grep "browser_download_url.*macOS.zip" \
+| cut -d : -f 2,3 \
+| tr -d \" \
+| wget -qi -
```
Justintime50, line 19: thought: I don't love this, very brittle. We can roll with this for now since it'll only affect us but we should try to avoid clever hacks like this

Justintime50 (approved):

PR #431, opened 2023-04-18
feat: Add carrier metadata endpoint (beta)

nwithan8:
# Description

- Add new carrier metadata endpoint (beta) to beta client
- Add custom assertion tools (.Any surprisingly missing from Xunit)

# Testing

- Add unit tests, cassettes for carrier metadata

# Pull Request Type


nwithan8:
The custom assertion toolset might be a bit overzealous. I was just surprised that Xunit has a `.All` assertion, but not a `.Any` (or `.Some`, etc.). Rather than dealing with a boolean in a one-off unit test, I figured making a proper assertion function for it might help for reusability later on.

Just tell me to remove it like I fully anticipate happening 😄

EasyPost/Models/API/Beta/CarrierMetadata.cs
```diff
@@ -0,0 +1,148 @@
...
+        public List<SupportedFeature>? SupportedFeatures { get; set; }
+
+        #endregion
+    }
+
+    public class PredefinedPackage
+    {
+        #region JSON Properties
+
+        [JsonProperty("carrier")]
```
Justintime50, line 46: nit: the `carrier` and `name` will always be populated on each of these types while everything else is optional. Could remove the question mark

Justintime50 (approved):

jchen293 (approved):
Need a changelog when releasing it later today

PR #433, opened 2023-04-20
Expose customizable connection timeout, improve HTTP logic, drop RestSharp dependency

nwithan8:
# Description

This PR:
- Cherry-picks most of the work done in https://github.com/EasyPost/easypost-csharp/pull/384
- Exposes a customizable connection timeout to end-users
- Gracefully handles API timeouts (should throw a clean `TimeoutError` exception rather than a `TaskCanceledException`)
- Deprecates current parameterized EasyPost client constructor, guides user to new `ClientConfiguration`-geared constructor
- Improves behind-the-scenes logic of HTTP request construction, execution and processing
- Improves behind-the-scenes efficiency of initializing and disposing clients and client services
- Internal alterations to PartnerService as needed to account for temporary API key swapping

No breaking changes expected. The only changes to end-user experience:
- Old EasyPost client constructor is deprecated (not removed), new constructor encouraged
- Connection Timeout is now a customizable option (technically, this was altered internally, but was never publicly available, so that change doesn't break any prior user behavior)

# Testing

- Unit tests, utilities updated as needed
  - Exception tests needed to account for different status codes, since RestSharp was mutating response codes before we intercepted them
  - Unit tests utilities adopt the new EasyPost client constructor

# Pull Request Type

Please select the option(s) that are relevant to this PR.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

PR #434, opened 2023-04-24
Better error message for API timeouts

nwithan8:
# Description

Better handling for API timeout errors. Rather than proceeding with error handling, which produces a `ConnectionError` exception with an ambiguous error message, an API timeout will instead produce a `TimeoutError` with an understandable error message.

Portion of #433 

# Testing

- New unit test to confirm functionality

# Pull Request Type

### DIFF
--- a/EasyPost/Exceptions/API/_Base.cs
+++ b/EasyPost/Exceptions/API/_Base.cs
@@ -4,6 +4,7 @@ using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Reflection;
+using System.Threading.Tasks;
 using EasyPost.Models.API;
 using EasyPost.Utilities.Internal;
 using RestSharp;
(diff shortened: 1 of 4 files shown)


EasyPost/Exceptions/API/_Base.cs
```diff
@@ -73,6 +74,13 @@ internal static ApiError FromErrorResponse(RestResponse response)
         {
             // NOTE: This method anticipates that the status code will be a non-2xx code.
             // Do not use this method to parse a successful response.
+
+            // short-circuit if the request timed out
+            if (response.ErrorException?.GetType() == typeof(TaskCanceledException))
```
Justintime50, line 79: question: is this check robust enough to catch all timeouts? Meaning, I know that for this particular use-case it triggered a `TaskCanceledException`, but will that always be the case?

Justintime50 (approved):
Perfect, thanks!

jchen293 (approved):

PR #438, opened 2023-04-24
Remove RestSharp usage

nwithan8:
# Description

Removes all usage of RestSharp and drops the package as a dependency.

This library was using RestSharp as an HTTP client, but was not utilizing the package's JSON deserialization capabilities after [a custom de/serializer was implemented last year](https://github.com/EasyPost/easypost-csharp/commit/d06961811ae613fd0d69808fae4e65ac1e143df9). Rather than continuing to include this package under-utilized, the library has been restructured internally to simply use the default `HttpClient` built into the .NET framework and drop an unnecessary dependency.

Due to the immutability of connection timeouts (once the timeout length for an HttpClient has been set, it cannot be changed), connection timeout must be set during the construction of the `EasyPostClient` (and subsequent Beta client). This PR changes the construction signature of the `EasyPostClient` constructor to accept an optional `timeoutMilliseconds` parameter. If not set, the default 30-second timeout will be used. This is an optional parameter, so it should not impact the majority of users. Because of its placement in the constructor parameter order, a user may have to account for this if they are using the optional `customHttpClient` parameter.

Part of #433 

# Testing

- Unit tests, utilities updated to account for change
  - Regex patterns for mocking had to be slightly tweaked to account for domain presence
- No cassettes needed to be re-recorded, suggesting that the RestSharp switch does not impact the final HTTP request, and that the previous HTTP responses that were deserialized properly with RestSharp present are still deserialized properly without RestSharp present.

# Pull Request Type

### DIFF
--- a/EasyPost.Tests/ClientTest.cs
+++ b/EasyPost.Tests/ClientTest.cs
@@ -21,12 +21,9 @@ namespace EasyPost.Tests
         {
             // ReSharper disable once UseObjectOrCollectionInitializer
             // we specifically want to test the getters/setters
-            Client client = new(FakeApikey);
-            client.Configuration.ConnectTimeoutMilliseconds = 5000;
-            client.Configuration.RequestTimeoutMilliseconds = 5000;
+            Client client = new(FakeApikey, timeoutMilliseconds: 5000);
 
-            Assert.Equal(5000, client.Configuration.ConnectTimeoutMilliseconds);
-            Assert.Equal(5000, client.Configuration.RequestTimeoutMilliseconds);
+            Assert.Equal(5000, client.ConnectTimeoutMillisecondsInUse);
         }
 
         [Fact]
--- a/EasyPost.Tests/ClientTest.cs
+++ b/EasyPost.Tests/ClientTest.cs
@@ -23,7 +23,7 @@ namespace EasyPost.Tests
             // we specifically want to test the getters/setters
             Client client = new(FakeApikey, timeoutMilliseconds: 5000);
 
-            Assert.Equal(5000, client.TimeoutMillisecondsInUse);
+            Assert.Equal(5000, client.TimeoutMilliseconds);
         }
 
         [Fact]
--- a/EasyPost.Tests/ClientTest.cs
+++ b/EasyPost.Tests/ClientTest.cs
@@ -23,7 +23,7 @@ namespace EasyPost.Tests
             // we specifically want to test the getters/setters
             Client client = new(FakeApikey, timeoutMilliseconds: 5000);
 
-            Assert.Equal(5000, client.ConnectTimeoutMillisecondsInUse);
+            Assert.Equal(5000, client.TimeoutMillisecondsInUse);
         }
 
         [Fact]
--- a/EasyPost/Exceptions/API/_Base.cs
+++ b/EasyPost/Exceptions/API/_Base.cs
@@ -3,11 +3,11 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
 using System.Reflection;
 using System.Threading.Tasks;
 using EasyPost.Models.API;
 using EasyPost.Utilities.Internal;
-using RestSharp;
 
 namespace EasyPost.Exceptions.API
 {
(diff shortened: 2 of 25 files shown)


jchen293 (approved):

EasyPost/Exceptions/API/_Base.cs
```diff
@@ -64,23 +64,17 @@ protected ApiError(string errorMessage, int? statusCode = null, string? errorTyp
...
+        /// <param name="response">HttpResponseMessage response to parse.</param>
+        /// <raises>EasyPostError if an unplanned response code is found (i.e. user passed in a non-error HttpResponseMessage response object with a 2xx status code).</raises>
         /// <returns>An instance of an HttpError-inherited exception.</returns>
-        internal static ApiError FromErrorResponse(RestResponse response)
+        internal static async Task<ApiError> FromErrorResponse(HttpResponseMessage response)
         {
             // NOTE: This method anticipates that the status code will be a non-2xx code.
             // Do not use this method to parse a successful response.
 
-            // short-circuit if the request timed out
```
Justintime50, line 78: question: where is this now captured since this is getting reworked? Are we certain we still raise a timeout error?

EasyPost/Http/ClientConfiguration.cs
```diff
@@ -120,11 +116,6 @@ private abstract class Defaults
             ///     The default connection timeout in milliseconds.
             /// </summary>
             internal const int DefaultConnectTimeoutMilliseconds = 30000;
```
Justintime50, line 118: issue: we want the timeout to be 60 seconds when only one of the two can be set. Please adjust accordingly.

Additionally, when only one is available, it's usually the request timeout but here we are retaining the connection timeout, can we determine for certain which one the .NET http package actually exposes and rename if necessary?

EasyPost/Http/Request.cs
```diff
@@ -1,37 +1,56 @@
...
+using System.Net.Http;
+using System.Text;
 using EasyPost._base;
 using EasyPost.Utilities.Internal;
-using RestSharp;
-using RestSharp.Serializers;
 
 namespace EasyPost.Http
 {
+    // TODO: Make disposable and dispose of the request message
```
Justintime50, line 10: question: why not do this now?

EasyPost/_base/EasyPostClient.cs
```diff
@@ -21,57 +18,68 @@ public abstract class EasyPostClient
...
+        public int ConnectTimeoutMillisecondsInUse => (int)HttpClient.Timeout.TotalMilliseconds; // public read-only property so users can audit the connect timeout used by the client
+
+        /// <summary>
+        ///     Gets the custom HTTP client used by this client.
+        /// </summary>
+        public HttpClient? CustomHttpClientInUse => Configuration.HttpClient; // public read-only property so users can audit the custom HTTP client they set to be used by the client
+
+        /// <summary>
+        ///     Gets the prepared HTTP client used by this client for all API calls.
+        /// </summary>
```
Justintime50, line 43: question: shouldn't a user only be able to get one or the other? Meaning, there is only ever a custom HTTP client or the default in use, never both?

Justintime50 (changes requested):

EasyPost/Exceptions/API/_Base.cs
nwithan8: Yes, the unit test for it still passes. It's captured in the `ExecuteRequest` function in the client upstream and earlier

EasyPost/Http/ClientConfiguration.cs
nwithan8: The Http client simply has a property called `Timeout`. We can rename it to that to make it simple.

image

EasyPost/Http/Request.cs
nwithan8: Going to in a separate PR. Gotta stay focused 😉

EasyPost/_base/EasyPostClient.cs
nwithan8: The `CustomHttpClient` is basically echoing what they passed in to the constructor, useful if they need to get that back (or use it to construct a new client, as we do for the Partner service).

The `PreparedClient` is basically the custom client or a standard client if one was not provided, with the timeout already set up. This is internal-only and therefore is purposefully inaccessible to the end user.

EasyPost/Http/Request.cs
Justintime50: My my, I may cry 😂

EasyPost.Tests/ClientTest.cs
```diff
@@ -23,7 +23,7 @@ public void TestClient()
             // we specifically want to test the getters/setters
             Client client = new(FakeApikey, timeoutMilliseconds: 5000);
 
-            Assert.Equal(5000, client.ConnectTimeoutMillisecondsInUse);
+            Assert.Equal(5000, client.TimeoutMillisecondsInUse);
```
Justintime50, line 26: question: can we just call it `TimeoutMilliseconds`? What's the `InUse` bit?

EasyPost/Http/ClientConfiguration.cs
```diff
@@ -91,12 +91,12 @@ internal ClientConfiguration(string apiKey, string? baseUrl, int? timeoutMillise
...
 
             // Optional constructor parameters with defaults
-            ApiBase = baseUrl ?? Defaults.DefaultBaseUrl;
+            ApiBase = baseUrl ?? "https://api.easypost.com";
             HttpClient = customHttpClient ?? new HttpClient();
 
             // Prepare the HttpClient
             PreparedHttpClient = HttpClient;
-            PreparedHttpClient.Timeout = TimeSpan.FromMilliseconds(timeoutMilliseconds ?? Defaults.DefaultConnectTimeoutMilliseconds);
+            PreparedHttpClient.Timeout = TimeSpan.FromMilliseconds(timeoutMilliseconds ?? 30000); // Default to 30 seconds
```
Justintime50, line 99: issue: as mentioned before, this needs to be 60 seconds, not 30

Justintime50 (changes requested):

EasyPost.Tests/ClientTest.cs
nwithan8: I have addressed this.

I plan on a separate PR later that will clean up the configuration process for clients, so this will likely change again in that work.

Justintime50 (approved):

Issue #435, opened 2023-04-24
[Feat]: Update to latest RestSharp 110


MaxwellDAssistek:
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

We use the latest version of RestSharp in our systems and we would prefer to avoid downgrading.

nwithan8:
Hello, thanks for reaching out!

We are actually right now in the process of dropping RestSharp as a dependency from our library, since it's not being used. That should remove the dependency conflict from your system. We'll keep you updated on the process!

nwithan8:
Hey @MaxwellDAssistek , wanted to let you know we are in officially in the process of dropping RestSharp in our next major release (along with a number of breaking changes and improvements).

While we encourage you to update the the latest version when it gets released, in the meantime, we've released [v4.6.1](https://github.com/EasyPost/easypost-csharp/releases/tag/v4.6.1), which simply extends the RestSharp dependency range so you'll be able to use RestSharp v110 in your solution.

I'm going to go ahead and close this ticket, but let us know if there's anything else you need help with!

MaxwellDAssistek:
@nwithan8 Unfortunately, I get the following exception when using the latest RestSharp:

```
Error: System.MissingMethodException: Method not found: 'Void RestSharp.RestClient..ctor(System.Net.Http.HttpClient, RestSharp.RestClientOptions, Boolean)'.
  at EasyPost._base.EasyPostClient..ctor(String apiKey, String baseUrl, HttpClient customHttpClient)
  at EasyPost.Client..ctor(String apiKey, String baseUrl, HttpClient customHttpClient)
  at LogisticsBusinessLogic.Services.ShippingService.WebhookTest() in <SNIP>\ShippingService.cs:line xx
```

Downgrading RestSharp to 108.0.1 fixes this.

nwithan8:
We'll look into this and see if we can fix it. Thanks for reporting.

MaxwellDAssistek:
@nwithan8 We are using .NET 7.

dobbamca:
hi
any update on this?

nwithan8:
[...]

PR #439, opened 2023-04-24
Extend RestSharp dependency range

nwithan8:
# Description

The next major release of the library will drop RestSharp as a dependency (#438 ), which has caused dependency conflicts for some of our users (#435 ).

However, we don't want to force users who simply need to resolve a dependency conflict to adopt the numerous breaking changes that will be introduced in the next version.

This PR simply extends the dependency range for RestSharp in our library up to and including all v110 versions, the [latest major version of RestSharp](https://github.com/restsharp/RestSharp/releases).

I have confirmed that this library, as written, works with RestSharp v110 with no issue.

# Testing

- All unit tests pass

# Pull Request Type


nwithan8:
Workflow failures due to not reaching GitHub to pull down examples repo. Sounds like GitHub Actions is having a rough day, will re-run tomorrow.

jchen293 (approved):

EasyPost/EasyPost.csproj
```diff
@@ -66,7 +66,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="[13.0.1, 14.0.0)" />
-    <PackageReference Include="RestSharp" Version="[108.0.1, 109.0.0)" />
+    <PackageReference Include="RestSharp" Version="110.2.0" />
```
Justintime50, line 69: This is a major breaking change since it'll require 110 and we are backporting this. I'd suggest instead that maybe we simply remove the max pin and keep the min pin. That way it should continue to work for folks until RestSharp break something bad.

Justintime50 (changes requested):

EasyPost/EasyPost.csproj
nwithan8: Oops, that was my original intention. Seems my IDE removed the range when I actually went to update RestSharp. This has been rectified.

Justintime50 (approved):
Nice!

PR #441, opened 2023-04-26
[chore] Move all instance (model) functions to services

nwithan8:
# Description

The initial thread-safe redesign for the .NET library was incomplete, and was refined for later rewrites such as the Java library. This PR brings the architectural design of this library in line with the other client libraries, namely by removing all "instance" functions. All API-calling functions occur on the various service classes.

- Instance-based functions removed
  - All models (i.e. a Shipment, an Order, a Parcel) are just JSON properties (and maybe a few, non-HTTP call helper methods), no longer contain API-calling functions, no long contain an embedded Client clone
- All HTTP methods are called via the services in a Client object
- Behind-the-scenes improvements, removing unnecessary abstractions that only clutter the code base

This is a major breaking change and will require users to adjust their implementation of this library accordingly.

# Testing

- Unit tests were consolidated and updated accordingly
- All cassettes were re-recorded to account for the change in cassette grouping

# Pull Request Type


Justintime50 (changes requested):
600+ files changed and 20k+ lines altered isn't feasible to review, my browser is crashing just trying to write this message. We'll need to break this down into smaller chunks.

nwithan8:
Closing in favor of partial PRs

PR #442, opened 2023-04-26
chore] Move all instance (model) functions to services (part 1)

nwithan8:
# Description

Subpart of #441 , see for more details

# Testing

- Unit tests moved, re-recorded as needed

# Pull Request Type


EasyPost/_base/EasyPostClient.cs
```diff
@@ -21,12 +21,12 @@ public abstract class EasyPostClient
         /// <summary>
         ///     Gets the API key used by this client.
         /// </summary>
-        public string ApiKey => Configuration.ApiKey; // public read-only property so users can audit the API key used by the client
+        public string ApiKeyInUse => Configuration.ApiKey; // public read-only property so users can audit the API key used by the client
```
Justintime50, line 24: question: I thought we were doing away with the `InUse` bit?

EasyPost/packages.lock.json
```diff
@@ -7,6 +7,20 @@
         "requested": "[13.0.1, 14.0.0)",
         "resolved": "13.0.1",
         "contentHash": "ppPFpBcvxdsfUonNcvITKqLl3bqxWbDCZIzDWHzjpdAHRFfZe0Dw9HmA0+za13IdyrgJwpkDTDA9fHaxOrt20A=="
+      },
```
Justintime50, line 10: question: why are these files continuing to get bumped on most PRs for this lib. The whole point of a lock file is they are locked so deps don't keep being transient 😉

EasyPost/Services/EndShipperService.cs
```diff
@@ -92,6 +93,32 @@ public async Task<EndShipper> Create(BetaFeatures.Parameters.EndShippers.Create
...
+
+        /// <summary>
+        ///     Update this <see cref="EndShipper"/>.
+        /// </summary>
+        /// <param name="parameters"><see cref="BetaFeatures.Parameters.EndShippers.Update"/> parameter set.</param>
+        /// <returns>This updated <see cref="EndShipper"/> instance.</returns>
+        [CrudOperations.Update]
+        public async Task<EndShipper> Update(string id, BetaFeatures.Parameters.EndShippers.Update parameters)
+        {
+            // EndShipper needs Put, not Patch
```
Justintime50, line 118: thought: I don't know that this needs a comment in-code. This was more something we discovered along the way and will be codified in the code itself.

EasyPost/Services/CarrierAccountService.cs
```diff
@@ -19,6 +20,8 @@ internal CarrierAccountService(EasyPostClient client)
 
         #region CRUD Operations
 
+        // TODO: Use ID or whole object as parameter?
```
Justintime50, line 23: question: what does this TODO mean? You should be able to pass either

EasyPost/Services/BatchService.cs
```diff
@@ -103,13 +106,163 @@ public async Task<Batch> CreateAndBuy(BetaFeatures.Parameters.Batches.Create par
 
         // TODO: Add GetNextPage function when Batches are sorted newest to oldest.
 
+        /// <summary>
+        ///     Add shipments to this batch.
+        /// </summary>
+        /// <param name="parameters">Update shipment parameters.</param>
+        /// <returns>The updated Batch.</returns>
+        [CrudOperations.Update]
+        public async Task<Batch> AddShipments(string id, Dictionary<string, object> parameters)
```
Justintime50, line 115: issue: we need to be addressing all these Roslyn analysis checks. Please add the missing ID to the docstring.

EasyPost/Services/BatchService.cs
```diff
@@ -103,13 +106,163 @@ public async Task<Batch> CreateAndBuy(BetaFeatures.Parameters.Batches.Create par
...
+        {
+            // parameters = parameters.Wrap("batch");  // TODO: Update docs to remove wrapped "batch" key
```
Justintime50, line 117: question: if we have a TODO about a project that isn't directly in this code, we should be creating tasks instead of leaving TODOs in a place they'll be forgotten about because it's not related.

Why do we need to remove the wrapped batch key from the docs?

Let's remove this TODO and instead create a more actionable follow up (if necessary)

EasyPost/EasyPost.csproj
```diff
@@ -10,7 +10,7 @@
     <Configurations>Release;Debug;Linting</Configurations>
     <Platforms>AnyCPU</Platforms>
     <PackageId>EasyPost</PackageId>
-    <LangVersion>9</LangVersion>
+    <LangVersion>10</LangVersion>
```
Justintime50, line 13: issue: we shouldn't be bumping dependencies along with code changes, these are mixing too many concerns and should typically happen separately

EasyPost/Client.cs
```diff
@@ -6,59 +6,58 @@ namespace EasyPost
 {
     public class Client : EasyPostClient
     {
-        public AddressService Address => GetService<AddressService>();
+        public AddressService Address { get; }
 
-        public ApiKeyService ApiKey => GetService<ApiKeyService>();
+        public ApiKeyService ApiKey { get; } // TODO: Recommend renaming. Because ApiKey is used as the name of the API key service, you have to access the actual configured API key via ApiKeyInUse
```
Justintime50, line 11: thought: ugh.

this is annoying

Justintime50 (commented):
I assume most of the tests were simply moved from the ModelsTest to ServiceTests?

Still don't love we split them out. Infact, can we just consolidate them all into Tests when this is over since models will have all of like 2-3 functions across the whole lib when done? I don't think at that point it makes sense to keep them separate and will reduce a bunch of testing complexity

EasyPost/_base/EasyPostClient.cs
nwithan8: Name clash with `ApiKey` meaning the `ApiKeyService`

EasyPost/Services/CarrierAccountService.cs
nwithan8: Note to self for later

EasyPost/Services/BatchService.cs
nwithan8: Going to do docstrings holistically in a separate PR

EasyPost/EasyPost.csproj
nwithan8: Not a dependency

EasyPost/packages.lock.json
nwithan8: I usually block these, it's my IDE. I'll reign it in. For what it's worth, these are dev utilities, not something that affects the end-user

EasyPost/Client.cs
nwithan8: We'll address it later

EasyPost/Services/BatchService.cs
nwithan8: Resolved

EasyPost/Services/EndShipperService.cs
nwithan8: Resolved

EasyPost/Client.cs
```diff
@@ -6,59 +6,58 @@ namespace EasyPost
...
 
-        public BatchService Batch => GetService<BatchService>();
+        public BatchService Batch { get; }
 
         // ReSharper disable once MemberCanBePrivate.Global
         // ReSharper disable once UnusedAutoPropertyAccessor.Global
         public BetaClient Beta { get; }
```
jchen293, line 17: Question: I believe C# is the only thread-safe CL that has a `Beta` namespace, and the rest of thread-safe CLs just prepend `Beta` in the class service name. Should we make this change in the V5 major bump?

EasyPost/Services/BatchService.cs
```diff
@@ -103,13 +106,161 @@ public async Task<Batch> CreateAndBuy(BetaFeatures.Parameters.Batches.Create par
 
         // TODO: Add GetNextPage function when Batches are sorted newest to oldest.
 
+        /// <summary>
+        ///     Add shipments to this batch.
+        /// </summary>
+        /// <param name="parameters">Update shipment parameters.</param>
+        /// <returns>The updated Batch.</returns>
+        [CrudOperations.Update]
+        public async Task<Batch> AddShipments(string id, Dictionary<string, object> parameters)
```
jchen293, line 115: Issue: need fix the docstring

EasyPost/Services/BatchService.cs
```diff
@@ -103,13 +106,161 @@ public async Task<Batch> CreateAndBuy(BetaFeatures.Parameters.Batches.Create par
...
-        public async Task<Batch> Retrieve(string id) => await Get<Batch>($"batches/{id}");
+        public async Task<Batch> Retrieve(string id) => await Request<Batch>(Method.Get, $"batches/{id}");
+
+        /// <summary>
+        ///     Add <see cref="Shipment"/>s to this <see cref="Batch"/>.
+        /// </summary>
+        /// <param name="parameters"><see cref="BetaFeatures.Parameters.Batches.AddShipments"/> parameter set.</param>
+        /// <returns>This updated <see cref="Batch"/> instance.</returns>
+        [CrudOperations.Update]
+        public async Task<Batch> AddShipments(string id, BetaFeatures.Parameters.Batches.AddShipments parameters)
```
jchen293, line 134: Issue: need to fix docstring. Please double check on the functions that were copied over, many of them are missing docstring for `id`

jchen293 (changes requested):
Code looks good, left two questions and there are some docstrings issues from the function that got moved around.

EasyPost/Services/BatchService.cs
nwithan8: Docstrings will be addressed in a later PR.

EasyPost.Tests/EasyPost.Tests.csproj
```diff
@@ -65,9 +65,28 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="BetaFeaturesTests\ModelsTests\EndShipperTest.cs" />
+    <Compile Remove="BetaFeaturesTests\ModelsTests\AddressTest.cs" />
+    <Compile Remove="BetaFeaturesTests\ModelsTests\BatchTest.cs" />
+    <Compile Remove="ModelsTests\CarrierAccountTest.cs" />
+    <Compile Remove="ModelsTests\BatchTest.cs" />
+    <Compile Remove="ModelsTests\EndShipperTest.cs" />
+    <Compile Remove="ModelsTests\AddressTest.cs" />
```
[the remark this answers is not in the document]
nwithan8: As I was moving files, looks like the IDE/solution was trying to modify the solution for me. I'll roll this back.

EasyPost/Client.cs
jchen293: I am leaning toward making this change in order to maintain consistency with other CLs.

jchen293 (approved):
lgtm so far as long as the docstring is addressed on a separate PR

Justintime50 (approved):

PR #446, opened 2023-04-27
Overhaul, consolidate client configuration

nwithan8:
# Description

- Users construct a ClientConfiguration object, pass in as sole parameter to a Client constructor
- Config values blackboxed internally, auditable via read-only properties on Client
- User-Agent header on requests now properly includes the targeted API version (previously, even requests made to beta endpoints were listed as `EasyPost/v2` in the User-Agent; now beta calls will say `EasyPost/Beta`. Important for eventual API versioning)

This PR alters how end-users will set the configuration for a Client.

```csharp
# Before
Client myClient = new Client("my_api_key", "optional_api_base_override", myOptionalCustomHttpClient, myCustomTimeoutMilliseconds)
```
```csharp
# After
Client myClient = new Client(new ClientConfiguration("my_api_key") {
  ApiBase = "optional_api_base_override",
  CustomHttpClient = myOptionalCustomHttpClient,
  Timeout = TimeSpan.FromMilliseconds(myCustomTimeoutMilliseconds)
}
```

Under the assumption that most users probably are not overriding the API base URL or providing a custom HTTP client or timeout, most users will simply see a change like this

```csharp
# Before 
Client myClient = new Client("my_api_key")
```
```csharp
# After
Client myClient = new Client(new ClientConfiguration("my_api_key"))
```

This change allows a few things.
- If additional configuration options get added in the future, they can simply be made as an additional optional/required property on the `ClientConfiguration` object, without changing the signature of the `Client` constructor (the `Client` constructor only takes one parameter, a `ClientConfiguration` object; this is similar to the idea behind the new Parameters set objects)
- The `ClientConfiguration` object passed into the `Client` constructor is stored privately, making it inaccessible (and therefore unmodifiable) by end users. 
  - The previous design allowed users to alter certain settings after construction, such as the timeout length. The problem/confusion there is that that alteration was never effective (the timeout for the internal HttpClient is immutable once set, so the users altering the `TimeoutMilliseconds` in the configuration actually did nothing).
  - Aspects of the configuration, such as which API key or what timeout is being used are available to the end-user in a readonly fashion via the `Client.ApiKeyInUse` and `Client.Timeout` properties.

# Testing

- Unit tests testing client configuration have been updated
- Unit test utilities setting up VCR and Mock clients have been updated
- No new cassettes needed to be recorded, suggesting this does not interfere with the resultant API/HTTP request.

# Pull Request Type


nwithan8:
Not sure why the linter here is complaining about some rogue spaces, can't reproduce locally...

nwithan8:
> Not sure why the linter here is complaining about some rogue spaces, can't reproduce locally...

Changed nothing, now it works.

Sufwear divelupmint

EasyPost/ClientConfiguration.cs
```diff
@@ -0,0 +1,123 @@
...
+    ///     The version of the current application's operating system.
+    /// </summary>
+    private readonly string _osVersion;
+
+    /// <summary>
+    ///     Get the User-Agent string to use for a request.
+    /// </summary>
+    /// <param name="apiVersion">Version of the API being used for this request.</param>
+    /// <returns>The prepared User-Agent string.</returns>
+    private string GetUserAgent(ApiVersion apiVersion) => $"EasyPost/{apiVersion.Value} CSharpClient/{_libraryVersion} .NET/{_dotNetVersion} OS/{_osName} OSVersion/{_osVersion} OSArch/{_osArch}";
```
Justintime50, line 75: question: has this changed at all? If so, we need to notify data who are tracking the agents of all libs

Justintime50 (approved):

EasyPost/ClientConfiguration.cs
nwithan8: The format has not changed.

jchen293 (approved):

PR #447, opened 2023-04-28
[chore] Explicit IDisposable implementations for better memory management

nwithan8:
# Description

This PR implements the IDisposable interface to a number of in-memory objects to improve memory management and avoid orphaned objects.

The garbage collector in .NET will likely catch the majority of these objects without explicit dispose instructions, but this guarantees that no-longer-referenced objects are in fact destroyed.

Specifically:
- After every HTTP request is made, the `Request` object is no longer needed, and so is explicitly destroyed once that process is finished.
- When an `EasyPostClient` that a user would create is no longer needed (perhaps it was a one-off instance to use a specific API key), this will ensure that the `ClientConfiguration` and all `EasyPostService`s associated with that client are also destroyed.

# Testing

- N/A

# Pull Request Type

### DIFF
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -40,7 +40,7 @@ namespace EasyPost
         public PickupService Pickup { get; }
 
         public RateService Rate { get; }
-        
+
         public ReferralCustomerService ReferralCustomer { get; }
 
         public RefundService Refund { get; }
--- a/EasyPost/Client.cs
+++ b/EasyPost/Client.cs
@@ -37,11 +37,11 @@ namespace EasyPost
 
         public ParcelService Parcel { get; }
 
-        public ReferralCustomerService ReferralCustomer { get; }
-
         public PickupService Pickup { get; }
 
         public RateService Rate { get; }
+        
+        public ReferralCustomerService ReferralCustomer { get; }
 
         public RefundService Refund { get; }
 
@@ -80,9 +80,9 @@ namespace EasyPost
             Insurance = new InsuranceService(this);
             Order = new OrderService(this);
             Parcel = new ParcelService(this);
-            ReferralCustomer = new ReferralCustomerService(this);
             Pickup = new PickupService(this);
             Rate = new RateService(this);
+            ReferralCustomer = new ReferralCustomerService(this);
             Refund = new RefundService(this);
             Report = new ReportService(this);
             ScanForm = new ScanFormService(this);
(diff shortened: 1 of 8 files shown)


EasyPost/ClientConfiguration.cs
```diff
@@ -120,4 +120,37 @@ internal void SetUp()
...
+
+            // dispose of the user-provided HTTP client
+            CustomHttpClient?.Dispose();
+        }
+
+        // Free native resources (unmanaged objects) and override a finalizer below.
+        _isDisposed = true;
+    }
+
+    ~ClientConfiguration()
```
jchen293, line 151: what does `~` do here?

nwithan8: This is a deconstructor, the opposite of a constructor. It runs when an object is destroyed in memory.

EasyPost/ClientConfiguration.cs
```diff
@@ -120,4 +120,37 @@ internal void SetUp()
 #pragma warning disable CA1307
     public override int GetHashCode() => ApiKey.GetHashCode() ^ ApiBase.GetHashCode() ^ Timeout.GetHashCode();
 #pragma warning restore CA1307
+
+    private bool _isDisposed;
+
+    public void Dispose()
```
Justintime50, line 126: suggestion: would love docstrings for these. Although "dispose" is pretty easy to follow, what exactly that does at runtime, the compiler isn't well known to everyone and could be documented.

Justintime50 (approved):

EasyPost/Client.cs
```diff
@@ -94,5 +94,49 @@ public Client(ClientConfiguration configuration)
...
+        protected override void Dispose(bool disposing)
+        {
+            // ref: https://dzone.com/articles/when-and-how-to-use-dispose-and-finalize-in-c
+            // ref: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1063#pseudo-code-example
+            if (disposing)
+            {
+                // Dispose managed state (managed objects).
+                // "disposing" inherently true when called from Dispose(), so don't need to pass it in.
+
+                // Dispose of the services
```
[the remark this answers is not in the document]
nwithan8: I'd rather not introduce reflection into a memory-management function.

EasyPost/ClientConfiguration.cs
nwithan8: Will do in the docstrings PR.

PR #457, opened 2023-05-02
[chore] Cancellation token best practice, parameter cleanup

nwithan8:
# Description

- Add optional cancellation token parameter to all HTTP functions, to allow end users to cancel operations if needed. This is [standard practice for async functions](https://medium.com/geekculture/how-to-support-the-http-request-cancelation-in-net-core-4b5561d9605d), especially network-related ones.
  - Stripe does it: https://github.com/stripe/stripe-dotnet/blob/c8bc4fce820ed394de85254a7e5a8302fe7b8182/src/Stripe.net/Services/TestHelpers/Issuing/Cards/CardService.cs#L28
- For functions that use the Parameters objects rather than dictionaries, consolidate all explicit parameters into the Parameters object, so that all functions accepting Parameters objects have one of these two signatures:
```csharp
Create(myParameterSet)
```

```csharp
Update(objectId, myParameterSet)
```
  - This only impacts the `All` function for retrieving all reports, which had an explicit separate `type` parameter that is now part of the `Report.All` parameter set passed into the function.
- Parameter-object functions no longer redirect to the Dictionary equivalent, but instead are self-contained. This makes them easier to maintain and read, and will make migration easier when Dictionary functions are eventually dropped entirely.
- Removed default "pdf" value for file format for labels and scan forms. Users must now explicitly declare what format they want.
- Removed ability to use `Retrieve` function in User service without passing in an ID to retrieve self.
  - `Retrieve` now requires a user ID be provided. To retrieve self, users should use the `RetreiveMe` function explicitly.
 

# Testing

- Updated unit tests as needed (file format required parameter)
  - No cassettes needed re-recording, suggesting that parameter changes do not impact the final HTTP request

# Pull Request Type

### DIFF
--- a/EasyPost.Tests/BetaFeaturesTests/ServicesTests/ReportServiceTest.cs
+++ b/EasyPost.Tests/BetaFeaturesTests/ServicesTests/ReportServiceTest.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using EasyPost.Exceptions.General;
 using EasyPost.Models.API;
 using EasyPost.Tests._Utilities;
 using EasyPost.Tests._Utilities.Attributes;
@@ -29,11 +30,12 @@ namespace EasyPost.Tests.BetaFeaturesTests.ServicesTests
             {
                 { "start_date", Fixtures.ReportDate },
                 { "end_date", Fixtures.ReportDate },
+                { "type", Fixtures.ReportType },
             };
 
             BetaFeatures.Parameters.Reports.Create parameters = Fixtures.Parameters.Reports.Create(data);
 
-            Report report = await Client.Report.Create(Fixtures.ReportType, parameters);
+            Report report = await Client.Report.Create(parameters);
 
             Assert.IsType<Report>(report);
             Assert.StartsWith(Fixtures.ReportIdPrefix, report.Id);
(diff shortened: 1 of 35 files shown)


EasyPost.Tests/BetaFeaturesTests/ServicesTests/ReportServiceTest.cs
```diff
@@ -85,11 +88,12 @@ public async Task TestCreateWithColumns()
                 { "columns", columns },
                 { "start_date", Fixtures.ReportDate },
                 { "end_date", Fixtures.ReportDate },
+                { "type", "shipment" },
```
Justintime50, line 91: question: why make this change? Does it better align with the other libs?

EasyPost/Services/UserService.cs
```diff
@@ -40,26 +41,26 @@ public async Task<User> CreateChild(Dictionary<string, object> parameters)
...
-        public async Task<User> CreateChild(BetaFeatures.Parameters.Users.CreateChild parameters)
+        public async Task<User> CreateChild(BetaFeatures.Parameters.Users.CreateChild parameters, CancellationToken cancellationToken = default)
         {
             // Because the normal CreateChild method does wrapping internally, we can't simply pass the parameters object to it, otherwise it will wrap the parameters twice.
-            return await RequestAsync<User>(Method.Post, "users", parameters.ToDictionary());
+            return await RequestAsync<User>(Method.Post, "users", cancellationToken, parameters.ToDictionary());
         }
 
         /// <summary>
-        ///     Retrieve a User from its id. If no id is specified, it returns the user for the api_key specified.
```
Justintime50, line 50: issue: I don't think we want to remove this functionality. Users have come to rely on it and our docs clearly state this is a reasonable thing to do.

Justintime50 (commented):
Let's not optimize too much more. We are getting back into the weeds here a bit.

jchen293 (approved):

EasyPost.Tests/BetaFeaturesTests/ServicesTests/ReportServiceTest.cs
nwithan8: Helps make the Parameters set consistent. This was the only function that takes an explicit parameter + a Parameters object.

EasyPost/Services/UserService.cs
nwithan8: Fair enough, reverting.

Justintime50 (approved):

PR #458, opened 2023-05-04
[chore] Models are immutable

nwithan8:
# Description

- All setters for model properties are internal (end-user cannot modify any property of an object outside of an Update API call)

# Testing

- No changes to unit tests
  - Test suite has internal access to source code, allowing it to use internal setters as needed during testing. End-users importing this library will not have internal access, and therefore won't be able to access the setters.

# Pull Request Type


nwithan8:
Decided against this approach.

PR #459, opened 2023-05-05
[chore] Docstrings

nwithan8:
# Description

- All classes have docstrings
- All methods have docstrings
- All parameter sets have docstrings
- Large chuck of model properties have docstrings
- Minor cleanup
  - Add missing class for carrier account `Fields` property
  - Remove unused imports
  - Remove deprecated lowest rate function in Rate service (users should use the function in `Utils` instead)

# Testing

- Two unit tests re-recorded due to outdated cassettes
- Unit test for deprecated rate function removed
- All other tests pass as expected with no modification

# Pull Request Type

Please select the option(s) that are relevant to this PR.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

PR #460, opened 2023-05-08
[chore] Docstrings, part 1

nwithan8:
# Description

Part 1 of https://github.com/EasyPost/easypost-csharp/pull/459

Everything except models and services now has docstrings

# Testing

- Re-recorded cassettes as needed

# Pull Request Type


EasyPost/BetaFeatures/Parameters/Addresses/Create.cs
```diff
@@ -4,21 +4,28 @@
...
         [NestedRequestParameter(typeof(Shipments.Create), Necessity.Optional, "id")]
         [NestedRequestParameter(typeof(Insurance.Create), Necessity.Optional, "id")]
         [NestedRequestParameter(typeof(Orders.Create), Necessity.Optional, "id")]
         [NestedRequestParameter(typeof(Pickups.Create), Necessity.Optional, "id")]
         [NestedRequestParameter(typeof(Beta.Rates.Retrieve), Necessity.Optional, "id")]
         public string? Id { get; set; }
 
+        /// <summary>
+        ///     Carrier facility for the new <see cref="Models.API.Address"/>.
+        /// </summary>
```
Justintime50, line 28: thought: as suggested initially, I feel that documenting all the properties is taking documentation to the extreme. Most should be self explanatory, would they even pop up in an IDE?

Documenting functions and classes makes sense because they have logic associated that may need additional explanations where properties should be exactly the name given to them.

Thoughts?

EasyPost/BetaFeatures/Parameters/ReferralCustomers/All.cs
```diff
@@ -6,7 +6,7 @@
 namespace EasyPost.BetaFeatures.Parameters.ReferralCustomers
 {
     /// <summary>
-    ///     Parameters for <see cref="EasyPost.Services.PartnerService.All(All)"/> API calls.
+    ///     <a href="https://www.easypost.com/docs/api#retrieve-a-list-of-referral-customers">Parameters</a> for <see cref="EasyPost.Services.PartnerService.All(All, System.Threading.CancellationToken)"/> API calls.
```
Justintime50, line 9: question: what is Roslyn on about here?

EasyPost/BetaFeatures/Parameters/Addresses/Create.cs
nwithan8: Yes, they would pop up in an IDE, definitely.

As a user, I might not know what some of these things are, or when/if I should set them to specific values.

EasyPost/BetaFeatures/Parameters/ReferralCustomers/All.cs
nwithan8: Fixed

PR #463, opened 2023-05-10
[chore] Move Parameters out of beta

nwithan8:
# Description

Our `Parameters` set objects are out of beta. This PR moves them from `EasyPost.BetaFeatures.Parameters` to `EasyPost.Parameters`

# Testing

- Unit tests updated, consolidated as needed, no cassettes need re-recording

# Pull Request Type


EasyPost.Tests/Fixture.cs
```diff
@@ -138,11 +143,11 @@ internal static class Parameters
         {
             internal static class Addresses
             {
-                internal static BetaFeatures.Parameters.Addresses.Create Create(Dictionary<string, object>? fixture)
+                internal static Create Create(Dictionary<string, object>? fixture)
```
Justintime50, line 146: question: how is this getting away with just `Create`? Shouldn't it be `Addresses.Create`? 

question: Can we include the type of object to be more explicit? This is true of the whole PR, it would be easier to follow if done that way

question: Do we want to do singular vs plural to match services?

Justintime50 (commented):
Overall seems fine, have three questions though I'd like to explore

EasyPost.Tests/Fixture.cs
nwithan8: 1. My IDE cleaned up the imports during namespace adjustments. `Create` and `All` is all that's needed because this file imports the `EasyPost.Parameters.Address` namespace.

2. Because of 1 above, it shortened the namespace. We can certainly go back and make explicit namespaces (I'd be concerned the IDE would constantly be trying to auto-optimize them in the future, however)

3. I'll defer, noting that it would be a breaking change.

jchen293: We use singular in other libs. Since this is already a major bump, we should consider making it singular to match the service.

Justintime50: As for item 1 I guess it's confusing because it's a fixture file with lots of different objects in one file, how does it know it's an Address.Create if it just says Create. 

For item 2, I'd prefer to be explicit and see if we can't override that so IDEs don't freak out. I'd be concerned with mixing namespaces for items in the same file

nwithan8: 1. If you concern was specifically for the Fixture file, already resolved.
2. Fine

nwithan8: Resolved

EasyPost.Tests/ParametersTests/ParametersTest.cs
```diff
@@ -71,13 +72,13 @@ public void TestParametersToDictionaryWithSubDictionary()
                 Street1 = streetA,
             };
 
-            EasyPost.BetaFeatures.Parameters.Addresses.Create newAddressCreationParameters = new EasyPost.BetaFeatures.Parameters.Addresses.Create
+            Create newAddressCreationParameters = new Create
```
Justintime50, line 75: issue: this should be the full namespace here

EasyPost.Tests/Fixture.cs
Justintime50: I'm still seeing a lot of items without the full namespace, by chance were some missed? There's a lot in here heh

EasyPost.Tests/ParametersTests/ParametersTest.cs
nwithan8: Oh, right, the unit tests...

EasyPost/packages.lock.json
```diff
@@ -7,6 +7,20 @@
         "requested": "[13.0.1, 14.0.0)",
         "resolved": "13.0.1",
         "contentHash": "ppPFpBcvxdsfUonNcvITKqLl3bqxWbDCZIzDWHzjpdAHRFfZe0Dw9HmA0+za13IdyrgJwpkDTDA9fHaxOrt20A=="
+      },
```
Justintime50, line 10: issue: these really need to stop showing up in your PRs 😉

Justintime50 (approved):
Looks good, please update the README to indicate it's no longer in beta

jchen293 (approved):
lgtm, need to add a changelog for changing plural to singular but won't block it.

PR #478, opened 2023-05-18
[POC] Audit HTTP requests

nwithan8:
# Description

- New option when configuring a client, allowing users to define an action to call when making an HTTP request. The pre-flight request will be passed into the action, allowing users to view/audit every HTTP request made by the client. Users could then interact with the request as they wish (e.g. logging)

# Testing

- New unit test to confirm action is called during HTTP request execution
- Manual debugging confirms the `HttpRequestMessage` object is complete at this point (shouldn't be missing any headers or data, for example)
- Manually confirmed that modifying the `HttpRequestMessage` in the auditor function does NOT affect the executing request (audit function is effectively read-only)

# Pull Request Type

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ All API-calling functions are made from the appropriate service object (rather t
 Shipment myPurchasedShipment = await myClient.Shipment.Buy(myShipment.Id, myShipment.LowestRate());
 ```
 
-## Parameters
+### Parameters
 
 Most functions in this library accept a `Dictionary<string, object>` as their sole parameter, which is ultimately used as the body of the HTTP request against EasyPost's API. If you instead would like to use .NET objects to construct API call parameters, you can use the various `Parameters` classes (currently in beta).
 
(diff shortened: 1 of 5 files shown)


jchen293 (approved):

Justintime50 (commented):
Would love to see a README entry documenting a simple "how to" use this feature since it won't land in our API docs

README.md
```diff
@@ -173,6 +173,30 @@ Using the `Parameters` classes is not required, but they can help in a number of
...
+
+```csharp
+void myRequestAuditingAction(HttpRequestMessage request)
+{
+    // Interact with the HttpRequestMessage here
+}
+
+Client client = new Client(new ClientConfiguration("EASYPOST_API_KEY")
+{
+    RequestAuditor = myRequestAuditingAction,
```
Justintime50, line 194: thought: I'd love to see a dead simple real example. If I'm looking at this, I wouldn't know what I could do. Can we make it as simple as printing to console in the example so people know exactly how to access the request?

Justintime50 (commented):
Seems fine, though can we elicit feedback from the greater team (and support) on the design? What we don't want to do is full send this when they need more out of it or something different

Justintime50 (changes requested):
Waiting to hear back from everyone.

nwithan8:
Closing for now. If we re-open this, reference the [`IntrospectiveClient` in EasyPost.Extensions](https://github.com/nwithan8/easypost-extensions-dotnet/blob/master/EasyPost.Extensions/Clients/IntrospectiveClient.cs)

PR #489, opened 2023-06-27
[feat] Callbacks (hooks) to audit pre-flight HTTP requests and post-flight HTTP responses

nwithan8:
# Description

- New optional Hooks object to include in ClientConfiguration to define pre- and post-flight HTTP callbacks
  - Event handlers for callbacks follow [.NET naming conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events).
- New README example for using pre- and post-flight HTTP callbacks
- Ignore invalid virtual warning

# Testing

- New unit test to verify that pre- and post-flight hooks are called correctly.
- New unit test to verify that multiple pre- and multiple post-flight hooks can be called (.NET event handlers can work with multiple callbacks natively)
- New unit test using pre-flight hook to test cancellation token for HTTP request.

# Pull Request Type


EasyPost/Hooks.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// <param name="guid">A unique identifier for the HTTP request-response pair.</param>
+    internal RequestBeforeExecutionEventArgs(HttpRequestMessage request, int timestamp, Guid guid)
+    {
+        Request = request;
+        Timestamp = timestamp;
+        Id = guid;
+    }
+}
+
+public class RequestResponseReceivedEventArgs : EventArgs
```
Justintime50, line 50: issue: missing `method`, `path`, `headers`. If they are in the response param, we need to break them out

README.md
```diff
@@ -173,6 +173,34 @@ Using the `Parameters` classes is not required, but they can help in a number of
 - Provides a more natural way to construct parameters
 - Facilitates ASP.NET Core model binding (bind an HTML form to a `Parameters` instance)
 
+## Debugging
+
+### View HTTP Requests
```
Justintime50, line 178: suggestion: we should call this `HTTP Hooks` as the header and describe that they can view the requests in the details below. Hooks is a more universal language to use here and not everyone may want to view it, they may just want to be notified when requests/responses occur and do something totally different.

Additionally, this should probably go under the `usage` section instead as a sub section.

README.md
```diff
@@ -173,6 +173,34 @@ Using the `Parameters` classes is not required, but they can help in a number of
...
+
+Users can audit the HTTP requests and responses being made by the library by setting the `Hooks` property of a `ClientConfiguration` with a set of event handlers. Available handlers include:
+
+- `OnRequestBeforeExecution` - Called before an HTTP request is made. An `RequestBeforeExecutionEventArgs` object is passed to the handler, which contains the `HttpRequestMessage` that will be sent to the server.
```
Justintime50, line 182: issue: per the design doc and where possible, we should try to keep the language close to `RequestHook`. We can then document that it fires before a request happens. These names are pretty verbose.

README.md
```diff
@@ -173,6 +173,34 @@ Using the `Parameters` classes is not required, but they can help in a number of
...
+  - The `HttpRequestMessage` at this point is configured with all expected data (headers, body, etc.). Modifying the `HttpRequestMessage` will NOT affect the actual request that is sent to the server.
+- `OnRequestResponseReceived` - Called after an HTTP request is made. An `RequestResponseReceivedEventArgs` object is passed to the handler, which contains the `HttpResponseMessage` that was received from the server.
```
Justintime50, line 184: issue: per the design doc and where possible, we should try to keep the language close to `ResponseHook`. We can then document that it fires after a response is received. These names are pretty verbose.

README.md
```diff
@@ -173,6 +173,34 @@ Using the `Parameters` classes is not required, but they can help in a number of
...
+{
+    Hooks = new Hooks {
+        OnRequestBeforeExecution = (_, args) => {
+            // Interact with the HttpRequestMessage here via args.Request
+            System.Console.WriteLine($"Making HTTP call to {args.Request.RequestUri}");
+        },
+        OnPostRequest = (_, args) => {
+            // Interact with the HttpResponseMessage here via args.Response
+            System.Console.WriteLine($"Received HTTP response with status code {args.Response.StatusCode}");
+        },
```
Justintime50, line 199: I'd suggest in the example to define these separately and then pass them into the hooks as I can see most people not doing this inline as the logic could be long. Would make this a bit cleaner

Justintime50 (dismissed):

README.md
nwithan8: This is standard with how event/delegate verbiage typically works in .NET.

EasyPost.Tests/ClientTest.cs
```diff
@@ -104,5 +107,71 @@ public void TestHttpClientOverride()
...
+                },
+            };
+
+            var client = new Client(new ClientConfiguration(FakeApikey)
+            {
+                Hooks = hooks,
+            });
+
+            // Make a request, doesn't matter what it is (catch the exception due to invalid API key)
+            await Assert.ThrowsAsync<UnauthorizedError>(async () => await client.Address.Create(new Parameters.Address.Create()));
```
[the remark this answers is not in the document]
nwithan8: Resolved

README.md
nwithan8: Resolved

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    /// <summary>
+    ///     The <see cref="HttpRequestMessage"/> about to be executed by the HTTP request.
+    /// </summary>
+    public HttpRequestMessage Request { get; }
+
+    /// <summary>
+    ///     The timestamp of the HTTP request.
+    /// </summary>
+    public int Timestamp { get; }
```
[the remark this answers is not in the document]
nwithan8: Resolved

README.md
Justintime50: Thanks, I hate it

EasyPost.Integration/Basics.cs
```diff
@@ -86,4 +86,54 @@ public void TestUserCanLocallyConstructResponseObject()
...
+    {
+        // Can set up each hook event handler during construction
+        var hooks = new Hooks()
+        {
+            OnRequestResponseReceived = new EventHandler<OnRequestResponseReceivedEventArgs>((sender, args) =>
+            {
+                Console.WriteLine(args.Response);
+                Console.WriteLine(args.RequestTimestamp);
+                Console.WriteLine(args.ResponseTimestamp);
+                Console.WriteLine(args.Id);
```
Justintime50, line 105: issue: these should not be writing out. We should be asserting like a unit test. Integration tests are meant to be automated as are unit tests but they test the integration of the parts. Having it print lines like this requires user intervention which I promise we will stop doing 2 weeks from now when we forget this does that.

EasyPost/Hooks.cs
nwithan8: Resolved, following spec.

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// </summary>
+    public EventHandler<RequestBeforeExecutionEventArgs>? OnRequestBeforeExecution { get; set; }
+
+    /// <summary>
+    ///     An <see cref="EventHandler{RequestResponseReceivedEventArgs}"/> to view an HTTP response received by the client.
+    /// </summary>
+    public EventHandler<RequestResponseReceivedEventArgs>? OnRequestResponseReceived { get; set; }
+}
+
+public class RequestBeforeExecutionEventArgs : EventArgs
```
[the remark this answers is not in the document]
nwithan8: Resolved, following spec.

.github/workflows/ci.yml
```diff
@@ -147,8 +147,9 @@ jobs:
       - name: Install .NET SDK
         uses: actions/setup-dotnet@v3
         with:
-          # .NET 5 is deprecated and removed from GitHub Actions, we need to manually install it
+          # .NET 3.1 and 5 are deprecated and removed from GitHub Actions, we need to manually install them
```
dcaballeroc, line 150: Do we have a plan to stop supporting it in the future?

dcaballeroc (commented):
LGTM

I'm seeing lots of obsolete class warnings. Do we want to address those in a future PR?

nwithan8:
> LGTM
> 
> I'm seeing lots of obsolete class warnings. Do we want to address those in a future PR?

Those will be removed in a future major release per our client library deprecation policy.

.github/workflows/ci.yml
nwithan8: We do have a deprecation policy of "LTS + 1", which would technically make .NET 5 obsolete. However, we haven't moved on that yet.

dcaballeroc:
> > LGTM
> > I'm seeing lots of obsolete class warnings. Do we want to address those in a future PR?
> 
> Those will be removed in a future major release per our client library deprecation policy.

Oooh, can I see that?

dcaballeroc (approved):

nwithan8:
> > > LGTM
> > > I'm seeing lots of obsolete class warnings. Do we want to address those in a future PR?
> > 
> > 
> > Those will be removed in a future major release per our client library deprecation policy.
> 
> Oooh, can I see that?

For historical record: https://www.easypost.com/blog/2022-08-04-lessons-learned-overhauling-client-libraries-across-seven-languages#:~:text=We%20determined%20the%20best%20option%20was%20to%20start%20dropping%20support%20for%20unsupported%20language%20versions%20until%20the%20oldest%20version%2C%20minus%20one%20(where%20possible)

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// <summary>
+    ///     The <see cref="HttpMethod"/> of the HTTP request.
+    /// </summary>
+    public HttpMethod RequestMethod => Request.Method;
+
+    /// <summary>
+    ///     The <see cref="Uri"/> of the HTTP request.
+    /// </summary>
+    public Uri? RequestUri => Request.RequestUri;
```
jchen293, line 38: Question: why are there `?`

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// </summary>
+    public EventHandler<OnRequestExecutingEventArgs>? OnRequestExecuting { get; set; }
+
+    /// <summary>
+    ///     An <see cref="EventHandler{OnRequestResponseReceivedEventArgs}"/> to view an HTTP response received by the client.
+    /// </summary>
+    public EventHandler<OnRequestResponseReceivedEventArgs>? OnRequestResponseReceived { get; set; }
+}
+
+public class OnRequestExecutingEventArgs : EventArgs
```
jchen293, line 23: We should rename these attribute names of the class to what is listed on the tech design.

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <param name="guid">A unique identifier for the HTTP request-response pair.</param>
+    internal OnRequestExecutingEventArgs(HttpRequestMessage request, int timestamp, Guid guid)
+    {
+        Request = request;
+        RequestTimestamp = timestamp;
+        Id = guid;
+    }
+}
+
+public class OnRequestResponseReceivedEventArgs : EventArgs
```
jchen293, line 74: We should rename these attribute names of the class to what is listed on the tech design.

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// </summary>
+    public Guid Id { get; }
+
+    /// <summary>
+    ///     Constructs a new instance of the <see cref="OnRequestExecutingEventArgs"/> class.
+    /// </summary>
+    /// <param name="request">The <see cref="HttpRequestMessage"/> about to be executed by the HTTP request.</param>
+    /// <param name="timestamp">The timestamp of the HTTP request.</param>
+    /// <param name="guid">A unique identifier for the HTTP request-response pair.</param>
+    internal OnRequestExecutingEventArgs(HttpRequestMessage request, int timestamp, Guid guid)
```
jchen293, line 66: Why does the constructor only have three parameters?

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+    public Guid Id { get; }
+
+    /// <summary>
+    ///     Constructs a new instance of the <see cref="OnRequestResponseReceivedEventArgs"/> class.
+    /// </summary>
+    /// <param name="response">The <see cref="HttpResponseMessage"/> returned by the HTTP request.</param>
+    /// <param name="requestTimestamp">The timestamp of the HTTP request.</param>
+    /// <param name="responseTimestamp">The timestamp of the HTTP response.</param>
+    /// <param name="guid">A unique identifier for the HTTP request-response pair.</param>
+    internal OnRequestResponseReceivedEventArgs(HttpResponseMessage response, int requestTimestamp, int responseTimestamp, Guid guid)
```
jchen293, line 128: Why does the constructor only have four parameters?

README.md
```diff
@@ -173,6 +173,39 @@ Using the `Parameters` classes is not required, but they can help in a number of
...
+    System.Console.WriteLine($"Received HTTP response with status code {args.ResponseStatusCode}");
+}
+            
+Client client = new Client(new ClientConfiguration("EASYPOST_API_KEY")
+{
+    Hooks = new Hooks {
+        OnRequestExecuting = OnRequestExecutingHandler,
+        OnRequestResponseReceived = OnRequestResponseReceivedHandler,
+    },
+});
```
jchen293, line 206: Nit: can we add how to unsubscribe from the request/response hooks?

jchen293 (dismissed):

EasyPost/Hooks.cs
nwithan8: Can be null

EasyPost/Hooks.cs
nwithan8: Object consumes an `HttpRequestMessage` object and stores it privately, all expected properties are derived from it internally.

EasyPost/Hooks.cs
nwithan8: Object consumes an `HttpResponseMessage` object and stores it privately, all expected properties are derived from it internally.

EasyPost/Hooks.cs
jchen293: The URI should never be null when making an API request, can we remove this?

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// <summary>
+    ///     The timestamp of the HTTP request.
+    /// </summary>
+    public int RequestTimestamp { get; }
+
+    /// <summary>
+    ///     A unique identifier for the HTTP request-response pair.
+    /// </summary>
+    public Guid Id { get; }
```
jchen293, line 58: This need to be renamed to `RequestUuid`

EasyPost/Hooks.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// <summary>
+    ///     The timestamp of the HTTP response.
+    /// </summary>
+    public int ResponseTimestamp { get; }
+
+    /// <summary>
+    ///     A unique identifier for the HTTP request-response pair.
+    /// </summary>
+    public Guid Id { get; }
```
jchen293, line 119: Same thing here, let's rename it to `RequestUuid`

README.md
```diff
@@ -173,11 +173,57 @@ Using the `Parameters` classes is not required, but they can help in a number of
...
+  - The `HttpRequestMessage` at this point is configured with all expected data (headers, body, etc.). Modifying
+    any data in the callback will NOT affect the actual request that is sent to the server.
+- `OnRequestResponseReceived` - Called after an HTTP request is made. An `RequestResponseReceivedEventArgs` object is
+  passed to the handler, which contains details about the `HttpResponseMessage` that was received from the server.
+
+Users can interact with these details in their callbacks as they see fit (e.g. logging).
+
+```csharp
+void OnRequestExecutingHandler(object? sender, OnRequestExecutingEventArgs args) {
+    // Interact with the HttpRequestMessage here via args
```
jchen293, line 191: I would modify this comment a little bit since the args have more than the `HttpRequestMessage` attribute. Same thing down below

jchen293 (changes requested):
Can we add a test that no hook is called after unsubscribing? Similar to what we have in Python: https://github.com/EasyPost/easypost-python/blob/master/tests/test_hook.py#L41

EasyPost/Hooks.cs
nwithan8: Not without the compiler complaining. We know it will never be null, but by the typing of the underlying `HttpRequestMessage` class, it COULD be null.
image

EasyPost/Hooks.cs
nwithan8: Pushing back against this one, since it's not really a request Uuid, it's a GUID of a request-response pair.

EasyPost.Tests/ClientTest.cs
```diff
@@ -184,6 +184,45 @@ public async Task TestMultipleRequestHookCallbacks()
...
+            Assert.Equal(1, preRequestCallbackCallCount);
+
+            // Unsubscribe from the pre-request callback
+            Client.Hooks.OnRequestExecuting -= PreRequestCallback;
+
+            // Make a request, doesn't matter what it is (catch the exception due to invalid API key)
+            await Assert.ThrowsAsync<UnauthorizedError>(async () => await Client.Address.Create(new Parameters.Address.Create()));
+
+            // Assert that the pre-request callback was not called again
+            Assert.Equal(1, preRequestCallbackCallCount);
```
jchen293, line 223: Shouldn't this be `0` if the called wasn't called?

jchen293 (commented):
One last question, then I will stamp

EasyPost.Tests/ClientTest.cs
nwithan8: No, it was called once, but not called a second time.

jchen293: Ah, i see

jchen293 (approved):

PR #491, opened 2023-07-11
[feat] Add ability to add custom options

nwithan8:
# Description

- Adds ability for users to provide additional option key-value pairs in case an option is not available in the library

# Testing

- Add unit test to confirm that additional, user-defined options will be included in final serialized JSON payload

# Pull Request Type

### DIFF
--- a/EasyPost/Models/API/Options.cs
+++ b/EasyPost/Models/API/Options.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using EasyPost._base;
+using EasyPost.Utilities.Internal.Extensions;
 using Newtonsoft.Json;
 
 namespace EasyPost.Models.API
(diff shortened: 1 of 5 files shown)


Justintime50:
Do not introduce breaking changes, even if they are fixes. They require a major version bump. Leave them as they are, mark them with TODO comments, open a task, and fix them when we do the next major release. People rely and have coded around the inconsistencies. "Fixing" them will still break integrations.

dcaballeroc (dismissed):

EasyPost/Models/API/Options.cs
```diff
@@ -262,26 +298,40 @@ public class Options : EasyPostObject
...
 
         /// <summary>
-        ///     The latest a package should be delivered.
+        ///     The earliest a package should be delivered.
         ///     Supported carriers vary.
         /// </summary>
-        [JsonProperty("delivery_max_datetime")]
-        public string? DeliveryMaxDatetime { get; set; }
+        [JsonProperty("delivery_min_datetime")]
+        public DateTime? DeliveryMinDatetime { get; set; } // datetime
```
Justintime50, line 318: question: are we sure we want datetimes? wasn't there just recently a bug report where this was not acting as a user expected?

dcaballeroc: IIRC the bug was that it was causing problems cuz it was a string instead of a DateTime

dcaballeroc (approved):

Issue #499, opened 2023-08-04
[Bug]: Deadlock when blocking on async methods


johnnyh-tactics:
### Software Version

v5.3.0

### Language Version

.NET Framework 4.8, Version 10.0.22621 Build 22621

### Operating System

Microsoft Windows 11 Pro

### What happened?

Hi,

I have been upgrading our system from using version 2.8.2 of the EasyPost NuGet Package to the most recent version 5.3.0 so that we can have access to some newer EasyPost features. One of the major changes between those versions is that all of the methods that make API calls are now async. Our calling code is primarily synchronous. Usually when an SDK is async-focused and I need it to work synchronously I just access the `Result` property on the `Task` that is returned from the async method (or similarly, call `GetAwaiter().GetResult()`). Unfortunately, when I do that with the new EasyPost SDK, it deadlocks and never completes.

I understand the preferred approach is to fully embrace the async/await pattern, and I see related issues in this repo where using `await` is the solution to the issue.  In my situation, I can't just use the await keyword because that would mean I have to make my encapsulating method async as well, which would just propagate the problem outward and affect parts of our system that aren't related to EasyPost at all. I suspect this isn't a problem specific to me, especially considering that this SDK was fully synchronous up until version 3 which is only 16 months old, so many existing customers are already using it synchronously.

I see usages `ConfigureAwait(false)` in some of the `await` calls in your source code, but also many `await` calls that do not have it. I suspect using `ConfigureAwait(false)` universally for `await` calls could prevent the deadlocking.

### What was expected?

Calling `.GetAwaiter().GetResult()` or `.Result` on any async method in the SDK to block main synchronous control flow only until the async action completes. 

### Sample Code

```csharp
public class TestController : System.Web.Mvc.Controller
{
    // this action completes successfully
    public async Task<ActionResult> TestEasyPostGetOrderAsync()
    {
        var order = await Client.Order.Retrieve("order_***");
        var timestampUpdated = order.UpdatedAt.ToString();
        return Content(timestampUpdated);
    }

    // this action deadlocks
    public ActionResult TestEasyPostGetOrder()
    {
        var retrieveOrderTask = Client.Order.Retrieve("order_***");
        // this line stalls indefinitely
        var order = retrieveOrderTask.Result;
        var timestampUpdated = order.UpdatedAt.ToString();
        return Content(timestampUpdated);
    }

    // this action deadlocks
    public ActionResult TestEasyPostGetOrder()
    {
[...]

PR #500, opened 2023-08-07
[chore] Add integration test for sync/async support

nwithan8:
# Description

- Add integration tests verifying that asynchronous functions can be run synchronously via multiple methods (re: #499)
- Add VCR, associated utilities for integration test suite

# Testing

- New integration tests, cassettes for async/sync tests

# Pull Request Type

### DIFF
--- a/EasyPost.Integration/TestUtils.cs
+++ b/EasyPost.Integration/TestUtils.cs
@@ -1,0 +1,174 @@
+using System.Runtime.CompilerServices;
+using EasyVCR;
+
+// ReSharper disable once CheckNamespace
+namespace EasyPost.Integration.Utilities
+{
+    public class Utils
+    {
+        internal const string ApiKeyFailedToPull = "couldnotpullapikey";
+
+        private static readonly List<string> BodyCensors = new()
+        {
+            "api_keys",
+            "children",
+            "client_ip",
+            "credentials",
+            "email",
+            "key",
+            "keys",
+            "phone_number",
+            "phone",
+            "test_credentials"
+        };
+
+        private static readonly List<string> HeaderCensors = new()
+        {
+            "Authorization",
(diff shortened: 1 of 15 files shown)


EasyPost.Integration/cassettes/net/synchronous/async.json
```diff
@@ -0,0 +1,101 @@
+[
```
Justintime50, line 1: thought: can we get these to be treated as binary like the other cassettes so they aren't expanded in reviews?

nwithan8: Resolved

EasyPost.Integration/TestUtils.cs
```diff
@@ -0,0 +1,174 @@
+using System.Runtime.CompilerServices;
```
Justintime50, line 1: question: can we not reuse the config from the unit tests here to avoid duplicating this?

Justintime50 (approved):
Nice work! Love extra tests to verify various usage patterns

EasyPost.Integration/TestUtils.cs
nwithan8: Unfortunately not, because of how this file looks up the location of cassettes. If used, cassettes for integration tests would be stored in the cassettes folder under Tests rather than Integrations

nwithan8: I can explore making a location-agnostic version of test utils in the future.

Justintime50: Cool, I'd love to revisit if it's possible. Dynamically supplying the cassette location so we can keep the other same config pieces DRY would be good.

jchen293 (approved):

Issue #501, opened 2023-08-07
[Feat]: Specify EasyPost.Parameters.Order.Create.CarrierAccounts by Id


johnnyh-tactics:
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

Hello again,

During my migration from EasyPost SDK v2.8.2 to v5.3.0, I've encountered a situation where an important feature existed in the v2.8.2 but is missing now.

In v2.8.2, when rating an order, I could do something like the following to get rates back for just the carrier accounts specified. 

```
new EasyPost.Order
{
    <other properies>...,
    carrier_accounts = CarrierAccountIds.Select(id => new CarrierAccount { id = id }).ToList()
}.Create();
```

where `CarrierAccountIds` is just an array of carrier account Id strings like `new[] {"ca_...", "ca_...", "ca_..."}`.

In v5.3.0, the `Id` property on `EasyPost.Models.API.CarrierAccount` has `internal` access only for setting, so this same approach is not available. As a workaround, I used the the SDK to get all CarrierAccounts, then filtered them to just the ones I want, e.g.

```
var carrierAccountsFromSDK = await Client.CarrierAccount.All();
var filteredCarrierAccounts = carrierAccountsFromSDK.Where(c => CarrierAccountIds.Contains(c.Id)).ToList();

await Client.Order.Create(new EasyPost.Parameters.Order.Create
{
    <other properies>...,
    CarrierAccounts = filteredCarrierAccounts
});
```

Unfortunately, that only works with production credentials. When testing with sandbox/test credentials, I get `EasyPost.Exceptions.API.UnauthorizedError: 'This resource requires a production API Key to access.'`. In theory I could use production credentials for this just this one API call, but I'd really rather keep production credentials out of my development and testing environment completely. Being able to specify carrier acount Ids without having the extra API call would sidestep the resource access issue.

nwithan8:
Hello again! Thanks for letting us know; we'll have to discuss the best way to proceed. We made `Id` internally-set only to avoid users accidentally modifying it, but it seems we've encountered an edge-case here. We'll let you know what we come up with!

nwithan8:
Hey @johnnyh-tactics , wanted to keep you updated.

After discussing internally, we decided that our locking down of the `Id` property was an over-reach, so we will be making the setter for the `Id` of any API object public.

Once done, you'll be able to accomplish your API call with something along the lines of:

```csharp
[...]

PR #502, opened 2023-08-08
[fix] Make `Id` property of all EasyPostObject publicly settable

nwithan8:
# Description

- Make `Id` property of all EasyPostObject public (re #501 )

# Testing

- Add integration test to verify users can set Id of an object
- Remove "Test" prefix from integration tests

# Pull Request Type

### DIFF
--- a/EasyPost.Integration/Basics.cs
+++ b/EasyPost.Integration/Basics.cs
@@ -91,7 +91,7 @@ public class Basics
     ///     Test that an end-user can locally construct an object and set its ID.
     /// </summary>
     [Fact, Testing.Access, Testing.Compile]
-    public void UsersCanSetObjectId()
+    public void UserCanSetObjectId()
     {
         // Construct a local object, setting its ID
         var address = new Address { Id = "some_id" };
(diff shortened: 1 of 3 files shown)


EasyPost.Integration/Basics.cs
```diff
@@ -11,7 +11,7 @@ public class Basics
     ///     If this test can be compiled, then the response objects have public constructors.
     /// </summary>
     [Fact, Testing.Access, Testing.Compile]
-    public void TestUserCanLocallyConstructResponseObject()
+    public void UserCanLocallyConstructResponseObject()
```
Justintime50, line 14: question: why remove the `Test` prefix? Keeping it quickly helps scanning eyes know exactly what it is

nwithan8: They're all tests, so it becomes a redundant word

Justintime50: We'll agree to disagree

jchen293 (approved):

Issue #365, opened 2022-11-01
Add Ability to Register Carriers with Custom Flows


datelcry:
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

Hello EasyPost team!

There is a request for a little but valuable improvement of CarrierAccountService.cs

The implementation of custom flows of registering carriers like Fedex/UPS requires us to write own code to reach your API endpoint despite the presence of official SDK
Your [article ](https://support.easypost.com/hc/en-us/articles/7469579510925-Adding-Carriers-with-Custom-Workflows) has the only example of curl request, but [CarrierAccountService](https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Services/CarrierAccountService.cs) doesn't have /carrier_accounts/register endpoint.

So to avoid monitoring your API for updates of this endpoint in the future and for better experience of using the API I ask you to add the register endpoint into your API

Thanks in advance!

TGKN:
Hi,
I'd like to second this request as I want to get rid of my custom code extensions for the easypost library.

nwithan8:
> Hi,
> I'd like to second this request as I want to get rid of my custom code extensions for the easypost library.

This feature was added in https://github.com/EasyPost/easypost-csharp/pull/375 and will be included in the next release, currently scheduled for later this week.

TGKN:
Hi,
I have switched to the version of the package in question and I am afraid it is not working as expected. Your code does abstract the different URLs that need to be called for FedEx and UPS but it does not take into consideration that I need to pass a differently structureed data. Instead of "credentials" I need to pass "registration_data". Everything else is the same. So if I still have to create a custom workflow for UPS and FedEx your fix did not really help me at all. Or am I missing something??
Greetz
Kai

Justintime50:
@TGKN that is correct. Based on the `type` field of the create/update request, we'll route the request to the correct endpoint. You will however need to pass in the right set of parameters (`credentials` for most carriers, `registration_data` for FedEx/UPS). Updating as you have and using the functions as normal should work correctly. You can view the tests as an example as we have both flows listed there.

nwithan8:
Hey @TGKN, just wanted to let you know that in v5.5.0, we've added custom parameter sets for FedEx and UPS (`CreateFedEx` and `CreateUps`) that will guide you through those unique parameter requirements.

PR #509, opened 2023-08-28
[feat] Add custom parameter sets for FedEx and UPS account creation

nwithan8:
# Description

Re: #365 

This PR introduces custom parameter sets for creating FedEx and UPS carrier accounts, as these two require different payload schemas than other carrier account types. These parameter sets (`CreateFedEx`, `CreateUps`) helps end-users know what parameters are required for these two specific carrier accounts compared to other carrier accounts, and handles elements such as parameter placement during serialization (standard for all parameter set objects) and auto-populating the carrier account type parameter.

These custom parameter sets can be used in the normal `Create` function. One step further, an exception will be raised if the user attempts to use the generic `Create` parameter set with, e.g. "FedExAccount" as its type (will force users to only use the `CreateFedEx` parameter set if they are attempting to make a FedEx carrier account (same for UPS)).

The parameter sets for FedEx and UPS have a number of required parameters that, if not completed, will raise a local exception during pre-request serialization (standard feature for parameter sets). This will additionally insure that users are not only using the proper parameter set for, eg., creating a FedEx account, but that all required data is being provided, all before the request is actually attempted.

# Testing

- New unit test for confirming new custom workflow parameter sets work versus generic ones
- New unit test to confirm user is barred from using generic parameter set for a custom-workflow-enabled carrier account
- New unit test to confirm missing required parameters raises an exception during serialization/validation.

# Pull Request Type

### DIFF
--- a/EasyPost/Services/CarrierAccountService.cs
+++ b/EasyPost/Services/CarrierAccountService.cs
@@ -68,6 +68,12 @@ namespace EasyPost.Services
                 throw new MissingParameterError(nameof(parameters.Type));
             }
 
+            // stop the user from using a generic Create parameter set when creating a carrier account with a custom workflow (e.g. FedExAccount)
+            if (parameters.GetType() == typeof(Parameters.CarrierAccount.Create) && Constants.CarrierAccountTypes.CarrierTypesWithCustomWorkflows.Contains(parameters.Type))
+            {
+                throw new InvalidParameterError("parameters", $"Use a {parameters.Type} custom workflow parameter set instead.");
+            }
+
             string endpoint = SelectCarrierAccountCreationEndpoint(parameters.Type);
 
             return await RequestAsync<CarrierAccount>(Method.Post, endpoint, cancellationToken, parameters.ToDictionary());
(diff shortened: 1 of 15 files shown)


dcaballeroc (approved):
Looks solid

jchen293 (approved):

EasyPost/Services/CarrierAccountService.cs
```diff
@@ -68,6 +68,12 @@ public async Task<CarrierAccount> Create(Parameters.CarrierAccount.Create parame
                 throw new MissingParameterError(nameof(parameters.Type));
             }
 
+            // stop the user from using a generic Create parameter set when creating a carrier account with a custom workflow (e.g. FedExAccount)
+            if (parameters.GetType() == typeof(Parameters.CarrierAccount.Create) && Constants.CarrierAccountTypes.CarrierTypesWithCustomWorkflows.Contains(parameters.Type))
```
Justintime50, line 72: thought: one thing we should consider is that CanadaPost is also a carrier that needs a custom workflow (same endpoint as old) - did we want to handle that in this PR too?

Justintime50 (approved):
Looks good!

EasyPost/Services/CarrierAccountService.cs
nwithan8: So it needs a custom parameter set, but needs to hit the generic endpoint, or both?

Justintime50: Disregard this actually, the endpoint and param key is the same - it's "custom" because you have to pass a special param to tell it not to authenticate via the browser but users can still use the function to do that without us needing to change anything.

Issue #377, opened 2022-11-26
Add equality overload for comparison of EasyPostObject and dictionary


nwithan8:
Per https://github.com/nopSolutions/nopCommerce/issues/6424#issuecomment-1327729733

> Sorry you had to make custom extensions to compare a dictionary of parameters to an existing EasyPostObject (e.g. Parcel, Shipment). We have a first-party way of comparing two EasyPostObjects, but especially since we removed the first-party ToDictionary function, there's no built-in way to do that via our library. Perhaps we can reintroduce that.

Justintime50:
Let's name this issue appropriately.

PR #511, opened 2023-08-30
[feat] Add ability to compare parameter set to an EasyPost object

nwithan8:
# Description

Re: #377 

Users can now define (and down the road we can add defaults) a `Match` function that evaluates if a parameter set object matches an EasyPost object. Some of our users have implemented their own version of this utility, to see if a prepared parameter set matches an existing object (e.g. if the address in a shipment parameter set matches the address of a given EasyPost object, they know they don't need to make a new object).

This PR:
- Add "Matches" function at base level of all parameter sets to test if a given set matches a provided EasyPostObject-based object
  - This can be implemented by library maintainers with "default" functions, or end-users can implement their own.

# Testing

- Add unit tests to confirm function can be implemented and works as expected

# Pull Request Type

### DIFF
--- a/EasyPost/Models/API/Batch.cs
+++ b/EasyPost/Models/API/Batch.cs
@@ -112,7 +112,7 @@ namespace EasyPost.Models.API
         /// <returns>A TParameters-type parameters set.</returns>
         protected internal override TParameters BuildNextPageParameters<TParameters>(IEnumerable<Batch> entries, int? pageSize = null)
         {
-            Parameters.Shipment.All parameters = Filters != null ? (Parameters.Shipment.All)Filters : new Parameters.Shipment.All();
+            Parameters.Batch.All parameters = Filters != null ? (Parameters.Batch.All)Filters : new Parameters.Batch.All();
 
             // TODO: Batches get returned in reverse order from everything else (oldest first instead of newest first), so this needs to be "after_id" instead of "before_id"
             parameters.AfterId = entries.Last().Id;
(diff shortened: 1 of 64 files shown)


EasyPost/Models/API/Batch.cs
```diff
@@ -112,7 +112,7 @@ public class BatchCollection : PaginatedCollection<Batch>
         /// <returns>A TParameters-type parameters set.</returns>
         protected internal override TParameters BuildNextPageParameters<TParameters>(IEnumerable<Batch> entries, int? pageSize = null)
         {
-            Parameters.Shipment.All parameters = Filters != null ? (Parameters.Shipment.All)Filters : new Parameters.Shipment.All();
+            Parameters.Batch.All parameters = Filters != null ? (Parameters.Batch.All)Filters : new Parameters.Batch.All();
```
Justintime50, line 115: whoops

EasyPost/Parameters/BaseParameters.cs
```diff
@@ -161,7 +174,7 @@ private void Add(RequestParameterAttribute requestParameterAttribute, object? va
                 // If the given value is another base-Parameters object, serialize it as a sub-dictionary for the parent dictionary
                 // This is because the JSON schema for a sub-object is different than the JSON schema for a top-level object
                 // e.g. the schema for an address in the address create API call is different than the schema for an address in the shipment create API call
-                case BaseParameters parameters:
+                case IBaseParameters parameters: // todo: if issues arise with this function, look at the type constraint on BaseParameters here
```
Justintime50, line 177: Easier to spot
```suggestion
                case IBaseParameters parameters: // TODO: if issues arise with this function, look at the type constraint on BaseParameters here
```

Justintime50 (approved):

Justintime50:
We'll want to remember to call out in the CHANGELOG that we fixed the Shipment -> Batch reference in addition to the new feature

dcaballeroc (approved):

Issue #512, opened 2023-08-31
[Bug]: Editing property values on parameters failing subsequent calls to paginated methods.


klkitchens:
### Software Version

5.5.0

### Language Version

C# .NET 6

### Operating System

Windows 10

### What happened?

1. Create a shipments.All parameters object (could be any paginated query, see below).
2. Call the Shipment.All method passing in the parameters object.
3. Process results.
4. Per API documentation, the results are 20 (default) - 100 (maximum) in count.   If there are more results available then the HasMore property is true and we are to call the API again using BeforeId property of the parameters object set to the ID of the last item in the results.
5. Set the BeforeId property (this was failing, so I was trying also to ignore the ID and just alter the dates and requerying).  I would set the EndDatetime property to the date (+1 second) of the last item in the result set (they come back in descending date order).
6. When calling Shipment.All method (using BeforeId, not altered dates), the second query would return the second batch just fine.  However, the 3rd batch and beyond was always the same as the second batch.  When calling Shipment.All (using the revised date range), the 2nd batch and beyond was always the same as the first.  Direct calls to the API confirmed that this should actually work and is not an issue with the API itself, but the library.

I downloaded the source and fitted it into my solution so I could step through the code itself and here's where the problem is (see below)

### What was expected?

Changing the parameter value SHOULD return the correct set of data each time.

### Sample Code

So the problem is here (line marked with ==>):

BaseParameters.cs Lines 216-228

```csharp
                    string soloKey = keys[0];
                    if (dictionary.ContainsKey(soloKey))
                    {
                        // Key-value pair already exists in dictionary (likely because of override parameters)
==>                        // Only change the value if the existing value is null
==>                        dictionary[soloKey] ??= value;
                    }
                    else
                    {
                        dictionary.Add(soloKey, value);
                    }

                    return dictionary;
```
When using BeforeId as noted, the first call, BeforeId is null (call works).  The second time, BeforeId is set to the proper value (call works).  The third time however, since BeforeId's previous value was not null, you're failing to set it to the new value.

When using a revised date range, the first call, EndDatetime has a value (call works).  The second time however, since EndDatetime previous value was not null, you're not setting it to the new passed in value.

I could try to fix the broken line, but not sure the underlying reasoning as to why that might have been there in the first place.  However, it certainly breaks the expected AND DOCUMENTED behavior of the API and thus the library.  The values should be revisable even from one value to another.

I suggest line 221 be corrected to the following: 

```csharp
[...]

PR #513, opened 2023-08-31
[fix] Fix parameter set serialization

nwithan8:
# Description

RE: #512 

Removes an unneeded null check (from an earlier implementation of parameter sets) that was causing reused parameter sets for GetNextPage functionality to not be updated correctly.

# Testing

- Add unit test to confirm GetNextPage parameter sets are getting updated properly behind-the-scenes by checking that multiple subsequent calls do in fact return different data.

# Pull Request Type


Justintime50 (approved):
praise: nice test!

jchen293 (approved):

PR #515, opened 2023-09-01
[chore, feat] Send FedEx SmartPost registration to correct endpoint, prep carrier account parameters

nwithan8:
# Description

- Fix FedEx SmartPost not going to the `/register` endpoint when creating a carrier account.
- Add custom parameter set for FedEx SmartPost (alongside FedEx and UPS)
- Add enums for common carrier account types under `CarrierAccountType`
  - Replaces `FedExAccount` and `UpsAccount` constants that were introduced only in the previous release (no concern for breaking change issue)
- Adds constructors for carrier account creation parameter sets that take in carrier account type as an argument
  - Old way (no argument) labeled as deprecated, as we want to move users towards defining the type during parameter construction
  - This will also come in handy if/when we introduce parameter sets for all carrier accounts
- Adjust the relationship between base classes for carrier account creation parameters to remove unnecessary parameters (e.g. `CreateFedEx` no longer has a useless `Credentials` property, `Create` no longer has a useless `RegistrationData` property)

# Testing

- Update unit test to account for changes to constants, re-record cassettes as needed
  - Minimal-to-no changes to unit tests suggest this will not be a breaking change for end-users (this alters unadvertised code that was only introduced a week ago anyway)

# Pull Request Type


dcaballeroc (approved):

EasyPost/Constants.cs
```diff
@@ -117,20 +118,22 @@ public static class ErrorMessages
         /// <summary>
         ///     Common carrier account type groups.
         /// </summary>
-        public static class CarrierAccountTypes
+        public static class CarrierAccounts
         {
-            public const string FedExAccount = "FedexAccount";
-
-            public const string UpsAccount = "UpsAccount";
```
Justintime50, line 124: issue: you mention "no concern for breaking change issue"; however, this is already in the wild and we can't remove it. For example, if someone has already updated to the last version and starts using these or updates to that version on their way to this upcoming one, this is a breaking change to have them released.

Please instead leave these here with a note to deprecate them in the next version (a TODO for easy searching) and we can remove them whenever we have another major release. It won't hurt anything to retain them. That way it's backwards compatible while still adding your new enums.

EasyPost/Models/API/CarrierAccountType.cs
```diff
@@ -0,0 +1,82 @@
+using System.Collections.Generic;
+using EasyPost.Utilities.Internal;
+
+namespace EasyPost.Models.API;
+
+public class CarrierAccountType : ValueEnum
+{
+    public static readonly CarrierAccountType AmazonMws = new CarrierAccountType(1, "AmazonMwsAccount");
```
Justintime50, line 8: suggestion: I would not go down this path. Carriers on the EasyPost platform are constantly coming and going. This adds much more maintenance to the project than we probably want as we've previously discussed - keep the libs dumb.

EasyPost/Parameters/CarrierAccount/Create.cs
```diff
@@ -8,7 +10,7 @@ namespace EasyPost.Parameters.CarrierAccount
     ///     <a href="https://www.easypost.com/docs/api#create-a-carrier-account">Parameters</a> for <see cref="EasyPost.Services.CarrierAccountService.Create(Create, System.Threading.CancellationToken)"/> API calls.
     /// </summary>
     [ExcludeFromCodeCoverage]
-    public class Create : BaseParameters<Models.API.CarrierAccount>, ICarrierAccountParameter
+    public class Create : ACreate
```
Justintime50, line 13: question: Why `ACreate`? Seems like a typo? If not can we instead define what this is better as `A` doesn't make sense alone?

Justintime50 (changes requested):

EasyPost/Parameters/CarrierAccount/Create.cs
nwithan8: `A` as in abstract, the same way `I` as in interface.

EasyPost/Constants.cs
nwithan8: While I disagree, I will re-add them, marked as deprecated.

EasyPost/Parameters/CarrierAccount/Create.cs
Justintime50: Ah, ok - thanks!

EasyPost/Constants.cs
Justintime50: Thank you, this would unfortunately break the semver contract without keeping them.

Justintime50 (approved):

jchen293 (approved):

PR #521, opened 2023-09-27
[feat] Add function to retrieve API keys for a specific user

nwithan8:
# Description

- Add function to retrieve API keys for a specific user

# Testing

- Add unit tests, cassettes

# Pull Request Type


EasyPost/Services/ApiKeyService.cs
```diff
@@ -33,6 +37,35 @@ internal ApiKeyService(EasyPostClient client)
...
+
+            if (apiKeyCollection.Id == id)
+            {
+                return apiKeyCollection.Keys;
+            }
+
+            if (apiKeyCollection.Children == null)
+            {
+                throw new FilteringError(string.Format(CultureInfo.InvariantCulture, Constants.ErrorMessages.NoObjectFound, "child"));
+            }
```
dcaballeroc, line 59: Wouldn't the `foreach` in line 61 be enough? Like, the LinQ query would find nothing, the loop doesn't execute and it throws at line 66

dcaballeroc (commented):
Have a small comment, but looks good overall

EasyPost/Services/ApiKeyService.cs
nwithan8: My compiler was initially complaining because the `Children` property is nullable (all our properties are nullable), but I'll double-check.

dcaballeroc: Might as well be the case. Keep thinking in terms of weakly-typed languages, which C# is not.

jchen293 (approved):
lgtm, can you add a changelog entry?

EasyPost/Services/ApiKeyService.cs
```diff
@@ -33,6 +37,35 @@ internal ApiKeyService(EasyPostClient client)
...
         public async Task<ApiKeyCollection> All(CancellationToken cancellationToken = default) => await RequestAsync<ApiKeyCollection>(Method.Get, "api_keys", cancellationToken);
 
+        /// <summary>
+        ///     Retrieve the <see cref="ApiKey"/>s for a specific <see cref="User"/>.
+        /// </summary>
+        /// <param name="id">The ID of the <see cref="User"/> to retrieve keys for.</param>
+        /// <param name="cancellationToken"><see cref="CancellationToken"/> to use for the HTTP request.</param>
+        /// <returns>A list of <see cref="ApiKey"/>s for the specified <see cref="User"/>.</returns>
+        /// <exception cref="FilteringError">Thrown if the specified <see cref="User"/> does not exist.</exception>
+        public async Task<List<ApiKey>?> RetrieveApiKeysFor(string id, CancellationToken cancellationToken = default)
```
jchen293, line 47: We should name this function as `api_keys()` to match other CLs

jchen293 (changes requested):

EasyPost/Services/ApiKeyService.cs
nwithan8: Resolved per our internal conversations.

dcaballeroc (approved):

PR #538, opened 2023-12-12
[feat] Add function to retrieve paginated list of child users

nwithan8:
# Description

- Add beta service for User
- Add `AllChildren` and `GetNextPageOfChildren` functions to beta User service
- Drop `BaseUser` class in favor of explicit re-implementation in `User` and `ReferralCustomer` classes (flatten out inheritance tree)

# Testing

- Add unit tests, cassettes
- Fix straggling back CRUD decorators

# Pull Request Type


EasyPost.Tests/ServicesTests/WithParameters/Beta/UserServiceTest.cs
```diff
@@ -0,0 +1,40 @@
...
+using EasyPost.Tests._Utilities;
+using EasyPost.Tests._Utilities.Attributes;
+using EasyPost.Utilities.Internal.Attributes;
+using Xunit;
+
+namespace EasyPost.Tests.ServicesTests.WithParameters.Beta
+{
+    public class UserServiceTests : UnitTest
+    {
+        public UserServiceTests() : base("beta_user_service_with_parameters", TestUtils.ApiKey.Production) =>
```
Justintime50, line 14: This isn't beta anymore right? It's also in the `Beta` directory?

```suggestion
        public UserServiceTests() : base("user_service_with_parameters", TestUtils.ApiKey.Production) =>
```

nwithan8: Good catch, forgot to delete the file once I migrated stuff out of it.

EasyPost/Models/API/ChildUserCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+/// </summary>
+public class ChildUserCollection : PaginatedCollection<User>
+{
+    #region JSON Properties
+
+    /// <summary>
+    ///     The Child <see cref="User"/>s in the collection.
+    /// </summary>
+    [JsonProperty("children")]
+    public List<User>? Children { get; set; }
```
jchen293, line 20: User and Children have different attributes, the list of users here doesn't seem correct. We should probably need to create a child model in C# lib.

nwithan8: After discussion, we decided not to make a separate `ChildUser` class due to serialization issues.

However, I took the opportunity to clean up some unnecessary inheritance by dropping the `BaseUser` class.

jchen293 (approved):

PR #539, opened 2023-12-21
[chore] Linting + docstrings

nwithan8:
# Description

Docstring linting has been enabled for the library, and as a result, every method, property and class needs to have a docstring (unless overridden individually).

This PR adds docstrings to almost all elements missing them, using existing documentation on EasyPost's website.

Some elements (mostly Options) do not have a docstring because they are not currently documented on EasyPost's website. These need to be individually verified and potentially removed if no longer valid.

This PR also addresses a number of additional, non-docstring linting issues. Some rules have been surpressed; these suppressions can be later removed once the styleguide has been altered.

Very few code changes have been made, and no breaking changes or end-user experience-altering changes have been introduced.

# Testing

- No changes to unit test functionality (some unit tests have been stylistically altered by linting). All unit tests pass; cassettes only re-recorded for expired cassettes.
- All remaining "missing docstring" linting issues are warning (not errors), so CI should pass.

# Pull Request Type


nwithan8:
Some cassettes had to be re-recorded because the order of the elements in the request JSON changed because properties in classes were re-ordered. Ultimately, the data was the exact same, but strict VCR rules made matching the original cassettes fail.

EasyPost/Models/API/Options.cs
```diff
@@ -116,8 +116,10 @@ public class Options : EasyPostObject
 
         [JsonProperty("carrier_insurance_amount")]
         public string? CarrierInsuranceAmount { get; set; }
```
Justintime50, line 118: question: can we globally ignore the comment requirements in just this file? That would clean up a bunch of these warnings on GitHub

Justintime50 (approved):
Property level comment documentation seems a bit excessive but is probably nice in an IDE.

Nice work

EasyPost/Models/API/Options.cs
nwithan8: Yes, I plan to make adjustments to the style guide for it.

jchen293 (approved):

PR #543, opened 2024-01-11
[chore] Make base EasyPostService extendable outside module

nwithan8:
# Description

The `EasyPostService` base class used for all services (e.g. `UserService`) was marked internal, meaning it could not be constructed or extended outside of the EasyPost module. This PR instead sets it to `protected`, meaning it can still not be constructed directly, but can be inherited. This can be helpful for end-users trying to make extensions or custom implementations of the services, allowing them to use the base service rather than having to re-implement all client HTTP functionality themselves.

# Testing

- All unit tests pass
- Change does not affect compilation, run-time or linting

# Pull Request Type


jchen293 (approved):

Justintime50 (approved):
Why not

PR #556, opened 2024-04-09
[chore] Display object type, ID when converting object to string

nwithan8:
# Description

- Override `ToString` for EasyPost objects to display object type and optional ID

# Testing

# Pull Request Type


jchen293:
Can we have a unit test to make sure the change is what we want?

PR #557, opened 2024-04-11
[fix] Make FedEx SmartPost parameter set accessible

nwithan8:
# Description

The constructor for `EasyPost.Parameters.CarrierAccount.CreateFedExSmartPost` was erroneously marked as internal and was not accessible to end-users. Because this is the only way users can create a FedEx Smartpost account using the Parameter objects in .NET, this was blocking some users from creating FedEx Smartpost accounts (without using the legacy raw dictionary approach).

# Testing

- Add unit test to integration test to confirm that every constructor for every parameter set is publicly accessible.

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Release
+
+- Fix accessibility for `CreateFedExSmartPost` parameter set
+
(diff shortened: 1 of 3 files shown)


Justintime50 (approved):
Changelog please

jchen293 (approved):

PR #558, opened 2024-04-11
[fix] Invalid `CarrierFields` structure

nwithan8:
# Description

This library was failing to deserialize a `CarrierAccount` from the API due to a failed deserialization of `CarrierFields`.

The previous assumption was that the API is returning "fields" in the JSON data for a `CarrierAccount` like so:

```json
{
  "id": "ca_123",
  "object": "CarrierAccount",
  ...
  "fields": {
    "credentials": [
      {
        "key": "account_number",
        "visibility": "visible",
        "label": "DHL Account Number",
        "value": "123456"
      },
      {
        "key": "country",
        "visibility": "visible",
        "label": "Account Country Code (2 Letter)",
        "value": "US"
      }
    ]
  }
}
```

When in fact, the API is returning "fields" like so:

```json
{
  "id": "ca_123",
  "object": "CarrierAccount",
  ...
  "fields": {
    "credentials": {
      "account_number": {
        "visibility": "visible",
        "label": "DHL Account Number",
        "value": "123456"
      },
      "country": {
        "visibility": "visible",
        "label": "Account Country Code (2 Letter)",
        "value": "US"
      }
    }
  }
}
```

Specifically, the "credentials" key under "fields" is not a list of dictionaries, but a dictionary of dictionaries, with the type of each credential field being the key in the key-value pair.

This PR updates the `CarrierFields` object structure to match the JSON schema.

Because the key for each type of credential field can by dynamic, users will have to access each credential field via a dictionary index:

```csharp
string countryLabel = myCarrierAccount.Fields.Credentials["country"].Label
```

This PR also adds the missing "logo" property for a `CarrierAccount`.

# Testing

- Add unit test to confirm deserialization works as expected
  - The carrier accounts used in our existing unit tests do not have fields returned by the API, so the serialization issue was not previously caught.

# Pull Request Type

### DIFF
--- a/EasyPost/Models/API/CarrierFields.cs
+++ b/EasyPost/Models/API/CarrierFields.cs
@@ -15,7 +15,7 @@ namespace EasyPost.Models.API
         ///     The credentials used in the production environment.
         /// </summary>
         [JsonProperty("credentials")]
-        public Dictionary<string, CredentialsField>? Credentials { get; set; }
+        public Dictionary<string, CarrierField>? Credentials { get; set; }
 
         /// <summary>
         ///     The credentials used in the test environment.
@@ -24,7 +24,7 @@ namespace EasyPost.Models.API
 #pragma warning disable SA1515
         // ReSharper disable once InconsistentNaming
 #pragma warning restore SA1515
-        public Dictionary<string, CredentialsField>? TestCredentials { get; set; }
+        public Dictionary<string, CarrierField>? TestCredentials { get; set; }
 
         /// <summary>
         ///     For USPS, this designates that no credentials are required.
@@ -44,7 +44,7 @@ namespace EasyPost.Models.API
     /// <summary>
     ///     Class representing a single EasyPost <see cref="CarrierAccount"/>'s credentials entry details.
     /// </summary>
-    public class CredentialsField : EasyPostObject
+    public class CarrierField : EasyPostObject
     {
         #region JSON Properties
 
--- a/EasyPost/Models/API/CarrierFields.cs
+++ b/EasyPost/Models/API/CarrierFields.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using EasyPost._base;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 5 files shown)


EasyPost/Models/API/CarrierFields.cs
```diff
@@ -14,7 +15,7 @@ public class CarrierFields : EasyPostObject
         ///     The credentials used in the production environment.
         /// </summary>
         [JsonProperty("credentials")]
-        public CarrierField? Credentials { get; set; }
+        public Dictionary<string, CredentialsField>? Credentials { get; set; }
```
Justintime50, line 18: issue: I don't like the name change here (even if the new name may be more correct or readable). It moves this from a bug fix to a breaking change which we should avoid. We can fix the underlying deserialization without users needing to alter their code aside from a version bump.

Justintime50 (changes requested):

EasyPost/Models/API/CarrierFields.cs
nwithan8: To be fair, because of this deserialization issue, a user has never been able to interact with a `CarrierField` object before, since every attempt to create one would have failed.

Justintime50: Correct, though again, this breaking change will require end-users to make changes in their code if they've already deployed this - even if today it's not working. Let's not major version bump over something that should just be a bug fix patch release.

Justintime50 (approved):

EasyPost/Models/API/CarrierFields.cs
```diff
@@ -47,12 +48,6 @@ public class CarrierField : EasyPostObject
     {
         #region JSON Properties
 
-        /// <summary>
-        ///     The key of the field.
-        /// </summary>
-        [JsonProperty("key")]
-        public string? Key { get; set; }
-
```
jchen293, line 55: Why is this removed?

nwithan8: Because it does not exist in the response JSON schema. Instead, the "key" is the actual key in the key-value dictionaries.

jchen293 (approved):

PR #561, opened 2024-04-24
[fix] Add missing parameters for creating orders

nwithan8:
# Description

- Add missing `BuyerAddress` and `ReturnAddress` parameters for `Order.Create` parameter set
# Testing

- N/A

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Release
+
+- Add missing parameters for `Order.Create` parameter set
+
 ## v6.3.1 (2024-04-12)
 
 - Fix `CarrierFields` serialization bug causing carrier account operations to fail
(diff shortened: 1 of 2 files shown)


Justintime50 (commented):
Can we get some tests for these or some e2e tests? Orders are kind of tricky where most of these details will typically be on each shipment and not on the parent order (addresses aside). I'd like to avoid adding params that aren't valid at the parent level by testing first.

nwithan8:
In draft while discussed internally.

nwithan8:
> Can we get some tests for these or some e2e tests? Orders are kind of tricky where most of these details will typically be on each shipment and not on the parent order (addresses aside). I'd like to avoid adding params that aren't valid at the parent level by testing first.

We have confirmed internally that these parameters are accepted in certain multi-parcel scenarios. It's not guaranteed that they will always be respected versus ignored, but it's important that we make them available for end-users to use as needed.

HottieHotton (approved):

Justintime50 (approved):
I feel better with just exposing return/buyer addresses

PR #574, opened 2024-07-09
[chore] Support new UPS carrier account create/update endpoint

nwithan8:
# Description

- Update carrier account creation/update functions to account for new UPS endpoint without introducing new functions
- Add/rework parameter sets for UPS carrier account creation to accommodate new schema
- Under-the-hood changes to carrier account parameter sets for determining endpoints based on type
  - Modeled after: https://github.com/nwithan8/easypost-extensions-dotnet/tree/master/EasyPost.Extensions/Parameters/CarrierAccount
- Add parameter sets for UPS carrier account update
- Mark existing legacy carrier account creation/update functions (using dictionaries) as obsolete, as they do not support the new UPS endpoints
  - Introduce guard clauses that will cause the legacy functions to error out if attempting to use them for a UPS account - guide users to the parameter set-based alternatives
- Under-the-hood consolidation of dictionary mutations for parameter JSON serialization

Outside of obsolete warnings (and guard clauses if users decide to ignore the obsolete warnings), there are no breaking changes to existing workflows. Only those creating/editing UPS accounts will need to account for the changes made in this PR.

# Testing

- New unit tests to confirm UPS endpoint switching, guard clauses, parameter set inheritance and serialization
- Added/re-recorded cassettes as needed

# Pull Request Type


EasyPost/Constants.cs
```diff
@@ -136,19 +138,72 @@ public static class CarrierAccounts
...
+                string endpoint = StandardCreateEndpoint;
+
+                var @switch = new SwitchCase
+                {
+                    { CarrierAccountType.FedEx.Name, () => endpoint = CustomCreateEndpoint },
+                    { CarrierAccountType.FedExSmartPost.Name, () => endpoint = CustomCreateEndpoint },
+                    { CarrierAccountType.Ups.Name, () => endpoint = UpsOAuthCreateEndpoint },
+                    { CarrierAccountType.UpsMailInnovations.Name, () => endpoint = UpsOAuthCreateEndpoint },
+                    { CarrierAccountType.UpsSurePost.Name, () => endpoint = UpsOAuthCreateEndpoint },
+                    { SwitchCaseScenario.Default, () => endpoint = StandardCreateEndpoint },
```
Justintime50, line 173: suggestion: just because we can use a switch case (and I know you were really gunning for it), doesn't mean we need to 😂 this seems to be way more complicated than it needs to be because we only have 2 options. If/else works super well in those cases. Same for the update below. I want to see much less code and complexity here instead of solving for a problem that doesn't yet exist (we don't have a massive matrix of urls to account types yet).

nwithan8: Improved

Justintime50 (approved):

EasyPost/Services/CarrierAccountService.cs
```diff
@@ -110,10 +120,23 @@ public async Task<CarrierAccount> Create(Parameters.CarrierAccount.ACreate param
         /// <param name="cancellationToken"><see cref="CancellationToken"/> to use for the HTTP request.</param>
         /// <returns>The updated <see cref="CarrierAccount"/>.</returns>
         [CrudOperations.Update]
+        [Obsolete("This function does not support UPS accounts. Please use the Update(string, Parameters.CarrierAccount.Update) function instead for all carrier account updates.")]
```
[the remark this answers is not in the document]
jchen293: Bump, this hasn't been addressed. Please let me know if I misunderstood anything

EasyPost/Services/CarrierAccountService.cs
```diff
@@ -37,19 +38,22 @@ internal CarrierAccountService(EasyPostClient client)
         /// <param name="cancellationToken"><see cref="CancellationToken"/> to use for the HTTP request.</param>
         /// <returns>A <see cref="CarrierAccount"/> object.</returns>
         [CrudOperations.Create]
+        [Obsolete("This function does not support FedEx or UPS account registrations. Please use the Create(Parameters.CarrierAccount.ACreate) function instead for all carrier account registrations.")]
```
jchen293, line 41: Question: has it been this way in the past? just want to confirm that we are not making breaking change

nwithan8: In general, we're trying to drive users away from using the legacy dictionary-based methods to the new parameter-set-based methods. While this function could technically support custom workflows like FedEx (but not UPS), it's easier to just say that it's not recommended for either of them, and suggest users use the parameter-set alternatives instead to avoid any potential incompatibilities.

jchen293 (approved):

PR #584, opened 2024-08-16
[chore] Drop .NET 5.0 support, add .NET 9.0 support

nwithan8:
# Description

- Drop .NET 5.0 support for EasyPost library
  - Remove .NET 5.0 from target frameworks for library and test suites
  - Remove .NET 5.0 target framework and binaries from NuGet package build
  - Remove .NET 5.0 from CI
  - Remove .NET 5.0 installation via environment setup scripts
- Add .NET 9.0 support
  - Add .NET 9.0 to target frameworks for library and test suites
    - Use .NET 9.0 for FSharp and VB compatibility check
  - Add .NET 9.0 target framework and binaries to NuGet package build
  - Add .NET 9.0 to CI
  - Add .NET 9.0 installation via environment setup scripts

# Testing

# Pull Request Type


nwithan8:
Good thing to know: Accidentally confirmed that even when .NET 5.0 is not explicitly listed as a target framework in the library, it did still run in a .NET 5.0 environment successfully. Given that we haven't made any framework-breaking changes, it's likely that our library will continue to work in .NET 5.0 (and probably .NET Core 3.1) even if we don't explicitly support/test it in those versions.

EasyPost/Parameters/Tracker/CreateList.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
Justintime50, line 1: question: didn't we remove the function this parameter set is used for? If so we should also clean this up since it cannot be used.

Justintime50 (approved):

EasyPost/Parameters/Tracker/CreateList.cs
nwithan8: We did, it's being removed in a separate PR of breaking changes.

jchen293 (approved):

Issue #520, opened 2023-09-22
[Feat]: Change pickup_min_datetime and pickup_max_datetime Shipment Options types to string


gkeasypost:
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

In the Options model, the `pickup_min_datetime` and `pickup_max_datetime` options should be a string type instead of a DateTime type in order to maintain consistency with what the EasyPost API returns for those options, as well as the same type as `delivery_min_datetime` and `delivery_max_datetime`. 

see: 
https://www.easypost.com/docs/api#options 
https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Models/API/Options.cs#L683
https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Models/API/Options.cs#L689

nwithan8:
Hey, thanks for reaching out! This is being discussed internally about the best way to handle non-primatives, particularly dates, in our client libraries, whether to keep them as strings as they are in the raw JSON, or whether to de/serialize them to language-specific elements (e.g. a `DateTime` in .NET).

Right now, it seems there is a mix of both throughout the library, so we will try to standardize it one way or another via a breaking change release.

We'll keep you updated. If any other developer would like to weigh in on what would be suit their expectations, please chime in!

joe-rohde:
Standardizing these dates is important for working with different carrier requirements.  Is there a timeline on when this will be addressed?

nwithan8:
Hey @joe-rohde , thanks for weighing in.

Unfortunately, we do not have a projected timeline or any updates to share regarding this.

If you could, can you elaborate on how the `DateTime` versus `string` type of a date is interfering with your usage of EasyPost? It will help us diagnose the issue and influence our decision about which paradigm to adopt.

joe-rohde:
We utilize multiple date fields across multiple carriers.

An example of one issue we are having is with the carrier CDL.  They require specific time windows for picking up and delivering.

When posting to the Easypost endpoint, we set specific date/times:
•             "pickup_min_datetime": "2023-09-21T06:00:00"
•             "pickup_max_datetime": "2023-09-21T08:00:00"
•             "delivery_min_datetime": "2023-09-21T08:00:00"
•             "delivery_max_datetime": "2023-09-21T20:00:00"
EasyPost converts these dates to different ways:
•             "pickup_min_datetime": "2023-09-21T06:00:00.000Z"
[...]

PR #593, opened 2024-10-04
[feat] Add missing options

### LINKED ISSUE #592: Add support for "merchant_id" shipment option
### Feature Request Is New

- [X] I have verified that the requested feature does not already exist or has not already been requested.

### Description of the feature

The `merchant_id` shipment option is required for shipping with the DHL eCommerce Solutions by Default carrier account (see in [DHL eCommerce (Default) International Shipping Guide](https://support.easypost.com/hc/en-us/articles/30037931035533-DHL-eCommerce-Default-International-Shipping-Guide#:~:text=options%20are%20required%3A-,merchant_id,-%3A%20Must%20be))

It looks like this option needs to be added here so that C# library users can use the DHL eCS default account: https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Models/API/Options.cs

nwithan8:
# Description

Add all missing shipping options.

NOTE: Some options are/continue to be the wrong type (e.g. string instead of bool, string instead of DateTime (#520)). This PR is purely additive, with only new options and documentation on existing options to point out where a type is invalid. We should consider fixing these in a breaking change in the future.

Closes #592 

# Testing

# Pull Request Type

### DIFF
--- a/EasyPost/Models/API/Options.cs
+++ b/EasyPost/Models/API/Options.cs
@@ -40,6 +40,9 @@ namespace EasyPost.Models.API
         [JsonProperty("address_validation_level")]
         public string? AddressValidationLevel { get; set; }
 
+        [JsonProperty("administratively_unpurchasable")]
+        public bool? AdministrativelyUnpurchasable { get; set; }
+
         /// <summary>
         ///     Set this option to true if your shipment contains alcohol.
         ///     <list type="table">
@@ -66,6 +69,9 @@ namespace EasyPost.Models.API
         /// </summary>
         [JsonProperty("alcohol")]
         public bool? Alcohol { get; set; }
+
+        [JsonProperty("auto_manifest")]
+        public bool? AutoManifest { get; set; }
 
         /// <summary>
         ///     A reference ID for aggregating DHL eCommerce billing data.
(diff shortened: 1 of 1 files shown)


EasyPost/Models/API/Options.cs
```diff
@@ -114,8 +120,14 @@ public class Options : EasyPostObject
...
 
+        [JsonProperty("carrier_branded")]
+        public bool? CarrierBranded { get; set; }
+
+        [JsonProperty("carrier_instructions")]
+        public Dictionary<string, object?>? CarrierInstructions { get; set; } // hash
+
         [JsonProperty("carrier_insurance_amount")]
-        public string? CarrierInsuranceAmount { get; set; }
+        public string? CarrierInsuranceAmount { get; set; } // number
```
jchen293, line 130: Should the comment be `any`? 
```suggestion
        public string? CarrierInsuranceAmount { get; set; } // any
```

jchen293 (commented):
Why do some options have a type comment while others don't?

nwithan8:
> Why do some options have a type comment while others don't?

Those with comments are to denote options that are technically the wrong type (e.g. a "string" type, but server-side, it parses it as a number. An end-user could potentially provide a non-numeric string, which would cause issues server-side). These should be considered for changes during the next breaking changes release cycle.

Justintime50:
As mentioned in the past, I don't believe exposing all available options is the right play. Many are intentionally undocumented for various reasons - many need to be vetted for validity. I would caution against moving forward with this PR and instead continue having users rely on the custom option override you built to allow passing options that may not be officially supported.

Justintime50:
Shelving for now due to prioritization

PR #601, opened 2024-10-29
[fix] ]Handle serializing array/list parameters in GET/DELETE query strings

nwithan8:
# Description

Using the new `tracking_codes` parameter for the `Tracker.All`parameter set was not serializing parameters properly, resulting in the value (a list of tracking codes) being lost in transit.

This PR introduces a fix in the query parameter construction process that will account for arrays. Specifically, the following list of tracking codes "001", "002", "003" will result in the following URL + query string being sent to the API: `https://api.easypost.com/v2/trackers?tracking_codes[]=001&tracking_codes[]=002&tracking_codes[]=003`. This is [the format expected by our backend](https://til.hashrocket.com/posts/syzssfozay-pass-array-of-values-as-query-string-in-rails-) for proper parsing.

Accomplishing this requires some string manipulation in the query building process, as this is not the typical way the built-in `System.Uri` in  .NET handles query arrays (typically would use comma-separated lists instead, producing `https://api.easypost.com/v2/trackers?tracking_codes=001,002,003`, which is incompatible with the API server).

`tracking_codes` is currently the only query-based parameter (GET/DELETE request) that is an array, so this issue did not arise until it was recently introduced.

# Testing

- New unit test to confirm serialization
- No existing cassettes needed to be re-recorded, suggesting change does not impact other serialization flows

# Pull Request Type

### DIFF
--- a/EasyPost.Tests/HttpTests/RequestTest.cs
+++ b/EasyPost.Tests/HttpTests/RequestTest.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Net.Http;
 using EasyPost._base;
 using EasyPost.Http;
 using Xunit;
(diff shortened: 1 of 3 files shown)


EasyPost.Tests/HttpTests/RequestTest.cs
```diff
@@ -17,5 +18,37 @@ public void TestRequestDisposal()
...
+        };
+        const string key2 = "key2";
+        const string value2 = "value3";
+        var parameters = new Dictionary<string, object>
+        {
+            { key, value },
+            { key2, value2 }
+        };
+
+        Request request = new("https://google.com", "not_a_real_endpoint", Method.Get, ApiVersion.V2, parameters, new Dictionary<string, string> { { "header_key", "header_value" } });
```
Justintime50, line 38: suggestion: we should be using `example.com` for these kinds of things to be nice to our google neighbors.

EasyPost.Tests/HttpTests/RequestTest.cs
```diff
@@ -17,5 +18,37 @@ public void TestRequestDisposal()
...
+
+        HttpRequestMessage httpRequestMessage = request.AsHttpRequestMessage();
+
+        string queryString = httpRequestMessage.RequestUri?.Query;
+
+        string keyWithArray = $"{key}[]";
+
+        foreach (string item in value)
+        {
+            Assert.Contains($"{keyWithArray}={item}", queryString); // Does not account for URL encoding
```
Justintime50, line 48: issue: assertions should be explicit and hard-coded. Otherwise this test will always pass. The assertion is the blueprint, the thing we actually expect which shouldn't be dynamic for that reason.

Justintime50 (changes requested):

EasyPost.Tests/HttpTests/RequestTest.cs
nwithan8: These requests aren't actually executed, it's just testing the construction process.

Justintime50: Even still, best practice

EasyPost.Tests/HttpTests/RequestTest.cs
nwithan8: > Otherwise this test will always pass

The test is not asserting the output, the test is asserting a given input produces an output in a specific format, where the input is irrelevant.

Regardless, resolved.

Justintime50 (approved):

jchen293 (approved):

PR #609, opened 2025-03-11
feat: adding custom webhook headers

hpatsalla:


EasyPost/Parameters/Webhook/Create.cs
```diff
@@ -24,8 +26,12 @@ public class Create : BaseParameters<Models.API.Webhook>, IWebhookParameter
         /// </summary>
         [TopLevelRequestParameter(Necessity.Required, "url")]
         public string? Url { get; set; }
-
-        // TODO: Add custom_headers
+                
+        /// <summary>
+        ///     custom headers
+        /// </summary>        
+        [TopLevelRequestParameter(Necessity.Optional, "customheaders")]
```
Justintime50, line 33:
```suggestion
        [TopLevelRequestParameter(Necessity.Optional, "custom_headers")]
```

issue: this will need to be an underscore

EasyPost/Models/API/WebhookCustomHeader.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    public class WebhookCustomHeader
+    {
+        #region JSON Properties
+
+        /// <summary>
+        ///     Custom Header name
+        /// </summary>
+        [JsonProperty("name")]
+        public string? Name { get; set; }
```
Justintime50, line 19:
```suggestion
        public string Name { get; set; }
```

issue: I don't believe these are optional, they're either there or not at all

EasyPost/Models/API/WebhookCustomHeader.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        ///     Custom Header value
+        /// </summary>
+        [JsonProperty("value")]
+        public string? Value { get; set; }
```
Justintime50, line 25:
```suggestion
        public string Value { get; set; }
```

Same here

EasyPost/Parameters/Webhook/Update.cs
```diff
@@ -18,8 +19,10 @@ public class Update : BaseParameters<Models.API.Webhook>
         /// </summary>
         [TopLevelRequestParameter(Necessity.Optional, "webhook_secret")]
         public string? Secret { get; set; }
-
-        // TODO: Add custom_headers
+        
+        
+        [TopLevelRequestParameter(Necessity.Optional, "customheaders")]
```
Justintime50, line 24:
```suggestion
        [TopLevelRequestParameter(Necessity.Optional, "custom_headers")]
```

issue: needs to be underscored here

Makefile
```diff
@@ -110,8 +110,9 @@ test:
 ## unit-test - Run the unit tests for a specific framework
 # @parameters:
 # fw= - The framework to build for.
+# Always run unit tests in Debug mode to allow access to internal members
```
Justintime50, line 113: question: was this change needed?

Justintime50 (changes requested):

Makefile
hpatsalla: Yes. The `make unit-test` command is failing when the comment is on the same line. So, I relocated the comment.

Justintime50 (approved):

Justintime50 (approved):
Nice work!

PR #612, opened 2025-03-24
Alternative error format

Justintime50:
# Description

- Fixes error parsing
  - Allows for alternative format of `errors` field (previously we deserialized the `errors` field into a list of `Error` objects; however, sometimes the errors are simply a list of strings. This change make the `errors` field a list of `objects` allowing for either the new `FieldError` object or a list of strings. Users will need to check for the type of error returned and handle appropriately)
  - Removed the unused `Error` model
  - Added an explicit `AddressVerificationFieldError` model
  - The `PaymentRefund` class now uses a list of `FieldError` instead of `Error` for the `errors` field
- Overhaul CI so we only download the .NET SDKs we actually need for each step. Should drastically reduce time spent on CI for the test job

# Testing

- Adds alternative error format test to ensure fixes work as expected
- Removes tests that were stale from before major changes
- Adds the same tests we use in all the other libs for error handling for a consistent experience (found bugs around error handling)

# Pull Request Type


ralexmatthews (approved):
I do not know c#, but the changes look fine to me 👍

PR #655, opened 2026-04-03
fix: lint

Justintime50:
# Description

Fixes various warnings from lint that have been around awhile

# Testing

# Pull Request Type


sroomberg-ep (approved):
Should the warnings be done at the top level rather than in each file?

Justintime50:
> Should the warnings be done at the top level rather than in each file?

For this one no, the warning we're ignoring is adding docstrings to properties. These 3ish files have properties that are undocumented so it makes sense to only ignore these ones.

